PR #4, opened 2016-12-03
Remove Zero width space in README.md

pocke:
I found zero width spaces in README.md .
It seems unnecessary. So, I removed it.

Unicode document: http://www.fileformat.info/info/unicode/char/200B/index.htm

GitHub UI maybe can't display the difference.
You can see it by `git diff` command in terminal.


wata727:
Oh... I did not notice, thanks!

PR #31, opened 2017-01-06
Return error code when issue exists

ytet5uy4:
## WHY
When used tflint on CI, it's didn't failed although tflint return some errors.

## WHAT
Update to return 2 when issue exists.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -16,6 +16,7 @@ import (
 const (
 	ExitCodeOK    int = 0
 	ExitCodeError int = 1 + iota
+	ExitCodeIssuesFound
 )
 
 // CLI is the command line object
@@ -37,17 +38,18 @@ type TestCLIOptions struct {
 // Run invokes the CLI with the given arguments.
 func (cli *CLI) Run(args []string) int {
 	var (
-		version      bool
-		help         bool
-		debug        bool
-		format       string
-		ignoreModule string
-		ignoreRule   string
-		configFile   string
-		deepCheck    bool
-		awsAccessKey string
-		awsSecretKey string
-		awsRegion    string
+		version         bool
+		help            bool
+		debug           bool
+		format          string
+		ignoreModule    string
+		ignoreRule      string
+		configFile      string
+		deepCheck       bool
+		awsAccessKey    string
+		awsSecretKey    string
+		awsRegion       string
+		errorWithIssues bool
 	)
 
 	// Define option flag parse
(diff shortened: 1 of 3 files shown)


wata727:
Thanks for your proposal.
However, I think that it is not desirable to return the same status code as it was when an error occurred in core. Perhaps, it should return status code for that reason.

In my opinion, status code returned by lint tools should be customizable. We should be able to control it by runtime arguments as follows without changing the default behavior.

```
$ tflint
# return exit status code is 0 

$ tflint --exit-error-status
# return exit status code is something that is neither 0 nor 1
```

ytet5uy4:
I've updated to match your opinion.
Is my understanding correct?

cli.go
```diff
@@ -148,6 +150,12 @@ func (cli *CLI) Run(args []string) int {
 
 		p := printer.NewPrinter(cli.outStream, cli.errStream)
 		p.Print(issues, format)
+
+		if exitErrorStatus {
+			if len(issues) > 0 {
```
wata727: [nitpicks] `if exitErrorStatus && len(issues) > 0`

cli.go
```diff
@@ -71,6 +72,7 @@ func (cli *CLI) Run(args []string) int {
 	flags.StringVar(&awsAccessKey, "aws-access-key", "", "AWS access key used in deep check mode.")
 	flags.StringVar(&awsSecretKey, "aws-secret-key", "", "AWS secret key used in deep check mode.")
 	flags.StringVar(&awsRegion, "aws-region", "", "AWS region used in deep check mode.")
+	flags.BoolVar(&exitErrorStatus, "exit-error-status", false, "return exit status code when issue exists.")
```
wata727: I'm sorry... Please use another option name 🙇 
I think `--exit-error-status` is bad name. (My recommendation is `--error-with-issues`)

cli.go
```diff
@@ -148,6 +150,12 @@ func (cli *CLI) Run(args []string) int {
 
 		p := printer.NewPrinter(cli.outStream, cli.errStream)
 		p.Print(issues, format)
+
+		if exitErrorStatus {
+			if len(issues) > 0 {
+				return ExitCodeError
```
wata727: Please use another status code. (ex: `ExitCodeIssuesFound`, etc.) `ExitCodeError` is reserved for execution error.

wata727:
Great! Finally, Please squash to one commit as well :)

PR #51, opened 2017-02-05
SideCI Setting

wata727:


wata727:
Umm... I want to ignore `go vet` only for test file...

PR #137, opened 2017-08-27
Fix unclear error messages

### LINKED ISSUE #136: [Question] Which files produced the lint error?
Running `tflint` on it's own does not give too much feedback at the moment.

```bash
$ tflint
1:3: unknown variable accessed: var.aws_allowed_account_ids
```

I'd like to know more specifically which files were inspected, so I've wrapped this command like so

```bash
find . -type f -not \( -path '*/.terraform/*' \) -name '*.tf' -exec \
  /bin/bash -c "echo 'Linting {}' && tflint {} --deep --error-with-issues" \;
Linting ./main.tf
1:3: unknown variable accessed: var.aws_region
Linting ./output.tf
Awesome! Your code is following the best practices :)
Linting ./variables.tf
Awesome! Your code is following the best practices :)
```

but by linting one file at a time, it has no knowledge of the other files. So, is there a way to add more information about what was linted and which file is causing issues?

wata727:
Fixes #136 

## Before

```
$ tflint
1:3: unknown variable accessed: var.instance_typ
```

## After

```
$ tflint
Evaluation error: 1:3: unknown variable accessed: var.instance_typ in template_override.tf:9
```

### DIFF
--- a/evaluator/module.go
+++ b/evaluator/module.go
@@ -1,9 +1,10 @@
 package evaluator
 
 import (
-	"strings"
-
+	"errors"
+	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/hashicorp/hil"
 	"github.com/hashicorp/hil/ast"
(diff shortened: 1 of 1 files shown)


missingcharacter:
LGTM

PR #147, opened 2017-10-01
Support variables from environment variables

### LINKED ISSUE #141: Support environment variable variables
I have being trying to use tflint but hit a roadblock when loading variables from env vars

```
> TF_VAR_THIS_THING=foo tflint
1:3: unknown variable accessed: var.THIS_THING
```

wata727:
Fixes #141 

See: https://www.terraform.io/docs/configuration/variables.html#environment-variables

### DIFF
--- a/evaluator/variable.go
+++ b/evaluator/variable.go
@@ -1,7 +1,9 @@
 package evaluator
 
 import (
+	"os"
 	"reflect"
+	"strings"
 
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


evaluator/variable.go
```diff
@@ -96,6 +147,13 @@ func parseVariable(val interface{}, varType string) hilast.Variable {
 		varType = HCL_STRING_VARTYPE
 	}
 
+	return varType
+}
+
+func parseVariable(val interface{}, varType string) hilast.Variable {
```
wata727: __[gometalinter]__
* [staticcheck] argument varType is overwritten before first use (SA4009)

PR #159, opened 2017-11-12
Use `cristim/ec2-instances-info` instead of hard-coded list

wata727:
- Add `db.r4.*` family (for Aurora only)
- Add `x1e.32xlarge`, `c5.*` family and `p3.*` family


wata727:
That's a good idea. I will try it :)

PR #181, opened 2018-04-20
(Feature) Support terraform.workspace variable

Stretch96:
* Using ${terraform.workspace} currently causes `tflint` to fail:
`unknown variable accessed: terraform.workspace in terraform.tf`
* Resolves https://github.com/wata727/tflint/issues/177


Stretch96:
Currently this just blindly sets `TerraformWorkspace` to `default`.

Ideally this would actually set it's value from the `terraform workspace show` output, but unsure of how to go about that. Also should decide whether it should only set it if deep check is enabled (`--deep`)

wata727:
Thanks for your contribution!

> Ideally this would actually set it's value from the terraform workspace show output, but unsure of how to go about that. Also should decide whether it should only set it if deep check is enabled

I guess it will behave similarly to `TerraformEnv`. Since the current environment is stored in `.terraform/environment`, `TerraformWorkspace` also should refer to this.
https://github.com/wata727/tflint/blob/30023dbfbdaf3a15feb28453cefcec4c6c8d196b/config/config.go#L38-L40

Stretch96:
Ah yes, I'll add that to the PR :+1:

Stretch96:
@wata727 That's added to the PR now 👍

evaluator/evaluator.go
```diff
@@ -25,6 +25,10 @@ func NewEvaluator(templates map[string]*hclast.File, schema []*schema.Template,
 		Type:  hilast.TypeString,
 		Value: c.TerraformEnv,
 	}
+	varMap["terraform.workspace"] = hilast.Variable{
```
wata727: Please add a test case for `${terraform.workspace}` syntax :)
Example:
https://github.com/wata727/tflint/blob/3cf4a7eff3d7869da768ae3e51d8699d77c4c1df/evaluator/evaluator_test.go#L267-L295

Stretch96: Yep sure, I'll add that in soon 😬

Stretch96: Ready for review :+1:

wata727:
👍

PR #194, opened 2018-05-16
Detect pinned sources using regular expressions

### LINKED ISSUE #193: terraform_module_pinned_source is not careful enough with it's regex
When this check reviews the `source` string, it simply looks for `git` and `bitbucket`, and not a proper git URL. Terms like `gitlab` or `foogitfoo` trigger the validation check and will fail. This leads to a false-positive.

For example: 	`WARNING:348 Module source "../modules/init-snippet-gitlab-docker" is not pinned (terraform_module_pinned_source)`

https://github.com/wata727/tflint/blob/v0.5.4/detector/terraform_module_pinned_source.go#L25-L37 needs to be updated to ensure there are no false-positives, it should check If the source URL is a valid git URL.

kerscher:
Resolves #193 

- Substitutes `strings.Contains` for `regexp.MatchString`
- Uses `regexp.MustCompile` as regexp is checked during tests
- Adds tests for generic git repositories with HTTPS and SSH formats

### DIFF
--- a/detector/terraform_module_pinned_source.go
+++ b/detector/terraform_module_pinned_source.go
@@ -2,6 +2,7 @@ package detector
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/wata727/tflint/issue"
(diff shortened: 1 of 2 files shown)


kerscher:
(I love this tool by the way. I can’t see myself developing without it.)

wata727:
Great! Thanks for your contributions 👏

PR #197, opened 2018-05-20
Add new `rule` configuration syntax

### LINKED ISSUE #191: Introduce new configuration format
TFLint has focused on detecting errors until now, but I'll focus on more other problems (e.g. https://github.com/wata727/tflint/issues/178, https://github.com/wata727/tflint/issues/174) in the future.

For that, I'm planning to introduce the new configuration format like the following:

```hcl
config {
  // it also works
  ignore_rule = {
    aws_instance_invalid_type  = true
    aws_instance_previous_type = true
  }
}

// new format
rule "aws_instance_invalid_type" {
  enabled = false
}

rule "aws_instance_previous_type" {
  enabled = false
}
```

This allows you to introduce rules that are disabled by default. In addition, you can pass more detailed settings for each rule.
I have not completed the design yet, so it may change in the future. Feedbacks are welcome. Thanks!

wata727:
Fixes #191


wata727:
Umm... If `config` block is not set, `DecodeBody` returns an error :(

```
.tflint.hcl:4,1-1: Missing config block; A config block is required.
```

Issue #205, opened 2018-06-08
Variables used when declaring modules aren't recognized


martinb3:
Hi there -- trying tflint for the first time, and wondering if I'm hitting something related to #167 or not.

I've got [a minimal example I've pushed up as a separate repo](https://github.com/martinb3/tflint-example-module-variables) if you'd like to see it there.

In short, if I have a variable declared and used to create a module, tflint says it isn't defined:
```
variable "bar" {}

module "module1" {
  source = "./module1"

  foo = "${var.bar}"
}
```

And tflint's response (with `--debug`):
```
$ tflint --debug
DEBUG:/loader.go:121 [INFO] Load environment...
DEBUG:/loader.go:129 [ERROR] open .terraform/environment: no such file or directory
DEBUG:/loader.go:132 [INFO] Load tfstate...
DEBUG:/loader.go:136 [ERROR] stat terraform.tfstate: no such file or directory
DEBUG:/loader.go:139 [ERROR] stat .terraform/terraform.tfstate: no such file or directory
DEBUG:/loader.go:162 [INFO] Load tfvars...
DEBUG:/loader.go:165 [INFO] Load `terraform.tfvars`
DEBUG:/loader.go:167 [ERROR] stat terraform.tfvars: no such file or directory
DEBUG:/loader.go:192 [INFO] Load HCL file: `main.tf`
DEBUG:/loader.go:60 [INFO] Load HCL file: `main.tf`
Evaluation error: 1:3: unknown variable accessed: var.bar in main.tf:6
```

It seems that some fields don't allow interpolation when calling a module, as a terraform limitation, but I should at least be able to pass a variable to a module's inputs.

Thanks!

wata727:
@martinb3 Thanks for your detailed report! I confirmed this problem.

In the example, TFLint cannot interpolate `var.bar` because this variable doesn't have a default value. In Terraform, you can enter a value when calling `terraform plan` or `terraform apply` but TFLint cannot. Maybe there is room for improvement in this behavior.

TFLint supports [varfiles](https://www.terraform.io/intro/getting-started/variables.html#from-a-file) and [TF_VAR environment variables](https://www.terraform.io/intro/getting-started/variables.html#from-environment-variables), so you can avoid the problem by using these.

```
$ tflint
Evaluation error: 1:3: unknown variable accessed: var.bar in main.tf:6
$ TF_VAR_bar=1 tflint
Awesome! Your code is following the best practices :)
$ tflint --var-file example.tfvars
[...]

Issue #204, opened 2018-06-07
File function is unknown


alshabib:
This valid terraform syntax 

private_key = "${var.use_literal_key ? var.private_key : file(var.private_key_file)}"

causes

Evaluation error: 1:43: unknown function called: file in build.tf:37

This is part of a module definition. Is there any workaround?

wata727:
Hi @alshabib,

Sorry for the inconvenience. This is a bug that occurs because Terraform's variable interpolation is not fully supported. In the future, I'm going to fix the bug by embedding Terraform as a library.

As a workaround, you can specify `ignore_module` like https://github.com/wata727/tflint/issues/167#issuecomment-385188499

wata727:
I just released TFLint v0.8.0 and fix the issue. Please try the latest version again. If you still have problems, please open a new issue instead of reopening.

Issue #167, opened 2018-01-24
Support Terraform 0.11 module sources


brainsik:
Terraform 0.11 introduced support for module installation from Terraform Registry. As a consequence, module source strings using relative paths must be expressed explicitly by beginning the string with `./` or `../`.  See [Relative Paths in Module source](https://www.terraform.io/upgrade-guides/0-11.html#relative-paths-in-module-source).

`tflint` is using the old interpretation of module `source`:

```
$ tflint --version
TFLint version 0.5.4
$ tflint 
ERROR: module `terraform-aws-modules/vpc/aws` not found. Did you run `terraform get`?
```

davewongillies:
I'm hitting on this issue too when using the Terraform Enterprise private module registry:

```
$ tflint
ERROR: module `app.terraform.io/my-org/my-provider/my-module` not found. Did you run `terraform get`?
```

wata727:
Currently, TFLint's module resolver is incomplete. Until now applied a partial patch from Terraform's implementation, but it seems that there was a big change in 0.11.

In the future, I plan to solve this problem by using Terraform's module resolver. However, It is necessary to wait for new Terraform version.

davewongillies:
For those looking for a workaround, I'm making use of `tflint`'s `--ignore-module` option, which can also be set in `.tflint.hcl`

`.tflint.hcl:`
```hcl
config {
  ignore_module = {
    "app.terraform.io/my-org/my-provider/my-module" = true
  }
```

wata727:
Hi there, I just released TFLint v0.8.0 that supports v0.11 and v0.12 modules. Please try the latest version again. If you still have problems, please open a new issue instead of reopening.

Issue #160, opened 2017-11-16
formatlist is unknown


matti:
```
Evaluation error: 1:83: unknown function called: formatlist in e.tf:6
```

wata727:
Thanks @matti 
This problem is caused by TFLint not supporting all Terraform syntax. Generally, unsupported syntaxes are ignored, but errors occurred when using them as attributes of modules.

This is a difficult problem. As a workaround, I plan to implement a feature that can ignore the module causing the error.

wata727:
Oh no, I was the misunderstanding. Originally it will not occur an error, just calling `formatlist` should be ignored.

Can you show me the `e.tf` for further investigation?

wata727:
Close this issue because the current implementation has changed significantly. Please try the latest version again. If you still have problems, please open a new issue instead of reopening.

PR #209, opened 2018-07-05
Improve TFLint 🚀 

wata727:
Fixes https://github.com/wata727/tflint/issues/179

This pull request changes the internal system drastically by embedding Terraform as a library. Therefore, some breaking changes are included, please see the following for details.

**NOTE: This changelog may be changed later because it is draft.**

## Breaking Changes 🔥 
- Always return an error when failed to evaluate an expression #211 
  - Until now, except for module attributes, even if an error occurred, it was ignored.
  - However, expressions including unsupported syntax (such as `${module.foo}`) are not evaluated, so no error occurs.
- Drop support for `${terraform.env}` #211 
  - Previously `${terraform.env}` was a valid expression that returned the same as `${terraform.workspace}`.
  - This is because Terraform v0.12 doesn't support `${terraform.env}`.
- The file name of a module includes module ID instead of the `source` attribute #211 
  - Up to now it was output like `github.com/wata727/example-module/instance.tf`, but it will be changed like `module_id/instance.tf`.
- Always parse all configuration files under the current directory #229 #232 
  - When passing a file name as an argument, TFLint only parsed that file so far, but it now parses all configuration files under the current directory.
  - Also, file arguments are only used to filter the issues obtained. Therefore, you cannot specify files other than under the current directory.
  - As a known issue, If file arguments are specified, module's issues are not reported. This will be improved by changing handling of module's issues in the future.
  - These behaviors have been changed as it depends on Terraform's `configload` package.
  - In addition, modules are always loaded regardless of `ignore_module`.
- Raise an error when using invalid syntax as the Terraform configuration #217 
  - For example, it didn't raise an error when using `resources`(not `resource`) block because it is valid as HCL syntax in previous versions. 
- Remove `--debug` option #223 
  - Please use `TFLINT_LOG` environment variables instead.
- Raise an error when the file specified by `--config` does not exist #224 
  - Previously the error was ignored and the default config was referenced.
- Remove duplicate resource rules #227 
  - This is due to technical difficulty and user experience.

## Enhancements 💪 
- Support new HCL syntax #211
  - See also: https://www.hashicorp.com/blog/terraform-0-1-2-preview
- Support built-in functions #211 
  - Until now, if an expression includes function, it was ignored.
- `TF_DATA_DIR` and `TF_WORKSPACE` environment variables are now available #211 
  - Until now, these variables are ignored.
- It is now possible to handle values doesn't have a default without raising errors #211 #220 
  - In the past, an error occurred when there was a reference to a variable that had no default value in an attribute of a module. See #205 
- Support Terraform v0.11 modules #217 
  - Until now, it is failed to properly load a part of Terraform v0.11 module. See also #167 
- Support for automatic loading `*.auto.tfvars` files #228 
  - Previously it was not loaded automatically.

## BugFixes 🙇 
- Improve expression checks #211 
  - Since it used to be checked by a regular expression, there were many bugs, but it was greatly improved by using the `terraform/lang` package. See #204 #160 
- Stop overwriting the config under the current directory by the config under the homedir #224 
  - Fixed the problem that overwrites the config under the current directory by homedir config.
- Improve to check for `aws_db_instance_readable_password` #233 
  - Previously, false positive occurred when setting values files or environment variables, but this problem has been fixed.

# ToDo

- [x] Introduce `Runner` that use `terraform.BuiltinEvalContext`
  - #211 #227 #228 #232 
- [x] Re-implement each detectors without `reflect` package
  - #230 #233 #235 #238 #239 #240 #242 #246 
- [x] Re-implement loader with Terraform's `configload` package
  - #217 
- [x] Handle resources and modules with Terraform's `configs` package
  - #219 #220 #223  #224 #229 
- [x] Test Terraform behavior
  - [x] Multiple block declaration. See #238 
  - [x] Map variable merging. See #241


yardensachs:
👌

hoeg:
Any idea about when this is going to be merged?

wata727:
@hoeg All tasks are already finished. However, these features depend on **unstable** version of Terraform v0.12. Thus, I can't merge this pull request until stable version is released...

howdoicomputer:
https://www.hashicorp.com/blog/announcing-terraform-0-1-2-beta1

PR #237, opened 2018-09-14
Add "features" word to docs for people explicitly looking

### LINKED ISSUE #236: List of features in README (or somewhere)
There is no easy way to see what this project do and do not do/check. Like what plugins exist if this were eslint. A bullet-point list would be helpful.

(I was specifically looking for inline JSON linting, as it's hard to not mess JSON up).

odinho:
Fixes #236.

Might not be how you'd want this fixed, but I looked through with only eyes for features, and this would have helped (at least since I also did `CTRL+F`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ template.tf
 Result: 2 issues  (1 errors , 0 warnings , 1 notices)
 ```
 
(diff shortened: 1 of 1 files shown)


wata727:
Thanks!

Issue #248, opened 2018-10-30
Disable module inspection by default


wata727:
Currently, I take first user experience seriously. I would like you to get a good experience without causing errors as much as possible.

However, the current module inspection is premised on `terraform init`, and when you write a module, you always need to run `terraform init` at first. It seems to be inconvenient when using an editor integration.

What should we do to prevent this? I guess that we should disable the module inspection by default (like Deep check mode). I will not start this work for a while, but I will keep thinking.

wata727:
Probably, it also relates to #198.

wata727:
v0.9.0 has been released.

Issue #247, opened 2018-10-30
Revise json and checkstyle report format


wata727:
The current json report format is as follows, but there are some problems:

```json
[
  {
    "detector": "aws_instance_invalid_type",
    "type": "ERROR",
    "message": "\"t2.mini\" is invalid instance type.",
    "line": 5,
    "file": "test.tf",
    "link": "https://github.com/wata727/tflint/blob/master/docs/aws_instance_invalid_type.md"
  }
]
```

- Terminology: `detector` will be renamed to `rule`.
- It points out a line, not a range.

In order to solve these problems, I'm planning to change the format as follows:

```diff
 [
   {
-    "detector": "aws_instance_invalid_type",
+    "rule": "aws_instance_invalid_type",
     "type": "ERROR",
     "message": "\"t2.mini\" is invalid instance type.",
-    "line": 5,
     "file": "test.tf",
+    "position": {
+      "start_line": 5,
+      "end_line": 5,
+      "start_column": 20,
+      "end_column": 41
+    },
     "link": "https://github.com/wata727/tflint/blob/master/docs/aws_instance_invalid_type.md"
   }
 ]
```

I believe that adding ranges is very beneficial, especially with editor integration. However, please note that this is a breaking change for third-party developers.

wata727:
Finally, these were changed as follows:

## JSON

```diff
-[
-  {
-    "detector": "aws_instance_invalid_type",
-    "type": "ERROR",
-    "message": "instance_type is not a valid value",
-    "line": 3,
-    "file": "template.tf",
-    "link": ""
-  }
-]
+{
+  "issues": [
+    {
+      "rule": {
+        "name": "aws_instance_invalid_type",
+        "severity": "error",
+        "link": ""
+      },
+      "message": "instance_type is not a valid value",
+      "range": {
+        "filename": "template.tf",
+        "start": {
+          "line": 3,
+          "column": 19
+        },
+        "end": {
+          "line": 3,
+          "column": 31
+        }
+      },
+      "callers": []
+    }
+  ],
+  "errors": []
+}
```

## Checkstyle

```diff
 <?xml version="1.0" encoding="UTF-8"?>
 <checkstyle>
   <file name="template.tf">
-    <error detector="aws_instance_invalid_type" line="3" severity="error" message="instance_type is not a valid value" link=""></error>
+    <error rule="aws_instance_invalid_type" line="3" column="19" severity="error" message="instance_type is not a valid value" link=""></error>
   </file>
[...]

PR #255, opened 2018-12-09
WORK IN PROGRESS: A simple, unfinished implementation of a plugin system.

howdoicomputer:
I've started on https://github.com/wata727/tflint/issues/121 and this is where I am at currently:

Utilizing Hashicorp's [go-plugin](https://github.com/hashicorp/go-plugin) library, I've put together an incredibly basic plugin system for tflint as well as a basic, [example plugin](https://github.com/howdoicomputer/customrules).

The go-plugin library is a wrapper around the `net/rpc` package that provides convenience functions for establishing a client/host relationship between a process and its associated plugins. I picked it because it made sense to me to stay within the Hashicorp ecosystem and its example implementations were small and easy to read.

The starting point for implementing a basic plugin using the library is to first define a shared interface that is used by both the host and the plugin. In this PR, I'm using the `RuleCollection` interface that is fulfilled by the `Process` method. After the plugin is compiled into a binary, its life-cycle is controlled by the host process. When that plugin binary is executed by the host, it binds a unix socket to expose the implemented interface that is commonly shared. The host then has access to that method.

With that context established, this PR would seek to establish the precedent that custom linting rules would exist in a compiled binary that tflint would talk to over rpc.

In order to accomplish that, I've broken apart the `cli.Run` function into several smaller functions so that tflint can be used as a library that can be imported by plugins. The plugin would then receive a list of Terraform files from the host process and then execute *what used to be* `cli.Run` and return a slice of `issue.Issue`s back to the host process to be printed out.

---

Given the [example plugin](https://github.com/howdoicomputer/customrules) - which is a plugin that has a custom rule that tests to see if an AWS security group is named 'jim' - below is an example of returned output.

Terraform code:

```
> cat main.tf
resource "aws_security_group" "foobar" {
        name = "jim"
}

resource "aws_instance" "foobar" {
  instance_type = "t2.foobar"
}
```

Output:

```
> tflint main.tf
main.tf
        ERROR:2 "aws_security_group" is named jim. (aws_security_group_cant_be_named_jim)
        ERROR:6 "t2.foobar" is invalid instance type. (aws_instance_invalid_type)

Result: 2 issues  (2 errors , 0 warnings , 0 notices)
```

In this situation, the `aws_instance_invalid_type` rule violation is returned by the main process and the `aws_security_group_cant_be_named_jim` is returned by the plugin.

---

This is untested and messy code. I need to do several tasks before I feel like this would be ready for a 'real' review.

Here is a rough checklist of stuff I will be working on as part of this PR:

- [ ] Tests.
- [ ] Additional logic needs to be written to better separate rule outputs coming from a plugin and outputs coming from the host.
- [ ] Function names and variable names should be changed to better clarify intent.
- [ ] Internal documentation needs to be written.
- [ ] Documentation needs to be written on how to write a plugin.
- [ ] There should be an environment variable that points tflint to a plugin directory *or* plugins need to follow a naming convention so that tflint can find a plugin from `$GOBIN`.
- [ ] Cleanups here and there.
- [ ] Fix all the tests that depended on the previous implementation.

However, this seems to work fairly well and could be a solution for not only supporting organization specific rule sets but also for supporting different providers. Do you feel like the direction of this feature implementation would be acceptable for tflint, @wata727?


wata727:
@howdoicomputer Thanks for your awesome contribution! It seems reasonable to adopt go-plugin as the plugin system. I agree with this direction of implementation.

By the way, I would like to minimize the dependency on TFLint API from plugins. Currently, this implementation seems to depend on `cli.NewCLI`, `SanityCheck ` and `ProcessRules`. Is it possible to implement such a plugin that only has an API like [`NewRules`](https://github.com/wata727/tflint/blob/next_tflint_dev/rules/provider.go#L64)?
Since I'm not familiar with rpc, I'm not sure if that is feasible...

FYI: In the next release, I will make big internal system changes, so I would like to reduce new features as possible. Probably this feature will be included after next release.

Also, I'm planning to breaking change TFLint's internal API. See https://github.com/wata727/tflint/issues/247 https://github.com/wata727/tflint/issues/248. As a result, I think that this plugin system will be broken periodically. Therefore, I would like to release it as an experimental feature for a while.

howdoicomputer:
> By the way, I would like to minimize the dependency on TFLint API from plugins. Currently, this implementation seems to depend on cli.NewCLI, SanityCheck and ProcessRules. Is it possible to implement such a plugin that only has an API like NewRules?
Since I'm not familiar with rpc, I'm not sure if that is feasible...

This should be possible and what I'll work on next. It makes more sense. I feel a bit more comfortable with `go-plugin` now that I got something working so it'll be easy to adapt.

> FYI: In the next release, I will make big internal system changes, so I would like to reduce new features as possible. Probably this feature will be included after next release.

> Also, I'm planning to breaking change TFLint's internal API. See #247 #248. As a result, I think that this plugin system will be broken periodically. Therefore, I would like to release it as an experimental feature for a while.

Yeah, this is why I was basing my work off of the `next_tflint_dev` branch. I'll try and keep pace by submitting plugin related PRs as changes develop. Once plugin support is released, I was imagining that it'd be a very 'do it yourself and at your own peril' for tflint users.

wata727:
@howdoicomputer ping :)

TFLint v0.8 has been released 1 month ago. What is the status of this pull request?
This work seems to be of interest to many users. Thank you again for the awesome work.

stephencroberts:
@howdoicomputer I'm very interested in this feature. Anything I can do to help?

PR #260, opened 2019-02-18
Change dockerfile design

inductor:
I'd like to propose this change for the following reasons.

1. ENTRYPOINT is hardly changeable, and it is not good when you use the Docker Image with CI tools since they require an actual command along with an Image.

2. Unnecessary packages such as ca-certificates should not be in the artifact so that I removed it with using Multi-stage Build.

3. I think this is all about your preference, but in most cases you compile a binary within a Dockerfile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ $ brew install tflint
 We provide Docker images for each version on [DockerHub](https://hub.docker.com/r/wata727/tflint/). With docker, you can run TFLint without installing it locally.
 
 ```
-$ docker run --rm -v $(pwd):/data -t wata727/tflint
+$ docker run -itv (pwd):/project -w /project --rm wata727/tflint tflint
 ```
 
 ## Quick Start
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,23 @@
-FROM alpine:3.5
+FROM golang:1.11.5-alpine3.9 as builder
 
 MAINTAINER wata727
 
-RUN apk add --no-cache ca-certificates
+WORKDIR /root
 
-COPY dist/linux_amd64/tflint /usr/local/bin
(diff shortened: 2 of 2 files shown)


wata727:
First of all, thanks for your contribution!

> ENTRYPOINT is hardly changeable, and it is not good when you use the Docker Image with CI tools since they require an actual command along with an Image.

In my opinion, ENTRYPOINT is better than CMD because users don't need to run binaries other than`tflint` in the Docker image. Currently, I don't assume to run other commands in this image.

> Unnecessary packages such as ca-certificates should not be in the artifact so that I removed it with using Multi-stage Build.

Multi-stage build is a good way 👍 But `ca-certificates` is required. See https://github.com/wata727/tflint/pull/155

> I think this is all about your preference, but in most cases you compile a binary within a Dockerfile.

👏

README.md
```diff
@@ -54,7 +54,7 @@ $ brew install tflint
 We provide Docker images for each version on [DockerHub](https://hub.docker.com/r/wata727/tflint/). With docker, you can run TFLint without installing it locally.
 
 ```
-$ docker run --rm -v $(pwd):/data -t wata727/tflint
+$ docker run -itv (pwd):/project -w /project --rm wata727/tflint tflint
```
wata727: I would like to keep this command as much as possible for users using the `latest` tag.

inductor:
Alright...

Any idea to use this image with CircleCI? :(

wata727:
I'm not familiar with CircleCI, but it seems to work with the following configuration.

```yaml
version: 2
jobs:
  build:
    docker:
      - image: wata727/tflint:0.7.4
    steps:
      - checkout
      - run: tflint --error-with-issues
```

PR #267, opened 2019-03-30
automate the installation of tflint on linux

tchia04:
Hopefully, this will address issue#263

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ $ install tflint /usr/local/tflint/bin
 $ tflint -v
 ```
 
+For Linux based OS, you can use the install_linux.sh to automate the installation process
+
 ### Homebrew
 
 macOS users can also use [Homebrew](https://brew.sh) to install TFLint:
--- a/install_linux.sh
+++ b/install_linux.sh
@@ -1,0 +1,59 @@
+#!/bin/bash -e
+
+processor=$(uname -m)
+
+if [ "$processor" == "x86_64" ]; then
+  arch="amd64"
+else
+  arch="386"
+fi
+
+case "$(uname -s)" in
+  Darwin*)
+    os="darwin_${arch}"
+    ;;
+  MINGW64*)
+    os="windows_${arch}"
+    ;;
+  MSYS_NT*)
+    os="windows_${arch}"
+    ;;
+  *)
+    os="linux_${arch}"
+    ;;
+esac
+
+echo "os=$os"
+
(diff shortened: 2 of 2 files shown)


wata727:
Thanks for the PR, but it does not seem to work correctly in my environment...

```
$  curl https://raw.githubusercontent.com/tchia04/tflint/install_linux/install_linux.sh | bash                                                                                          [0:25:15]
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1398  100  1398    0     0   8630      0 --:--:-- --:--:-- --:--:--  8683
os=darwin_amd64


Downloading latest version of tflint which is v0.7.4
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   613    0   613    0     0    990      0 --:--:-- --:--:-- --:--:--   990
100 10.6M  100 10.6M    0     0  2606k      0  0:00:04  0:00:04 --:--:-- 3603k
Download was successfully


Unpacking and Installing tflint ...
Archive:  /tmp/tflint.zip
install: /tmp/tflint: Inappropriate file type or format
```

tchia04:
When I tested this on another mac without /usr/local/bin directory, I got a "permission denied" error. So I added the "mkdir -p /usr/local/bin" to make sure the directory exist first. Also I have added sudo to the install command. Hopefully that will resolve your issue.

wata727:
After deleting `tmp/tflint*`, I confirmed that it works well. However, I'm not sure if we should use `sudo` in such installation scripts. Do you have any other good ideas? Also, is it common for `sudo` to be used in such scripts?

tchiaspko:
On the Mac, I think usually the /usr/local/ folder is owned by one user so we probably won't need sudo. 

However in other flavors of Linux . the /usr/local/... is usually owned by root and not by a particular user.

tchia04:
Yes installing to the /usr/local/bin folder usually requires root permission.  
I chose this location because usually this folder is already in the PATH and doesn't require update the PATH in .bashrc or .bash_profile file

One solution to not use sudo is to copy the executable to ~/bin/ folder instead of /usr/local/bin/ but this may require setting up the PATH to include ~/bin which the script doesn't checks for

i.e. Need something like this in ~/.bash_profile
export PATH="$HOME/bin:$PATH"

wata727:
Thanks!

PR #280, opened 2019-05-19
Bring Alpine current in Dockerfile

gliptak:


wata727:
This patch is merged via https://github.com/wata727/tflint/pull/281

PR #281, opened 2019-05-21
Rewrite Dockerfile multi-stage

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,20 @@
-FROM alpine:3.5
+FROM golang:1.12-alpine as builder
 
-MAINTAINER wata727
+RUN apk --no-cache add git make gcc musl-dev zip
+
+WORKDIR /go/src/github.com/wata727/tflint/
+
+ADD . /go/src/github.com/wata727/tflint
+
+RUN make build
+
+FROM alpine:3.9 as prod
+
+LABEL maintainer=wata727
 
 RUN apk add --no-cache ca-certificates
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,10 +1,22 @@
...
+FROM golang:1.12-alpine as builder
 
-MAINTAINER wata727
+RUN apk --no-cache add git make gcc musl-dev zip
+
+WORKDIR /go/src/github.com/wata727/tflint/
+
+ADD . /go/src/github.com/wata727/tflint
+
+RUN make release
```
wata727: `make release` builds binaries for all architectures. It might be better to just run `make build` for the linux platform.

Dockerfile
```diff
@@ -1,10 +1,22 @@
...
+RUN find /go/src/github.com/wata727/tflint/dist/
+
+FROM alpine:3.9 as prod
+
+LABEL maintainer=wata727
 
 RUN apk add --no-cache ca-certificates
 
-COPY dist/linux_amd64/tflint /usr/local/bin
+COPY --from=0 /go/src/github.com/wata727/tflint/dist/linux_amd64/tflint /usr/local/bin
```
wata727: I guess that `--from=builder` is more easy to understand.

gliptak: updated

wata727:
@gliptak Thank you!

PR #319, opened 2019-06-07
Added pre-commit hooks

krzyzakp:
For now added only gofmt, which is required. Worth to check other
possibilities under: https://github.com/dnephin/pre-commit-golang

Put into Developing documentation also info about it.


wata727:
👍

PR #322, opened 2019-06-09
Add new rule: aws_s3_bucket_invalid_acl

ineffyble:
Add a new rule to check if the `acl` value specified for an `aws_s3_bucket` matches one of the [canned ACL types](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl).


wata727:
Awesome! Thank you!

PR #334, opened 2019-06-24
Check mapping attribute types

wata727:
Automatic rule generations from mappings are introduced in #274. We introduce automatically mapping's type checking in this PR.

terraform-provider-aws exposes its schemas, so the rule generator uses this to check types. If a wrong type is mapped, panic will occur and the process will be stopped.


rules/awsrules/models/mappings/kinesis-firehose.hcl
```diff
@@ -10,8 +10,8 @@ mapping "aws_kinesis_firehose_delivery_stream" {
...
   elasticsearch_configuration  = ElasticsearchDestinationConfiguration
   splunk_configuration         = SplunkDestinationConfiguration
-  cloudwatch_logging_options   = CloudWatchLoggingOptions
-  processing_configuration     = ProcessingConfiguration
-  processors                   = ProcessorList
-  parameters                   = ProcessorParameterList
+  // cloudwatch_logging_options   = CloudWatchLoggingOptions
+  // processing_configuration     = ProcessingConfiguration
+  // processors                   = ProcessorList
+  // parameters                   = ProcessorParameterList
```
wata727: Some resources seem to provide schemas in another way. Here, I will comment out and update it later.

wata727:
I'm now encountered with Bazaar's issue... This causes the build on Windows to fail.
Unfortunately, there seems to be no way to get around this. I'm considering if we can remove the dependency that needs bzr.

PR #341, opened 2019-06-26
Fix false negatives in the S3 invalid ACL rule

wata727:
This PR replaces `aws_s3_bucket_invalid_acl` rule with the auto-generated rule introduced in #274.

As a result, it also fixes false negatives in the rule. The previous list included object ACLs such as `bucket-owner-read`. These are invalid values for bucket ACL. See https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl


rules/awsrules/models/aws_s3_bucket_invalid_acl.go
```diff
@@ -1,35 +1,32 @@
...
 		attributeName: "acl",
-		aclTypes: map[string]bool{
-			"private":                   true,
-			"public-read":               true,
-			"public-read-write":         true,
-			"aws-exec-read":             true,
-			"authenticated-read":        true,
-			"bucket-owner-read":         true,
-			"bucket-owner-full-control": true,
-			"log-delivery-write":        true,
```
ricoli: why was this one removed though? It seems perfectly valid to me?

ricoli: In fact, all of the above are valid according to the same link you've referenced in the PR description (https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl), so I don't quite understand the change?

wata727: Umm, what you are saying seems right. It's weird...
The list is generated from [aws-sdk API definition](https://github.com/aws/aws-sdk-go/blob/v1.20.4/models/apis/s3/2006-03-01/api-2.json#L1071-L1079), so It may be something wrong with how to use. I will investigate that.

ricoli: Looks like that block is way out of date - hasn't been updated in 5 years...

ricoli: There were commits to add new Canned ACLs since then (https://github.com/aws/aws-sdk-go/commit/0922c232017dc730e908d7528741ceed9d30c796 for example), but that particular block was never updated :(

ricoli: https://github.com/aws/aws-sdk-go/issues/2683

PR #368, opened 2019-07-21
Update brew instructions

arbourd:
Homebrew is now building and packaging tflint itself:
https://github.com/Homebrew/homebrew-core/blob/master/Formula/tflint.rb


wata727:
👏

PR #379, opened 2019-08-03
Remove aws_db_instance_readable_password rule

wata727:
In my current opinion, the `aws_db_instance_readable_password` rule makes no sense.

For example, the following way is recommended to handle sensitive values:

> We don't recommend saving usernames and password to version control, but you can create a local secret variables file and use -var-file to load it.

https://learn.hashicorp.com/terraform/getting-started/variables.html#from-a-file

However, the `aws_db_instance_readable_password` loads the local secret variable file even if the file is outside of version control. As a result, the rule reports a warning. This is wrong. 

In addition to this, current TFLint cannot accurately determine whether sensitive values are handled correctly. So, remove this rule to avoid confusing users.


PatMyron:
related discussions of sensitive properties that shouldn't be hardcoded if we ever figure out how to re-write this rule:
aws-cloudformation/cfn-lint#1181
stelligent/cfn_nag#253

PR #431, opened 2019-09-09
Disable CGO in GoReleaser

craigfurman:
To produce statically linked binaries even when not cross-compiling.

Related to https://github.com/wata727/tflint/issues/430.


wata727:
Thank you @craigfurman.
I confirmed this problem. And this fix seems to make sense. I'm going to release the next patch version as soon as possible.

PR #435, opened 2019-09-13
Add Linix oneliner to download latest `tflint`

abitrolly:


wata727:
Why we need the oneliner script? Is `install_linux.sh` not enough?
I think it would be better to keep the README as simple as possible...

abitrolly:
Onliner has an advantage that it can be copy-pasted into terminal and recorded in shell history in case the program needs to be updated again. With `install_linux.sh` I need to keep it around or download every time.

wata727:
OK, it makes sense. How about the following oneliner script instead?

```sh
$ curl https://raw.githubusercontent.com/wata727/tflint/master/install_linux.sh | sh
```

abitrolly:
I personally still prefer to understand what a script does. Nevertheless that's also a good solution for my user story.

wata727:
I personally prefer not this oneliner, but I respect your opinion. If we get any other feedback, I will reconsider it again.

wata727:
I personally prefer not this oneliner, but I respect your opinion. If we get any other feedback, I will reconsider it again.

PR #437, opened 2019-09-15
Fix typo in dash rule

abitrolly:
This fixes unclear wording.

On a side note, the rationale to avoid dashes in names is because they make expressions `${var.instance-count-1}` unclear - it it hard to say if is subtraction `${var.instance-count - 1}` or just a reference to `instance-count-1` variable.

https://www.terraform.io/docs/configuration-0-11/interpolation.html#math


wata727:
👍

PR #439, opened 2019-09-15
Experimental language server support

### LINKED ISSUE #275: LSP support
[Language Server Protocol](https://microsoft.github.io/language-server-protocol/) is a good way to integrate with editors. I believe that we need LSP support to provide a more comfortable editor experience.

The big challenge is cache handling. We need to invalidate the API responses cache properly, and to control the size of the cache properly. Probably, it is better to use [`hashicorp/golang-lru`](https://github.com/hashicorp/golang-lru) etc.

wata727:
Fixes #275

This PR introduces a new option `tflint --langserver`. This server communicates with a client via stdio according to the specification of [Language Server Protocol](https://microsoft.github.io/language-server-protocol/specification). In this experimental implementation, only `textDocument/publishDiagnostics` is supported.

You can try the server immediately with [A PoC of the language server client](https://github.com/wata727/vscode-tflint-lsc):

lsp_demo

This client is a PoC and will not be published. In the future, we hope to support this LSP server with vscode-terraform plugin etc.


FernandoMiguel:
wonder if these two issues can be tracked together ? 
https://github.com/hashicorp/vscode-terraform/issues/337

PR #442, opened 2019-09-17
aws_s3_bucket_invalid_region_rule: Fix false positives

### LINKED ISSUE #441: Consider removing/changing aws_s3_bucket_invalid_region check
I just tried to do the first linting on my project, and got an error:
```
Error: region is not a valid value (aws_s3_bucket_invalid_region)
  on x.tf line 202:
 202:   region   = "us-east-2"
```
But the bucket is definitely there, and this error seems to be a false-positive.

Digging into this a bit I found that:
- This seems to be coming from https://github.com/wata727/tflint/blob/283037d25804adfd3032e3857121d8e0cd6e98d3/rules/awsrules/models/aws_s3_bucket_invalid_region.go
- Which in turn generates that list from a constant named `BucketLocationConstraint`
- Which is mentioned in the aws-sdk-cpp as outdated in 2015: https://github.com/aws/aws-sdk-cpp/issues/48
- In addition I can't seem to find any AWS region that doesn't have S3 at the moment: https://aws.amazon.com/about-aws/global-infrastructure/regional-product-services/

Therefore I'd propose either dropping this check altogether, or updating it to use some other constant to determine whether a region is valid.

wata727:
Fixes #441 

Unfortunately, `BucketLocationConstraint` is also outdated. This PR switches the rule into manual maintained one, updates the valid region list.

See also https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region


rules/awsrules/aws_s3_bucket_invalid_region.go
```diff
@@ -23,16 +21,27 @@ func NewAwsS3BucketInvalidRegionRule() *AwsS3BucketInvalidRegionRule {
 		attributeName: "region",
 		enum: []string{
 			"EU",
```
wata727: If you use EU, terraform will fail when reading the resource, but the creation will succeed. Although it should not be used newly, EU may already exist as a valid value.
This means to allow false negatives, but it makes sense here.

Issue #436, opened 2019-09-13
Detect invalid variable/expression usage with missing curly braces


abitrolly:
Currently, `tflint` is unable to find cases when people forget to use `{` `}` to render variables.
```
  triggers = {
    template_rendered = "$data.template_file.inventory_template.rendered"
  }
```
This code is wrong, because `"$data.template_file.inventory_template.rendered"` is treated as literal string, and it should be written as "`${data.template_file.inventory_template.rendered}`".

* [Invalid Terraform report](https://github.com/hashicorp/terraform/issues/21211#issuecomment-510703355)
* [Original bad code](https://github.com/abitrolly/XE3Bot/commit/ee4cd4b7e3fa9a61ac0f3c030146fee39f389929#diff-0e98d84c23e702e3baa05095730661b0)

wata727:
Umm, this seems a little difficult.
First of all, it's not clear what the invalid variable/expression is. For instance, the above example is clearly a valid string literal. Is the following expression invalid? Or is it a string literal?

```hcl
value = "This is var.number colors"
```

```hcl
value = "This is $var.number colors"
```

Although we may be able to create regular expressions that can detect them well, we may need to be very careful with the implementation.

abitrolly:
Does `tflint` maintain the list of variables found in `.tf` script? It could match $ regexp against that list.

wata727:
Yes, TFLint has a list of variables and resources as a result of parsing.

https://github.com/wata727/tflint/blob/1c31b742deb0a556520dec70ae9d4da5740f2b48/tflint/runner.go#L26

https://github.com/hashicorp/terraform/blob/e97ad4ebef504883229a1c299be617dfb7bec017/configs/module.go#L32-L39

abitrolly:
I understand that there should be a file in https://github.com/wata727/tflint/tree/master/rules/terraformrules that traverses all expanded string literals, then gets all substrings matching the regexp `\$\w[.\w]` and then check those list against the list of variables and give a warning. Is that right?

wata727:
That's right. I think that it is the most difficult to implement a parser to search for "reference-like" from a string literal.

abitrolly:
The parser is just a simple `regex` matching, no? https://play.golang.org/p/Nu4ZapS272N

wata727:
Umm, probably it will work. However, I don't fully understand the definition of HCL expressions, so I can't ensure that this regular expression is perfectly correct.
[...]

PR #469, opened 2019-10-02
Rule for checking missing braces for variable references (#436)

abitrolly:
The following template reference is invalid, is treated as
a literal string and not expanded.

    resource "null_resource" "generate_inventory" {
      triggers = {
        template_rendered = "$data.template_file.inventory_template.rendered"
      }
    ...


bendrucker:
Hi there:

* This style of interpolation is deprecated in Terraform 0.12
* It's going to be fairly difficult to avoid false positives

We've added some methods to the runner recently that would help with rules like this. If you're interested in pursuing further happy to discuss how viable possible approaches might be.

abitrolly:
So what is the 0.12 alternative?

bendrucker:
0.12 will print a warning for "${data.template_file.inventory_template.rendered}" and Terraform 0.13  will error. TFLint can enforce this for you now:

https://github.com/terraform-linters/tflint/blob/master/docs/rules/terraform_deprecated_interpolation.md

The quotes and interpolation is unnecessary, simply `data.template_file.inventory_template.rendered` will do. 

This problem obviously still exists if you do this:

```
"foo-$data.template_file.inventory_template.rendered"
```

However, I'll warn you that the difficulty of detecting this accurately is pretty high.

abitrolly:
It evolves faster than me. )

PR #490, opened 2019-10-13
Detect invalid rule name in command line option

### LINKED ISSUE #470: No error on `tflint --enable-rule blablabla`
```
✗ ./tflint --enable-rule blablabla                         
✗ echo $?                         
0
```

:/  I expected some error if `tflint` can not find the rule (and `--debug` key that makes it report what rules are found).

abitrolly:
This adds INFO level stats about total amount of rules and enabled
rules, makes `defaultRules` private, because they are not accessed
from anywhere. It also shows ERROR level message for invalid name
specified for --enable-rule or --disable-rule options.

The missing thing is to stop processing and with the error at this
point.

Fixes #470.

### DIFF
--- a/rules/provider.go
+++ b/rules/provider.go
@@ -43,7 +43,7 @@ var manualDeepCheckRules = []Rule{
 	awsrules.NewAwsLaunchConfigurationInvalidImageIDRule(),
 }
 
-// AllRules returns map of rules indexed by name
+// AllRulesMap returns map of rules indexed by name
 func AllRulesMap() map[string]Rule {
 	ret := map[string]Rule{}
 	for _, rule := range append(DefaultRules, deepCheckRules...) {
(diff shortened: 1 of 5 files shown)


abitrolly:
@wata727 this is now ready for review.
```
$ tflint --enable-rule=blablabla                      
Failed to prepare rules. An error occurred:

Error: Rule not found: blablabla
```
Right now I am unable to write test for that case. I am not familiar with testing in Go yet, and it may take a while before learn that.

rules/provider.go
```diff
@@ -43,11 +44,35 @@ var manualDeepCheckRules = []Rule{
...
+	ret := map[string]Rule{}
+	for _, rule := range append(DefaultRules, deepCheckRules...) {
+		ret[rule.Name()] = rule
+	}
+	return ret
+}
+
 // NewRules returns rules according to configuration
-func NewRules(c *tflint.Config) []Rule {
+func NewRules(c *tflint.Config) ([]Rule, error) {
```
wata727: Sorry, I'm planning to make a major change to the implementation of this provider, so I don't encourage you to change `NewRules`. Perhaps we should implement a method that simply returns a list of rule names.

This allows for the possibility that the Provider will be provided by the user as a plug-in in the future. Since `NewRules` does not return all rules, we cannot check whether the passed rule name exists.

wata727: See #491

abitrolly: I almost finished [error processing](https://github.com/golang/go/wiki/ErrorValueFAQ) code. :D

Makes total sense to check options separately before `NewRules` is called. Now that I know the codebase, I can try to do this.

abitrolly:
I will refactor and open a new PR.

abitrolly:
This is now in #496.

PR #491, opened 2019-10-14
Introduce RuleSet

wata727:
This pull request introduces RuleSet to implement the plug-in system.

Each plug-in serves a RuleSet that implements `Check` and returns Issues via RPC. This is the pre-stage of the plug-in system implementation, and there is no change in functionality.


cmd/cli.go
```diff
@@ -160,14 +160,14 @@ func (cli *CLI) Run(args []string) int {
 
 	issues := tflint.Issues{}
 
-	provider := rules.NewProvider()
+	ruleset := rules.NewRuleSet()
 	for _, runner := range runners {
-		providerIssues, err := provider.Check(runner)
+		ret, err := ruleset.Check(runner)
```
abitrolly: I would keep `issues` in the name of returned variable as it makes code more readable.

tflint/runner.go
```diff
@@ -205,6 +207,32 @@ func NewModuleRunners(parent *Runner) ([]*Runner, error) {
...
+			log.Printf("[DEBUG] `%s` is disabled", rule.Name())
+		}
+		enabled = r.Enabled
+	}
+
+	if enabled {
+		return rule.Check(r)
+	}
+	return nil
+}
```
abitrolly: From one side I don't like that rules depend on Config, which is passed through Runner. From the other side, in the end it is easy to track who Enables and Disables rules.

wata727:
Thank you for reviewing. But now I'm freezing this pull request because of a plugin system issue.
First, I confirmed that go-plugin has the following restrictions:

- We cannot pass `Runner` to a plugin via RPC
- A plugin cannot serve functions/interfaces (`Rule` is an interface)

The latter had already confirmed, but it was unexpected that the former could not do. This refactoring is for the former.

Based on these, it was necessary to reconsider the architecture of the plug-in system. The result will determine whether this pull request will be merged.

wata727:
Closed via #500

PR #496, opened 2019-10-17
Check invalid rule names in command line options (fixes #470)

### LINKED ISSUE #470: No error on `tflint --enable-rule blablabla`

abitrolly:
Exit with an error when --enable-rule name is not found.

Also adds INFO level stats about total amount of rules and enabled rules.

### DIFF
--- a/cmd/cli.go
+++ b/cmd/cli.go
@@ -158,6 +158,16 @@ func (cli *CLI) Run(args []string) int {
 	}
 	runners = append(runners, runner)
 
+	var ruleNames []string
+	for name := range cfg.Rules {
+		ruleNames = append(ruleNames, name)
+	}
+	err = rules.CheckRuleNames(ruleNames)
+	if err != nil {
+		formatter.Print(tflint.Issues{}, tflint.NewContextError("Failed to check rule config", err), cli.loader.Sources())
+		return ExitCodeError
+	}
+
 	for _, rule := range rules.NewRules(cfg) {
 		for _, runner := range runners {
 			err := rule.Check(runner)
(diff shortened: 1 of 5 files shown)


cmd/cli.go
```diff
@@ -158,6 +158,16 @@ func (cli *CLI) Run(args []string) int {
...
 	runners = append(runners, runner)
 
+	var ruleNames []string
+	for name := range cfg.Rules {
+		ruleNames = append(ruleNames, name)
+	}
+	err = rules.CheckRuleNames(ruleNames)
+	if err != nil {
+		formatter.Print(tflint.Issues{}, tflint.NewContextError("Failed to check rule config", err), cli.loader.Sources())
+		return ExitCodeError
```
wata727: It would be better to add a test case for the error like the following:
https://github.com/wata727/tflint/blob/96dd0d36f38515b8def4d4ccb37ff31da28bcb7c/cmd/cli_test.go#L97-L102

cmd/cli.go
```diff
@@ -158,6 +158,16 @@ func (cli *CLI) Run(args []string) int {
 	}
 	runners = append(runners, runner)
 
+	var ruleNames []string
+	for name := range cfg.Rules {
+		ruleNames = append(ruleNames, name)
+	}
+	err = rules.CheckRuleNames(ruleNames)
```
wata727: I'm happy if an equivalent check will be also performed in LSP mode:
https://github.com/wata727/tflint/blob/96dd0d36f38515b8def4d4ccb37ff31da28bcb7c/langserver/handler.go#L22-L35

cmd/cli.go
abitrolly: Done.

cmd/cli.go
abitrolly: Also done.

How these options are set for LSP?

wata727: These can be set via `.tflint.hcl` file.

PR #503, opened 2019-10-25
add note about recursive check

IslamAzab:


README.md
```diff
@@ -21,6 +21,10 @@ resource "aws_instance" "foo" {
 
 Since `t1.2xlarge` is a nonexistent instance type, an error will occur when you run `terraform apply`. But `terraform plan` and `terraform validate` cannot find this possible error beforehand. That's because it's an AWS provider-specific issue and it's valid as a Terraform configuration.
 
+**P.S.**
+TFLint always checks only the current root module (no recursive check).
+For more info please see: https://github.com/wata727/tflint/issues/493 and https://github.com/wata727/tflint/issues/502
+
```
wata727: I think it would be nice to add FAQ section like at the bottom. What do you think?

IslamAzab: Okay I will update it then with my question. Let me know if you want to add any more questions.

README.md
```diff
@@ -116,6 +120,10 @@ TFLint returns the following exit statuses on exit:
 - 2: Errors occurred
 - 3: No errors occurred, but issues found
 
+## FAQ
+- Does TFLint check modules recursively?
+- No. FLint always checks only the current root module (no recursive check)
```
wata727: How about adjusting the following layouts?

```
### Does TFLint check modules recursively?
No. TFLint always checks only the current root module (no recursive check)
```

スクリーンショット 2019-10-28 23 18 40

README.md
wata727: Thanks! In that case, I think that we don't need to add this line.

README.md
```diff
@@ -116,6 +116,10 @@ TFLint returns the following exit statuses on exit:
 - 2: Errors occurred
 - 3: No errors occurred, but issues found
 
+## FAQ
+### Does TFLint check modules recursively?
+- No. FLint always checks only the current root module (no recursive check)
```
wata727:
```suggestion
- No. TFLint always checks only the current root module (no recursive check)
```

wata727:
Thank you!

Issue #513, opened 2019-11-09
Add a check that modules are pinned to semantic versions


alexwlchan:
terraform_module_pinned_source does some checks that a module source isn’t ambiguous, but it could be stronger.

Terraform encourages us to use semantic versioning, right? So how about this check:

**Warn if a module is not pinned to a semantic version vX.Y.Z.**

This is stronger than the existing rule:

* Covers all module sources (except local path)
* Avoids ambiguous references in Git/Mercurial. Pinning to a non-default branch will pass the existing check, but might still be ambiguous.
* Gets easier to reason about references. I can use a commit hash as an unambiguous Git ref (`?ref=0ad21e6`), but it’s hard to know if that’s current or outdated. Forcing me to use a semver references fixes that.

The latter reason is my primary interest in the feature.

I’d be happy to have a first attempt at this, if you think it’s a good idea.

If so, do you think it should be a new rule, or a change to the existing rule?

wata727:
Hi @alexwlchan. Thank you for opening the issue.

That makes sense. On the other hand, we may have to consider the case where you cannot use the semantic version for a module (e.g. untagged module).

Based on the above, it is good to create it as another rule or add a new option for `terraform_module_pinned_source` rule. There is currently no way to define per-rule options, so it might be a good idea to create a new rule.

alexwlchan:
Having another rule sounds good (and maybe not one that’s on by default). 👍

I’ll have a play over the weekend and let you know how I get on.

PR #517, opened 2019-11-09
When checking if a source is pinned, allow for Mercurial/Bitbucket

alexwlchan:
Bitbucket repos can be Git or Mercurial (for now, at least!), and Terraform can use either of them as a source.  If you were using a Mercurial Bitbucket repo as your module source, it would be reported as unpinned – even if you were pinning to a non-default Mercurial revision.

---

Spotted while working on #513.

### DIFF
--- a/rules/terraformrules/terraform_module_pinned_source.go
+++ b/rules/terraformrules/terraform_module_pinned_source.go
@@ -56,8 +56,10 @@ func (r *TerraformModulePinnedSourceRule) Check(runner *tflint.Runner) error {
 
 		lower := strings.ToLower(module.SourceAddr)
 
-		if reGithub.MatchString(lower) || reBitbucket.MatchString(lower) || reGenericGit.MatchString(lower) {
+		if reGithub.MatchString(lower) || reGenericGit.MatchString(lower) {
 			r.checkGitSource(runner, module)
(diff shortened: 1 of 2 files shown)


rules/terraformrules/terraform_module_pinned_source.go
```diff
@@ -56,8 +56,10 @@ func (r *TerraformModulePinnedSourceRule) Check(runner *tflint.Runner) error {
 
 		lower := strings.ToLower(module.SourceAddr)
 
-		if reGithub.MatchString(lower) || reBitbucket.MatchString(lower) || reGenericGit.MatchString(lower) {
+		if reGithub.MatchString(lower) || reGenericGit.MatchString(lower) {
 			r.checkGitSource(runner, module)
+		} else if reBitbucket.MatchString(lower) {
+		    r.checkBitbucketSource(runner, module)
```
wata727: FYI, [gofmt](https://golang.org/cmd/gofmt/) is helpful for formatting your code. In many Go projects (including TFLint), it adopts gofmt to avoid style discussions.

alexwlchan: Thanks, done!

wata727:
👍

PR #518, opened 2019-11-09
Add an optional checker for semver versions

### LINKED ISSUE #513: Add a check that modules are pinned to semantic versions

alexwlchan:
Closes #513.

Thanks for providing good developer documentation. Made it really easy to write this patch! 😄

### DIFF
--- a/docs/rules/terraform_module_semver_source.md
+++ b/docs/rules/terraform_module_semver_source.md
@@ -1,0 +1,61 @@
+# terraform_module_semver_source
+
+Disallow specifying a Git or Mercurial repository as a module source without pinning to a semantic version reference.
+
+## Example
+
+```hcl
+module "unpinned" {
+  source = "git://hashicorp.com/consul.git"
+}
+
+module "git_commit_ref" {
+  source = "git://hashicorp.com/consul.git?ref=8afd21a"
+}
+
+module "git_semver_ref" {
+  source = "git://hashicorp.com/consul.git?ref=v1.0.0"
+}
+
+module "mercurial_branch_rev" {
+  source = "hg::http://hashicorp.com/consul.hg?rev=branch"
+}
+
+module "mercurial_semver_rev" {
+  source = "hg::http://hashicorp.com/consul.hg?rev=1.2.3"
+}
+```
+
+```
+$ tflint example.tf --disable-rule=terraform_module_pinned_source --enable-rule=terraform_module_semver_source
+3 issue(s) found:
+
+Warning: Module source "git://hashicorp.com/consul.git" is not pinned (terraform_module_semver_source)
+
+  on example.tf line 2:
+   2:   source = "git://hashicorp.com/consul.git"
+
+Reference: https://github.com/wata727/tflint/blob/v0.12.1/docs/rules/terraform_module_semver_source.md
+
+Warning: Module source "git://hashicorp.com/consul.git?ref=8afd21a" uses a ref which is not a version string (terraform_module_semver_source)
+
+  on example.tf line 6:
+   6:   source = "git://hashicorp.com/consul.git?ref=8afd21a"
+
+Reference: https://github.com/wata727/tflint/blob/v0.12.1/docs/rules/terraform_module_semver_source.md
+
+Warning: Module source "hg::http://hashicorp.com/consul.hg?rev=branch" uses a rev which is not a version string (terraform_module_semver_source)
+
+  on example.tf line 14:
+  14:   source = "hg::http://hashicorp.com/consul.hg?rev=branch"
+
+Reference: https://github.com/wata727/tflint/blob/v0.12.1/docs/rules/terraform_module_semver_source.md
+```
+
+## Why
+
+Terraform allows you to checkout module definitions from source control. If you specify a non-permanent reference (for example, a branch name), the dependency may change without your awareness. To prevent this, always specify an explicit version to checkout.
+
+## How To Fix
+
+Specify a semantic version pin, of the form `vX.Y.Z`.  The leading `v` is optional.
--- a/rules/terraformrules/terraform_module_pinned_source.go
+++ b/rules/terraformrules/terraform_module_pinned_source.go
@@ -42,9 +42,17 @@ func (r *TerraformModulePinnedSourceRule) Link() string {
 	return tflint.ReferenceLink(r.Name())
 }
 
-var reGithub = regexp.MustCompile("(^github.com/(.+)/(.+)$)|(^git@github.com:(.+)/(.+)$)")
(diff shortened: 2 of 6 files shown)


docs/rules/terraform_module_pinned_source_is_semver.md
```diff
@@ -0,0 +1,54 @@
+# terraform_module_pinned_source_is_semver
```
wata727: The naming feels a little redundant. How about `terraform_module_semver_source`?

docs/rules/terraform_module_pinned_source_is_semver.md
```diff
@@ -0,0 +1,54 @@
...
+  source = "git://hashicorp.com/consul.git"
+}
+
+module "git_commit" {
+  source = "git://hashicorp.com/consul.git?ref=8afd21a"
+}
+
+module "mercurial_rev" {
+  source = "hg::http://hashicorp.com/consul.hg?rev=branch"
+}
```
wata727: It would be more helpful to add a good example where no issues are detected.

docs/rules/terraform_module_pinned_source_is_semver.md
wata727: Please also add it to the documentation index :)
https://github.com/terraform-linters/tflint/blob/629009c1430cbb92bcd7b79d7b7150ae6b2b01d2/docs/rules/README.md

alexwlchan:
Thanks for the fast review; I've addressed all the comments (and remembered to run gofmt this time!).

alexwlchan:
Rebased against #517.

docs/rules/terraform_module_semver_source.md
```diff
@@ -0,0 +1,53 @@
...
+  source = "git://hashicorp.com/consul.git"
+}
+
+module "git_commit" {
+  source = "git://hashicorp.com/consul.git?ref=8afd21a"
+}
+
+module "mercurial_rev" {
+  source = "hg::http://hashicorp.com/consul.hg?rev=branch"
+}
```
wata727: It would be nice to add an example like the following:

```hcl
module "pinned_git" {
  source = "github.com/hashicorp/consul.git?ref=v1.2.3"
}
```

docs/rules/terraform_module_semver_source.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+```
+$ ./dist/tflint example.tf --disable-rule=terraform_module_pinned_source --enable-rule=terraform_module_semver_source
```
wata727:
```suggestion
$ tflint example.tf --disable-rule=terraform_module_pinned_source --enable-rule=terraform_module_semver_source
```

rules/terraformrules/terraform_module_pinned_source.go
```diff
@@ -42,9 +42,17 @@ func (r *TerraformModulePinnedSourceRule) Link() string {
...
+// See https://www.terraform.io/docs/modules/sources.html#github
+var ReGitHub = regexp.MustCompile("(^github.com/(.+)/(.+)$)|(^git@github.com:(.+)/(.+)$)")
+
+// ReBitbucket matches a module source which is a Bitbucket repository
+// See https://www.terraform.io/docs/modules/sources.html#bitbucket
+var ReBitbucket = regexp.MustCompile("^bitbucket.org/(.+)/(.+)$")
+
+// ReGenericGit matches a module source which is a Git repository
+// See https://www.terraform.io/docs/modules/sources.html#generic-git-repository
+var ReGenericGit = regexp.MustCompile("(git://(.+)/(.+))|(git::https://(.+)/(.+))|(git::ssh://((.+)@)??(.+)/(.+)/(.+))")
```
wata727: 👍

rules/terraformrules/terraform_module_semver_source_test.go
```diff
@@ -0,0 +1,383 @@
...
+module "pinned_git" {
+  source = "bitbucket.org/hashicorp/consul.git?ref=v1.2.3"
+}`,
+			Expected: tflint.Issues{},
+		},
+		{
+			Name: "bitbucket git module reference is pinned to semver (no leading v)",
+			Content: `
+module "pinned_git" {
+  source = "bitbucket.org/hashicorp/consul.git?ref=v1.2.3"
```
wata727:
```suggestion
  source = "bitbucket.org/hashicorp/consul.git?ref=1.2.3"
```

rules/terraformrules/terraform_module_semver_source_test.go
```diff
@@ -0,0 +1,383 @@
...
+module "pinned_git" {
+  source = "bitbucket.org/hg/mercurial?ref=v1.2.3"
+}`,
+			Expected: tflint.Issues{},
+		},
+		{
+			Name: "bitbucket mercurial module reference is pinned to semver (no leading v)",
+			Content: `
+module "pinned_git" {
+  source = "bitbucket.org/hg/mercurial?ref=v1.2.3"
```
wata727:
```suggestion
  source = "bitbucket.org/hg/mercurial?ref=1.2.3"
```

docs/rules/terraform_module_semver_source.md
alexwlchan: Damnit, thought I’d added those, they must have got lost! Will fix.

alexwlchan: Ah, I had it in the example file I was testing with, just forgot to add it to the docs. 😅

alexwlchan:
Thanks for the review; think I've fixed everything.

rules/terraformrules/terraform_module_semver_source_test.go
wata727: Is this resolved? Although the test name says "no leading v", the actual source starts with "v".

alexwlchan: 🤦‍♀ 

Try now?

wata727:
Thanks!

PR #530, opened 2019-11-28
update the docker image name

### LINKED ISSUE #529: Docker Hub repository does not exist
You advise to use `docker run --rm -v $(pwd):/data -t terraform-linters/tflint` for running with Docker, but `terraform-linters/tflint` doesn't exist

ozbillwang:
The docker image `terraform-linters/tflint` is not exist. Correct the name.

fix #529


wata727:
Thanks!

PR #531, opened 2019-12-01
Add python3.8 lambda runtime

fionn:
AWS Lambda supports the `python3.8` runtime. See https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html.


wata727:
Thanks for your contribution. But if you want to change an existing rule, you need to change the mappings, not directly. See https://github.com/terraform-linters/tflint/blob/master/docs/DEVELOPING.md#editing-the-existing-rules

wata727:
This change has been merged in https://github.com/terraform-linters/tflint/pull/541

PR #534, opened 2019-12-05
Bump the base image to v1.13.5

chenrui333:


wata727:
CI was failed, but this is not due to the change, so I'll merge it. Thank you as always!

PR #542, opened 2019-12-09
Fix the pre-commit hook

Dunedan:
The existing `pre-commit` configuration doesn't seem to work, as
`pre-commit` needs to be pointed to the compiled binary and not to
`main.go`. This commit fixes that and ensures `pre-commit` only tries to
lint terraform files.


wata727:
👍

PR #555, opened 2019-12-17
Parse absolute paths in TF_DATA_DIR correctly

madddi:
Resolves #554 

I didn't see how I could add an integration test for this, though.

PR #556, opened 2019-12-17
Remove legacy TF 0.11 fields from module records

madddi:
While working on #555 I discovered that the `Root` field is no longer present in TF 0.12, so I think you could remove it:

https://github.com/hashicorp/terraform/blob/master/internal/modsdir/manifest.go#L19-L41


wata727:
Good catch, thanks

PR #563, opened 2019-12-24
Add elasticache support for t3 family

BrunoChauvet:
Add support for `t3` family type as per [https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/CacheNodes.SupportedTypes.html](https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/CacheNodes.SupportedTypes.html)


wata727:
Thanks!

PR #608, opened 2020-02-02
Remove go111module on env variable in Dockerfile.

cedarkuo:
go 1.13 change go111module auto setting behavior.

```
the auto setting now activates the module-aware mode of the go command whenever the current working directory contains, or is below a directory containing, a go.mod file
```
https://golang.org/doc/go1.13#modules

So I think maybe this environment variable could be removed from Dockerfile.


wata727:
👍

Issue #150, opened 2017-10-16
Check for tags rule


mateusduboli:
Check for AWS resources which are tagable but doesn't have any tags specified.

A bonus point would be to check for specific tag keys like `Name` and `Team`.

wata727:
@mateusduboli Thanks for your suggestion!

However, I think that whether to specify tags on resources depends on a project. For that reason, I think that you should use [terraform_validate](https://github.com/elmundio87/terraform_validate) etc. that you can define rules yourself in such cases. WDYT?

mateusduboli:
Thanks for the suggestion!
But my doubt would be why not include in `tflint` too?

cristim:
I also +1 this, I  see no harm in tflint having this as well, much like pylint overlaps some of the pep8 rules for python code.

The key is to carefully implement the rules so that they don't conflict with the other tools in the default configuration.

The reality is that there are usually multiple choices and it is not always easy to cleanly segment their scope. Some users will only run one of those but still expect some decent support.

aferg439:
+1 from me, as well. Is this still being tracked? We'd love to implement this tool in our automation to check for tagging of resources where applicable. There's only so much that can be forced through modules.

wata727:
Now I think implementing this is not a bad option. We will implement it if necessary.

aferg439:
There aren't a whole lot of options out there that accomplish this. It would definitely be a win for the community.

wata727:
Closed this in #617

PR #617, opened 2020-02-10
Check for tags on AWS resources

bwhaley:
This PR adds code generated rules for checking resources for named AWS tags. 

In config, you can specify a `tags` attribute in the `config` block:

```
config {
    tags = ["foo", "bar"]
}
```

All the AWS resources with `tags` supported will be checked for tags matching these keys.

This addresses #150.


bwhaley:
> First of all, what do you think about integrating these tags rules together? In my understanding, these rules don't have a difference for each resource, so implementing them as a single rule is better in terms of maintainability.

I actually implemented it this way first, but I wasn't able to import `github.com/terraform-providers/terraform-provider-aws/aws`. Unfortunately, it resulted in a bunch of errors like:

```
panic: gob: registering duplicate types for "*tfdiags.rpcFriendlyDiag": *tfdiags.rpcFriendlyDiag != *tfdiags.rpcFriendlyDiag
```

Debugging this led me to [this doc](https://www.terraform.io/docs/extend/best-practices/depending-on-providers.html), which suggested that maybe depending on the provider in `tflint` was a bad idea. I saw that there was no other place in `tflint` where it was importing `terraform-provider-aws` - only in the tools. We need to be able to walk the resources to find which of them have `tags`. Otherwise assembling that list would be manual. Hence my approach of code generated rules, although I certainly agree that it's an ugly solution with so much duplication! I'm open to ideas on how to tackle that.

> Secondly, What do you think about implementing tags config as rule configurations? 

I thought about this too since the top level config block doesn't seem like the right place for an AWS-specific configuration. However, it strikes me as a terrible user experience to define the tags for each type of resource individually in a `rule` block. What about adding another type of block? E.g.

```
aws_config = {
    tags = ["foo", "bar"]
}
```

This would allow for additional AWS-specific resource configs in the future.

bwhaley:
In the most recent committed I've updated the code generation to create a list of the types of resources that have tags, rather than a duplicated rules file for each type. This seems much cleaner.

What do you think of the idea of a top-level `aws_config` block as suggested above?

wata727:
> I've updated the code generation to create a list of the types of resources that have tags, rather than a duplicated rules file for each type

Great. In addition, it is a good idea to make this list in `NewTagsRules()`. Imagine something like https://github.com/terraform-linters/tflint/blob/master/rules/awsrules/models/aws_instance_invalid_type.go

> it strikes me as a terrible user experience to define the tags for each type of resource individually in a rule block.

This problem may be solved by not defining rules for each resource. I think there is no reason to split rules for each resource. What do you think?

bwhaley:
This is a good idea! I will work on it tomorrow. 👍🏻

wata727:
@bwhaley I'm now working on a more flexible rule configuration in https://github.com/terraform-linters/tflint/pull/624. If this were merged, you can define the configuration of the tags into the rule block in the same way.

bwhaley:
Oh, okay, I'll wait for that to be merged then. I have the single rule version ready to go now. Just need to know how to load the config. Will you update the docs to explain the change in #624? I don't quite follow how it will work yet.

wata727:
Maybe this feature is as simple as no need to update the developer guide. You just need to make the struct of the configuration you want to reference in the rule and pass it to `runner.DecodeRuleConfig`.

The only caveat is that the struct must have tags for `hcl` package, but it's best to refer to the hcl documentation: https://github.com/hashicorp/hcl/blob/v2.3.0/gohcl/doc.go

bwhaley:
I've incorporated `runner.DecodeRuleConfig`. Let me know what you think.

rules/awsrules/aws_resource_tags.go
```diff
@@ -0,0 +1,311 @@
...
+					if _, ok := hash[tag]; ok {
+						found = append(found, tag)
+					}
+				}
+				if len(found) != len(config.Tags) {
+					sort.Strings(config.Tags)
+					sort.Strings(tags)
+					wanted := strings.Join(config.Tags, ",")
+					found := strings.Join(tags, ",")
+					runner.EmitIssue(r, fmt.Sprintf("Wanted tags: %v, found: %v", wanted, found), attribute.Expr.Range())
```
wata727: Perhaps there is a better message that users can read and understand exactly what to do. For example, what about the following?

> "aws_instance.instance" is tagged with "foo" and "bar", but it should be tagged with "Foo" and "Bar".

rules/awsrules/aws_resource_tags.go
```diff
@@ -0,0 +1,311 @@
...
+				for _, k := range tags {
+					hash[k] = true
+				}
+				var found []string
+				for _, tag := range config.Tags {
+					if _, ok := hash[tag]; ok {
+						found = append(found, tag)
+					}
+				}
+				if len(found) != len(config.Tags) {
```
wata727: In this case, the rule will emit an issue if the resource has extra tags. Intuitively, if it has been tagged with configured tags, I think that other tags should be allowed, but what about?

rules/awsrules/aws_resource_tags.go
```diff
@@ -0,0 +1,311 @@
...
+		"aws_workspaces_ip_group",
+	}
+	return &AwsResourceTagsRule{
+		resourceTypes: resourceTypes,
+	}
+}
+
+// Name returns the rule name
+func (r *AwsResourceTagsRule) Name() string {
+	return "aws_resource_tags"
```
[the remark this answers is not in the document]
bwhaley: Okay, renamed to `aws_resource_missing_tags`.

rules/awsrules/aws_resource_tags.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+// Enabled returns whether the rule is enabled by default
+func (r *AwsResourceTagsRule) Enabled() bool {
+	return true
```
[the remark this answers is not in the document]
bwhaley: Sure, makes sense. Done.

rules/awsrules/aws_resource_tags.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+// Severity returns the rule severity
+func (r *AwsResourceTagsRule) Severity() string {
+	return tflint.ERROR
```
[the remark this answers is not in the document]
bwhaley: changed to `NOTICE` severity.

rules/awsrules/aws_resource_tags.go
bwhaley: Updated the error message to look like this, which seems clear enough to me:

> aws_instance.instance is missing the following tags: "Foo".

rules/awsrules/aws_resource_tags.go
bwhaley: Actually, the logic I had there was correct, if a bit confusing. The `found` slice only contained the subset of tags that had been found in the config. If the length of the slices didn't match, it meant that some of the expected tags were missing. 

However, I've updated the logic in the most recent commit and it's now checked differently anyway, hopefully in a way that is more explicit.

bwhaley:
I've incorporated your feedback. Given a resource like this:

```
resource "aws_instance" "instance" {
  instance_type = "m5.large"
  tags = {
    foo = "bar"
    Bar = "baz"
    Baz = "qux"
  }
}
```

and a configuration like this:

```
rule "aws_resource_missing_tags" {
  enabled = true
  tags = ["Foo", "Bar"]
}
```

it produces an error like this:

```
1 issue(s) found:

Notice: aws_instance.instance is missing the following tags: "Foo". (aws_resource_missing_tags)

  on test.tf line 3:
   3:   tags = {
   4:     foo = "bar"
   5:     Bar = "baz"
   6:     Baz = "qux"
   7:   }

Reference: https://github.com/terraform-linters/tflint/blob/v0.14.0/docs/rules/aws_resource_missing_tags.md
```

Note that in order to access `resource` data in the `Check()` func, I had to essentially duplicate `WalkResourceAttributes` to call `walker()` with the active resource. Let me know if you'd suggest another approach on that.

wata727:
Umm... you are right. the current API has no way to access the resource details.

Temporarily creating a similar API is not a bad idea, but given the current situation, it might be a good idea to substitute an ambiguous message like "The resource". What do you think?

bwhaley:
Well I did like your suggestion to reference which resource was missing tags. Without it, the user will have to search through all the code to find which resource is missing them. A few options come to mind:

1) as you suggest, simplify to “the resource”
2) keep the duplicated code
3) update the API to include the resource (would entail changes to all the rules)
4) It could at least mention the resource type since I am crawling all of them that support tags. Only the resource name is unknown

wata727:
In my opinion, the location (file name, line number) is always displayed, so the resource name is not always important. So I would like to suggest 1.

bwhaley:
Okay, let me know if you have any further comments.

wata727:
👍

PR #645, opened 2020-02-29
Emit an issue when there is no tags definition

### LINKED ISSUE #644: Missing tag rules not triggering
I'm attempting to use the new [tag config rules](https://github.com/terraform-linters/tflint/pull/617) and I'm having trouble getting it to alert me to missing `tags`. 

For example, I have a dynamodb table declared which looks something like:
```terraform
resource "aws_dynamodb_table" "upgrade_locks" {
  name     = "StagingUpgradeLocks"
  hash_key = "StackId"

  write_capacity = 5
  read_capacity  = 5

  attribute {
    name = "StackId"
    type = "S"
  }
}
```

I'd expect this to alert me that the `tags` have not been set, but from the debug output I'm only seeing:
```
17:34:22 runner.go:506: [DEBUG] Walk `aws_dynamodb_table.upgrade_locks.hash_key` attribute
17:34:22 runner.go:506: [DEBUG] Walk `aws_dynamodb_table.upgrade_locks.name` attribute
```

And I receive no warning that the resource is missing `tags`.

I'm using a basic rule in my tflint.hcl config file, which doesn't have anything that I can see would cause an issue:
```hcl
rule "aws_resource_missing_tags" {
  enabled = true
  tags = ["tags_version"]
}
```

wata727:
Fixes #644 

`WalkResourceAttributes` walks only resources having "tags" definition. However, the rule should emit an issue even if there is no tags definition.

In this PR, the rule also walks resources even if the resource doesn't have tags definition.

/cc @bwhaley

### DIFF
--- a/rules/awsrules/aws_resource_missing_tags.go
+++ b/rules/awsrules/aws_resource_missing_tags.go
@@ -2,6 +2,7 @@ package awsrules
 
 import (
 	"fmt"
+	"log"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


bwhaley:
Ahh, sorry I didn't see this in time. Funny that we ended up not calling `WalkResourceAttributes` after all. Thanks for fixing this edge case that I missed.

PR #664, opened 2020-03-13
Use checkout action v2

chenrui333:


chenrui333:
relates to this, https://github.com/actions/checkout/issues/116

New syntax introduced in here, https://github.com/actions/checkout/issues/176

PR #666, opened 2020-03-14
Add install guide for Windows (choco install)

aaronsteers:
Windows users can also install `tflint` using Chocolatey: https://chocolatey.org/packages/tflint

```
choco install tflint
```

Equivalent to brew, but for Windows. 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,14 @@ macOS users can also use [Homebrew](https://brew.sh) to install TFLint:
 $ brew install tflint
 ```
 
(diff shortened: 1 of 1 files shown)


wata727:
👍

jsoconno:
When I install this way, `tflint` is not added to the PATH.  What is the recommended way to fix this?

To clarify, when I use `choco install tflint` and then try to run the command `tflint` it says the executable can't be found.

Issue #662, opened 2020-03-12
aws_resource_missing_tags rule flags tagged resource aws_autoscaling_group


nitrocode:
```
TFLint v0.15.1
Terraform v0.12.23
```

In `~/.tflint.hcl`
```hcl
# missing tags
rule "aws_resource_missing_tags" {
  enabled = true
  tags    = ["Name", "env"]
}
```

terraform

```hcl
resource "aws_autoscaling_group" "this" {
  ...
  dynamic "tag" {
    for_each = local.tags

    content {
      key                 = tag.key
      value               = tag.value
      propagate_at_launch = true
    }
  }
}
```

Since `tags` is an argument and `tag` is an argument for resource [`aws_autoscaling_group`](https://www.terraform.io/docs/providers/aws/r/autoscaling_group.html), there needs to be additional logic for this resource to check if it has the correct values.

Or perhaps we can exclude resources from rules. Is there an exclusion argument for rules?

@bwhaley

bwhaley:
Ahh, I didn't realize that `aws_autoscaling_group` approached tags differently. 

There are a couple of distinct issues here:

1. The rule should handle both the `tag` block and the `tags` attribute for `aws_autoscaling_group`. I have some prototype logic that seems to work, and once I finish that up I'll submit a pull request.
2. AFAICT, [`dynamic` blocks with the `for_each` iterator are not expanded](https://github.com/terraform-linters/tflint/blob/master/tflint/runner.go#L541), so those will be unevaluable. I might be missing something here. Perhaps @wata727 can comment on that.

wata727:
Yes, dynamic blocks cannot be expanded, so they need to be treated specially. There are several options.

1. Ignore resources that have a dynamic `tag`. This is a design that allows false negatives and focuses on not reporting any extra issues.
2. If a resource has a dynamic `tag` but cannot verify that the key exists, emit an issue. This is a design that allows false positives and focuses on not missing out on problems.

I'm not sure which one is better, but probably the former is more acceptable to most users.

nitrocode:
That's too bad about the `dynamic` block unable to be expanded. Is there a ticket for terraform to fix this limitation?

Would you be open to a `exclude` clause for particular resources? or for now, should we just remove the `aws_autoscaling_group` from the list of resources that the rule checks for until we have better logic for the `tags` argument?

https://github.com/terraform-linters/tflint/blob/master/rules/awsrules/aws_resource_missing_tags.go#L48

[...]

PR #670, opened 2020-03-16
Adds support for aws_autoscaling_group tag blocks and tags attributes

bwhaley:
Fixes https://github.com/terraform-linters/tflint/issues/662

### DIFF
--- a/rules/awsrules/aws_resource_missing_tags.go
+++ b/rules/awsrules/aws_resource_missing_tags.go
@@ -300,7 +300,13 @@ func (r *AwsResourceMissingTagsRule) Check(runner *tflint.Runner) error {
 
 		// Special handling for tags on aws_autoscaling_group resources
 		if resourceType == "aws_autoscaling_group" {
-			r.checkAwsAutoScalingGroups(runner, config)
+			err := r.checkAwsAutoScalingGroups(runner, config)
+			err = runner.EnsureNoError(err, func() error {
+				return nil
+			})
+			if err != nil {
+				return err
+			}
 			continue
 		}
 
@@ -410,20 +416,25 @@ func (r *AwsResourceMissingTagsRule) checkAwsAutoScalingGroupsTag(runner *tflint
 			return tags, tagBlock.DefRange, diags
 		}
 
+		if _, ok := attributes["key"]; !ok {
+			err := &tflint.Error{
+				Code:  tflint.UnevaluableError,
+				Level: tflint.WarningLevel,
+				Message: fmt.Sprintf("Did not find expected field \"key\" in aws_autoscaling_group \"%s\" starting at line %d",
+					resource.Name,
+					resource.DeclRange.Start.Line,
+				),
+			}
+			return tags, resource.DeclRange, err
+		}
+
 		var key string
 		err := runner.EvaluateExpr(attributes["key"].Expr, &key)
 		if err != nil {
-			return tags, tagBlock.DefRange, diags
+			return tags, tagBlock.DefRange, err
 		}
-
-		var val string
-		err = runner.EvaluateExpr(attributes["value"].Expr, &val)
-		if err != nil {
-			return tags, tagBlock.DefRange, diags
-		}
-		tags[key] = val
+		tags[key] = ""
 	}
-
 	return tags, resource.DeclRange, nil
 }
 
@@ -491,3 +502,4 @@ func stringInSlice(a string, list []string) bool {
 	}
 	return false
 }
+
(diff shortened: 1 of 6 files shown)


rules/awsrules/aws_resource_missing_tags.go
```diff
@@ -279,30 +287,41 @@ func (r *AwsResourceMissingTagsRule) Link() string {
...
 
 	for _, resourceType := range r.resourceTypes {
+		// Skip this resource if its type is excluded in configuration
+		if stringInSlice(resourceType, config.Exclude) {
+			continue
+		}
+
+		// Special handling for tags on aws_autoscaling_group resources
+		if resourceType == "aws_autoscaling_group" {
+			r.checkAwsAutoScalingGroups(runner, config)
```
wata727: The returned error appears to be discarded. Please use the `EnsureNoError` API to test if you should stop the inspection because this error can be application-specific errors.

tflint/runner.go
```diff
@@ -316,12 +317,24 @@ func (r *Runner) EvalExpr(expr hcl.Expression, ret interface{}) (cty.Value, erro
...
+	val, err := r.EvalExpr(expr, ret, cty.Type{})
 	if err != nil {
 		return err
 	}
+	return r.fromCtyValue(val, expr, ret)
+}
 
-	err = gocty.FromCtyValue(val, ret)
+// EvaluateExprType is like EvaluateExpr, but also accepts a known cty.Type to pass to EvalExpr
+func (r *Runner) EvaluateExprType(expr hcl.Expression, ret interface{}, wantType cty.Type) error {
```
wata727: 👍

It will be great if the two APIs are integrated in the future.

rules/awsrules/aws_resource_missing_tags.go
```diff
@@ -320,6 +339,135 @@ func (r *AwsResourceMissingTagsRule) Check(runner *tflint.Runner) error {
...
+	}
+
+	for _, tagBlock := range body.Blocks {
+		attributes, diags := tagBlock.Body.JustAttributes()
+		if diags.HasErrors() {
+			return tags, tagBlock.DefRange, diags
+		}
+
+		var key string
+		err := runner.EvaluateExpr(attributes["key"].Expr, &key)
```
wata727: It has a possibility of a null pointer exception.

rules/awsrules/aws_resource_missing_tags.go
```diff
@@ -320,6 +339,135 @@ func (r *AwsResourceMissingTagsRule) Check(runner *tflint.Runner) error {
...
+		if err != nil {
+			return tags, tagBlock.DefRange, diags
```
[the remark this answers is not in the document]
bwhaley: Good catch.

rules/awsrules/aws_resource_missing_tags.go
```diff
@@ -320,6 +339,135 @@ func (r *AwsResourceMissingTagsRule) Check(runner *tflint.Runner) error {
...
+		}
+
+		var val string
+		err = runner.EvaluateExpr(attributes["value"].Expr, &val)
```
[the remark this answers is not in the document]
bwhaley: Makes sense, I'll just remove those 4 lines since we don't even use `val`.

rules/awsrules/aws_resource_missing_tags.go
bwhaley: Oops, I had meant to come back to that but forgot. I will emit a message with `err.Error()`.

tflint/runner.go
bwhaley: Agreed. I didn't love this solution, but fixing it seemed like a lot of refactoring which didn't seem prudent.

rules/awsrules/aws_resource_missing_tags.go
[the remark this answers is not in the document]
bwhaley: d583da6

rules/awsrules/aws_resource_missing_tags.go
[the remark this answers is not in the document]
bwhaley: d583da6

rules/awsrules/aws_resource_missing_tags.go
bwhaley: d583da6

rules/awsrules/aws_resource_missing_tags.go
bwhaley: d583da6

bwhaley:
Thanks for the review! I believe I've incorporated all the feedback in d583da6. Let me know if any other changes are needed.

wata727:
👍

PR #679, opened 2020-03-20
Add bucket-owner-full-control to allowed S3 ACLs

sds:
The `bucket-owner-full-control` ACL is mentioned in AWS documentation: https://docs.aws.amazon.com/AmazonS3/latest/dev/example-walkthroughs-managing-access-example3.html#access-policies-walkthrough-cross-account-acl-acctB-tasks

Add it to the list of allowed ACLs.


wata727:
👍

PR #686, opened 2020-03-23
Fixes the example of excluding resource types

bwhaley:
The current example doesn't make sense. Probably a copypasta error.


wata727:
👍

PR #721, opened 2020-04-21
Add a rule to enforce Terraform types for variables

mveitas:
This rule enforces that a type needs to be declared for variables. Enforcing this rule provides consistency across modules and ensures that engineers think about the type of variable being declared and makes it easier to consume the module knowing the type of the variable.

### DIFF
--- a/docs/rules/terraform_typed_variables.md
+++ b/docs/rules/terraform_typed_variables.md
@@ -20,7 +20,7 @@ variable "enabled" {
 $ tflint
 1 issue(s) found:
 
-Notice: `no_type` variable has no type (terraform_typed_variables)
+Warning: `no_type` variable has no type (terraform_typed_variables)
 
   on template.tf line 1:
    1: variable "no_type" {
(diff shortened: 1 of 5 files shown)


docs/rules/terraform_typed_variables.md
```diff
@@ -0,0 +1,37 @@
...
+  description = "This is description"
+  type        = bool
+}
+```
+
+```
+$ tflint
+1 issue(s) found:
+
+Notice: `no_type` variable has no type (terraform_typed_variables)
```
wata727, line 23: If you update severity to warning, you must update this as well.

mveitas: Good catch and this has been updated

wata727:
Please register the created rule in provider :)

https://github.com/terraform-linters/tflint/blob/527ac4da3cf588aaf632500db5c27776f235a574/rules/provider.go#L39-L46

wata727:
👍

PR #725, opened 2020-04-23
Adding rule for terraform_required_version

### LINKED ISSUE #711: Rule for Terraform version constraint
Hi there,

It would be helpful to have a rule that ensures that a Terraform version constraint is both present within a module and also specifies a specific minimum version (say `>=0.12.0`).

mveitas:
This rule will enforce that a required_version constraint be applied to the `terraform` block


mveitas:
Alternative name for the rule might be `terraform_required_version`. Thoughts?

wata727:
I'm not sure it would be useful to be able to enforce version constraints by this rule.
What kind of situations do you expect to be useful? In most cases, I think it is good enough to edit `required_version` directly. Thank you for your contribution.

mveitas:
Originally the intent was to just enforce that the `required_version` existed, but as mentioned in #711, someone had a use case to enforce which version was specified.

wata727:
Oops, I may have misunderstood.
I have thought this issue mentioned a rule that enforces the Terraform version constraints that TFLint works.

@DWSR Does this rule meet your needs? Also, please tell us under what situations you find this rule useful. Thank you.

mveitas:
@wata727 Do you find that this is a valid rule to add to enforce that the `required_version` exists? I could see if an organization had a number of repos that contain modules, they might share a common configuration which makes it easier to enforce this sort of constraint on a specific matching version

Our use case is that we have defined a number of standards when building our TF modules, and ensuring that a version is defined is something we want automate rather than relying on humans to enforce this.

wata727:
Yes, a rule that simply enforces the presence of `required_version` makes a lot of sense.
I'm wondering if it's worth enforcing the contents of the constraint. Is there a case where we want to impose the same constraints on all modules ...?

mveitas:
Let me change the rule to just do enforcement of the attribute as a starting point and keep this simple. We can continue the discussion of validating the contents on the attribute in #711

wata727:
👍

mveitas:
@wata727 Do you think a release could be cut with the last two rules so they could start to be used? Not sure what the cadence is for doing a release.

wata727:
I just released v0.15.5. Please try it :)

PR #728, opened 2020-04-25
Allow empty string as a valid of dynamodb table stream view type

### LINKED ISSUE #723: Rule aws_dynamodb_table_invalid_stream_view_type doesn't accept empty string when stream is disabled
As explained in this AWS provider issue: https://github.com/terraform-providers/terraform-provider-aws/issues/1134
We need to be able to pass empty string ("") as stream_view_type to the aws_dynamodb_table when stream_enabled is false, it is the input the AWS API expects, I got the described error with different inputs with the stream disabled.

Right now, tflint only allows the valid non empty values and returns an error message:
"" is an invalid value as stream_view_type (aws_dynamodb_table_invalid_stream_view_type)

wata727:
Fixes #723


PatMyron:
similar: https://github.com/terraform-linters/tflint/pull/355, https://github.com/terraform-linters/tflint/pull/401, https://github.com/terraform-linters/tflint/pull/442, https://github.com/terraform-linters/tflint/pull/604

PR #731, opened 2020-05-01
Add terraform_required_providers rule

bendrucker:
This PR adds a rule that checks for `terraform.required_providers` and falls back to `provider.version` if set. It borrows copy from the Terraform docs:

https://www.terraform.io/docs/configuration/providers.html#provider-versions

### DIFF
--- a/docs/rules/terraform_required_providers.md
+++ b/docs/rules/terraform_required_providers.md
@@ -13,9 +13,7 @@ rule "terraform_required_providers" {
 ## Example
 
 ```hcl
-provider "template" {
-  version = "2"
-}
+provider "template" {}
 ```
 
 ```
--- a/docs/rules/terraform_required_providers.md
+++ b/docs/rules/terraform_required_providers.md
@@ -38,7 +38,7 @@ Add the `required_providers` attribute to the `terraform` configuration block an
 ```tf
 terraform {
   required_providers {
-    aws = "~> 2.0"
+    template = "~> 2.0"
   }
 }
 ```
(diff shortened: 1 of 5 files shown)


mveitas:
👍 I recently added a rule to enforce that the terraform `required_version` attribute be set for a module and thought about making it a rule to perform enforcement of the version and the providers within the terraform block. Separating this out keep the code simple and easy to understand!

mveitas:
You are going to need to add the rule here: https://github.com/mveitas/tflint/blob/master/rules/provider.go

bendrucker:
Thanks, will do!

docs/rules/terraform_required_providers.md
```diff
@@ -0,0 +1,46 @@
...
+  enabled = true
+}
+```
+
+## Example
+
+```hcl
+provider "template" {
+  version = "2"
+}
```
wata727, line 18: This example doesn't seem to report the following warning 🤔

docs/rules/terraform_required_providers.md
```diff
@@ -0,0 +1,46 @@
...
+Providers are plugins released on a separate rhythm from Terraform itself, and so they have their own version numbers. For production use, you should constrain the acceptable provider versions via configuration, to ensure that new versions with breaking changes will not be automatically installed by `terraform init` in future.
+
+## How To Fix
+
+Add the `required_providers` attribute to the `terraform` configuration block and include current versions for all providers. For example:
+
+```tf
+terraform {
+  required_providers {
+    aws = "~> 2.0"
```
wata727, line 41: It would be easier to understand if it was a fix of the above example.

```suggestion
    template = "~> 2.0"
```

docs/rules/terraform_required_providers.md
bendrucker: Good catch!

bendrucker:
Fixed the two mistakes in the docs, thanks!

PR #732, opened 2020-05-01
Copy edits for rules docs

bendrucker:
This PR makes assorted grammar edits to the rules docs that I noticed as I was reading them and enabling rules.

Thank you for all your work here, excited to see the project moving in a modular direction!


bendrucker:
No problem, always happy to do it!

PR #738, opened 2020-05-06
add terraform_workspace_remote rule

bendrucker:
This PR adds a rule that warns against the use of `terraform.workspace` interpolations when the user has a "remote" backend configured. This backend is used with Terraform Cloud and most TF Cloud users want to execute Terraform remotely. In these circumstances, `terraform.workspace` will always evaluate to `"default"`. More here:

https://www.terraform.io/docs/backends/types/remote.html#workspaces

While you can use the remote backend as a state store and [execute Terraform runs locally](https://www.terraform.io/docs/cloud/run/index.html#disabling-remote-operations), this is of limited use and anyone using Terraform Cloud will likely want to move away from `terraform.workspace` for all of their configuration.

If `terraform.workspace` does not evaluate to the expected value, Terraform configuration may break or TF may plan unexpected changes.


rules/terraformrules/terraform_workspace_remote.go
```diff
@@ -0,0 +1,108 @@
...
+		r.checkForTerraformWorkspaceInBody(runner, block.Body)
+	}
+
+	return
+}
+
+func (r *TerraformWorkspaceRemoteRule) checkForTerraformWorkspaceInExpr(runner *tflint.Runner, expr hcl.Expression) {
+	var used bool
+	for _, t := range expr.Variables() {
+		if len(t) != 2 || t.IsRelative() || t.RootName() != "terraform" {
```
wata727, line 91: `lang.ReferencesInExpr` is a better way to parse expressions. TFLint uses it to determine if it has an unevaluable reference.

https://github.com/terraform-linters/tflint/blob/689426718cd1d2c499f8de996fbcf06e1533e387/tflint/runner.go#L748-L754

https://github.com/terraform-linters/tflint/blob/689426718cd1d2c499f8de996fbcf06e1533e387/tflint/runner.go#L773-L784

https://github.com/hashicorp/terraform/blob/v0.12.24/addrs/terraform_attr.go

rules/terraformrules/terraform_workspace_remote_test.go
```diff
@@ -0,0 +1,200 @@
...
+	backend "remote" {}
+}
+resource "null_resource" "a" {
+	triggers = {
+		w = terraform.workspace
+	}
+}`,
+			Expected: tflint.Issues{
+				{
+					Rule:    NewTerraformDeprecatedInterpolationRule(),
```
wata727, line 29: `NewTerraformWorkspaceRemoteRule`...

It seems that tests are passed because `AssertIssues` ignores the `Rule` field.

bendrucker: Whoops, obviously did some copy-pasting here. Will send a fix for `AssertIssues` as well.

bendrucker:
Updated this to use `lang`/`addrs` and `WalkExpressions`, super concise now!

PR #739, opened 2020-05-06
Add unused_declarations rule

bendrucker:
Declaring something and then not using it is typically either:

* A bug, and you need to reference it
* A leftover from removed code, and you can remove it

Variables, locals, and data sources are all declared inputs and if they are not direct referenced, they are effectively dead code. This rule works by:

* Copying the maps representing these values
* Visiting all expressions that could reference something, i.e. `resource`, `data`, `locals`, `provider`, `output`, and `module` blocks
* Examining each expression's variables and if they begin with a matched prefix (var, data, local),  deleting the corresponding pointer from the appropriate map
* Emitting any remaining pointers in each map as declared but unused

### DIFF
--- a/rules/terraformrules/terraform_unused_declaration.go
+++ b/rules/terraformrules/terraform_unused_declaration.go
@@ -5,7 +5,6 @@ import (
 	"log"
 
 	"github.com/hashicorp/hcl/v2"
-	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/hashicorp/terraform/addrs"
 	"github.com/hashicorp/terraform/configs"
 	"github.com/hashicorp/terraform/lang"
@@ -51,23 +50,11 @@ func (r *TerraformUnusedDeclarationsRule) Check(runner *tflint.Runner) error {
 	log.Printf("[TRACE] Check `%s` rule for `%s` runner", r.Name(), runner.TFConfigPath())
 
 	decl := r.declarations(runner.TFConfig.Module)
-	for _, resource := range runner.TFConfig.Module.ManagedResources {
-		r.checkForRefsInBody(runner, resource.Config, decl)
-	}
-	for _, data := range runner.TFConfig.Module.DataResources {
-		r.checkForRefsInBody(runner, data.Config, decl)
-	}
-	for _, provider := range runner.TFConfig.Module.ProviderConfigs {
-		r.checkForRefsInBody(runner, provider.Config, decl)
-	}
-	for _, module := range runner.TFConfig.Module.ModuleCalls {
-		r.checkForRefsInBody(runner, module.Config, decl)
-	}
-	for _, output := range runner.TFConfig.Module.Outputs {
-		r.checkForRefsInExpr(runner, output.Expr, decl)
-	}
-	for _, local := range runner.TFConfig.Module.Locals {
-		r.checkForRefsInExpr(runner, local.Expr, decl)
+	err := runner.WalkExpressions(func(expr hcl.Expression) error {
+		return r.checkForRefsInExpr(expr, decl)
+	})
+	if err != nil {
+		return err
 	}
 
 	for _, variable := range decl.Variables {
@@ -115,27 +102,10 @@ func (r *TerraformUnusedDeclarationsRule) declarations(module *configs.Module) *
 	return decl
 }
 
-func (r *TerraformUnusedDeclarationsRule) checkForRefsInBody(runner *tflint.Runner, body hcl.Body, decl *declarations) {
-	nativeBody, ok := body.(*hclsyntax.Body)
-	if !ok {
-		return
-	}
-
-	for _, attr := range nativeBody.Attributes {
-		r.checkForRefsInExpr(runner, attr.Expr, decl)
-	}
-
-	for _, block := range nativeBody.Blocks {
-		r.checkForRefsInBody(runner, block.Body, decl)
-	}
-
-	return
-}
-
-func (r *TerraformUnusedDeclarationsRule) checkForRefsInExpr(runner *tflint.Runner, expr hcl.Expression, decl *declarations) {
+func (r *TerraformUnusedDeclarationsRule) checkForRefsInExpr(expr hcl.Expression, decl *declarations) error {
 	refs, diags := lang.ReferencesInExpr(expr)
 	if diags.HasErrors() {
-		return
+		return diags.Err()
 	}
 
 	for _, ref := range refs {
@@ -150,4 +120,6 @@ func (r *TerraformUnusedDeclarationsRule) checkForRefsInExpr(runner *tflint.Runn
 			delete(decl.DataResources, sub.Resource.String())
 		}
 	}
+
+	return nil
 }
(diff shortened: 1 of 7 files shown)


rules/terraformrules/terraform_unused_declaration.go
```diff
@@ -0,0 +1,169 @@
...
+	for _, block := range nativeBody.Blocks {
+		r.checkForVariablesInBody(runner, block.Body, decl)
+	}
+
+	return
+}
+
+func (r *TerraformUnusedDeclarationsRule) checkForVariablesInExpr(runner *tflint.Runner, expr hcl.Expression, decl *declarations) {
+	for _, variable := range expr.Variables() {
+		split := variable.SimpleSplit()
```
wata727, line 135: Using `lang.ReferencesInExpr` may simplify the code.

rules/terraformrules/terraform_unused_declaration.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	for _, module := range runner.TFConfig.Module.ModuleCalls {
+		r.checkForVariablesInBody(runner, module.Config, decl)
+	}
+	for _, output := range runner.TFConfig.Module.Outputs {
+		r.checkForVariablesInExpr(runner, output.Expr, decl)
+	}
+	for _, local := range runner.TFConfig.Module.Locals {
+		r.checkForVariablesInExpr(runner, local.Expr, decl)
+	}
```
wata727, line 69: Implementing functions like `WalkExpressions` in the `Runner` can simplify the implementation of rules that visit all expressions. Imagine something like `WalkResourceAttributes`.

https://github.com/terraform-linters/tflint/blob/689426718cd1d2c499f8de996fbcf06e1533e387/tflint/runner.go#L505-L521

It's a refactoring idea, so you can add all the rules and then work on another pull request 👍

bendrucker:
Thank you for the reviews, learning a lot here! Have dabbled with high level config inspection in the past and spent some time learning hclwrite and the hcl package more broadly recently in https://github.com/bendrucker/terraform-cloud-migrate. Still getting acquainted with all the APIs and how to get them working together in a way that doesn't effectively reimplement Terraform internals the way I initially was.

I went ahead and implemented `WalkExpressions` in this PR. Since it has a much bigger potential audience than #738, I'll just finish that PR after this. It simplifies the rule code significantly!

tflint/runner.go
```diff
@@ -501,6 +502,69 @@ func (r *Runner) LookupIssues(files ...string) Issues {
...
+// returns an error.
+func (r *Runner) WalkExpressions(walker func(hcl.Expression) error) error {
+	visit := func(expr hcl.Expression) error {
+		return r.WithExpressionContext(expr, func() error {
+			return walker(expr)
+		})
+	}
+
+	for _, resource := range r.TFConfig.Module.ManagedResources {
+		if err := r.walkBody(resource.Config.(*hclsyntax.Body), visit); err != nil {
```
wata727, line 517: This type-casting panics when inspecting JSON syntax.

```
% tflint
Panic: interface conversion: hcl.Body is *json.body, not *hclsyntax.Body
 -> 0: main.main.func1: /main.go(30)
 -> 1: runtime.gopanic: /panic.go(967)
 -> 2: runtime.panicdottypeE: /iface.go(260)
 -> 3: runtime.panicdottypeI: /iface.go(270)
 -> 4: github.com/terraform-linters/tflint/tflint.(*Runner).WalkExpressions: /runner.go(517)
 -> 5: github.com/terraform-linters/tflint/rules/terraformrules.(*TerraformUnusedDeclarationsRule).Check: /terraform_unused_declaration.go(53)
 -> 6: github.com/terraform-linters/tflint/cmd.(*CLI).inspect: /inspect.go(57)
 -> 7: github.com/terraform-linters/tflint/cmd.(*CLI).Run: /cli.go(82)
 -> 8: main.main: /main.go(43)
 -> 9: runtime.main: /proc.go(203)
 -> 10: runtime.goexit: /asm_amd64.s(1373)
```

However, I'm not sure how to walk all expressions on `json.body`. In this case, one way is skipping the analysis.

tflint/runner.go
```diff
@@ -501,6 +502,69 @@ func (r *Runner) LookupIssues(files ...string) Issues {
 	return issues
 }
 
+// WalkExpressions visits all blocks that can contain expressions:
+// resource, data, module, provider, locals, and output. It calls the walker
+// function with every expression it encounters and halts if the walker
+// returns an error.
+func (r *Runner) WalkExpressions(walker func(hcl.Expression) error) error {
```
wata727, line 509: 👍 

It would be better if it has tests.

bendrucker: Will do!

tflint/runner.go
bendrucker: Makes sense. I'll look into what it might take to walk it, but the unexported type is a bad sign. Might just have to skip for now. I can imagine visiting every leaf node but that's not entirely correct since a `map` can have keys that are non-static expressions.

bendrucker:
Back with good news!

I added tests for `WalkExpressions` that covers the different block types and asserts the ranges of all the expressions in each test. Also covers recursing into nested blocks.


Figured out how to deal with JSON! A [`json.body`](https://github.com/hashicorp/hcl/blob/hcl2/json/structure.go#L14) is unexported so initially I expected that'd be a problem since we're accessing `.Block` on an `*hclsyntax.Body` to traverse. On further reflection, there's a reason these types are pretty different. HCL distinguishes blocks with syntax. JSON cannot do that.

For example, the following distinct HCL documents would appear identical in JSON:

```hcl
foo {
  bar {
    baz = "qux"
  }
}
```

```hcl
foo {
  bar = [{
    baz = "qux"
  }]
}
```

As JSON, both are:

```json
{
  "foo": {
    "bar": [{
      "baz": "qux"
    }]
  }
}
```

So if we're dealing with JSON, everything has to be treated as an attribute until you have a schema that can be used to read blocks:

https://github.com/hashicorp/hcl/blob/hcl2/json/structure.go#L133-L138

So calling `JustAttributes` will work. The only caveat is that elements that HCL would consider syntax (e.g. nested blocks) will be considered expressions. I've been trying to think up ways this cause issues. It's possible that a tflint rule could be confused by configuration that Terraform would ultimately reject as invalid once a provider's resource schema is applied. For unused declarations:

Given an schema matching the following HCL:

```
resource "foo" "bar" {
  nested {
    again {}
  }
}
```

The following JSON would treat `var.again` as used:

```json
{
  "resource": {
    "foo": {
      "bar": {
        "nested": [{
          "${var.again}": []
        }]
      }
    }
  }
}
```

I doubt Terraform will accept that configuration, even if `var.again == "again"`, given that `again` is a schema attribute of each `Elem` of `nested` as opposed to a user-specifiable `map`. But the rule will see `var.again` as used. That seems like an acceptable trade-off to me, but each rule may vary.

wata727:
Excellent!

> That seems like an acceptable trade-off to me

Agreed. Probably it's a rare case, so we don't need to pay much attention to it. This function, which works in many cases, is very valuable.

PR #745, opened 2020-05-11
Expose raw hcl.File objects to rules

### LINKED ISSUE #741: terraform_typed_variables - unable to distinguish between missing type and "any" type
The following variable declarations will fail for the rule `terraform_typed_variables`
```hcl
variable "lifecycle_rules" {
   description = "S3 bucket lifecycle rules"
   type        = any
}
```

This appears to be an issue with the underlying typing system used by Terraform and not being able to distinguish between a missing and `any` type declaration

## Version
TFLint v0.15.5

bendrucker:
This PR adds an initial proof of concept of reading files, parsing as HCL/JSON, and then traversing the content based on a partial schema for exactly the attributes we want. It would close #741 and close #744. 

Happy to factor the parser/parsing into the runner as well. Just wanted to get this up and see if you're interested in this direction for this and similar rules.

### DIFF
--- a/tflint/runner.go
+++ b/tflint/runner.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"path/filepath"
 	"strings"
 	"sync"
 
@@ -13,6 +14,7 @@ import (
 	"github.com/hashicorp/terraform/configs"
 	"github.com/hashicorp/terraform/lang"
 	"github.com/hashicorp/terraform/terraform"
+	"github.com/spf13/afero"
 	"github.com/terraform-linters/tflint/client"
 	"github.com/zclconf/go-cty/cty"
 )
@@ -25,6 +27,7 @@ type Runner struct {
 	Issues    Issues
 	AwsClient *client.AwsClient
 
+	fs          afero.Fs
 	ctx         terraform.BuiltinEvalContext
 	annotations map[string]Annotations
 	config      *Config
@@ -54,6 +57,7 @@ func NewRunner(c *Config, ants map[string]Annotations, cfg *configs.Config, vari
 		Issues:    Issues{},
 		AwsClient: &client.AwsClient{},
 
+		fs: afero.NewOsFs(),
 		ctx: terraform.BuiltinEvalContext{
 			PathValue: cfg.Path.UnkeyedInstanceShim(),
 			Evaluator: &terraform.Evaluator{
@@ -227,6 +231,10 @@ func (r *Runner) LookupIssues(files ...string) Issues {
 		}
 	}
 	return issues
+}
+
+func (r *Runner) ReadFile(filename string) ([]byte, error) {
+	return afero.ReadFile(r.fs, filepath.Join(r.TFConfig.Module.SourceDir, filename))
 }
 
 // EnsureNoError is a helper for processing when no error occurs
--- a/tflint/runner.go
+++ b/tflint/runner.go
@@ -233,6 +233,7 @@ func (r *Runner) LookupIssues(files ...string) Issues {
 	return issues
 }
 
+// ReadFile reads a file from the current module from disk by filename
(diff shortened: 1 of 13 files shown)


mveitas:
> Happy to factor the parser/parsing into the runner as well. Just wanted to get this up and see if you're interested in this direction for this and similar rules.

👍 This would be a great addition and open up the potential for allowing for additional linting capabilities.

wata727:
Yeah, this seems like a very good idea.

The way I was thinking was to add a low-level HCL files field (like `map[string]*hcl.File`) to the Runner. This allows to each rule to have access to `TFConfig` or low-level HCL APIs as needed.

In such cases, it is better the `Loader` loads object from a file system. Variable files and annotations are read in that way and build the runner.
https://github.com/terraform-linters/tflint/blob/v0.15.5/cmd/inspect.go#L95-L113

This change can be a bit painful, so it's not a bad option to add such file-reading functionality to the `Runner` once.

bendrucker:
Sounds good! Happy to work on the surrounding implementation a bit and move these pieces into the `Loader`. 

Looks like `LoadHCLFile` takes care of HCL + JSON so  that should be a good fit:

https://github.com/hashicorp/terraform/blob/v0.12.24/configs/parser.go#L65

bendrucker:
Ok, updates! 

* Added a `Files` method to the loader that returns a map of `hcl.File` objects that the loader has seen
  * Terraform has a cache of `*hcl.File` instead of `[]bytes` but it's not exported in a way we can access
  * It seems like `LoadAnnotations` should also use `Sources()` rather than loading from disk again? Happy to do that in a separate PR.
* Added a `files` field to Runner
* Called `loader.Files` when constructing a runner to set `files`
* Added a `File` method to Runner that returns a file from that map

Rules can call that method, presumably with a `DeclRange.Filename` or something similar, to get the raw HCL representation of any config file they want to inspect.

tflint/runner.go
```diff
@@ -186,7 +188,7 @@ func NewModuleRunners(parent *Runner) ([]*Runner, error) {
 			}
 		}
 
-		runner, err := NewRunner(parent.config, parent.annotations, cfg)
+		runner, err := NewRunner(parent.config, parent.files, parent.annotations, cfg)
```
wata727, line 191: In the module inspection, the second argument should be the files of the module to be parsed, not the parent module. This means that we need to use Loader here.

However, it would not be a good design for Runner to rely on Loader internally. One idea is to preload all files for all modules on the first load. Maybe it's a bit of a daunting task, what do you think?

bendrucker: Ooh yes good catch. I think this map will actually contain all files loaded by the loader, since `BuildConfig` will walk all of the child modules:

https://github.com/terraform-linters/tflint/blob/be3aa7bbfe941ca64d7b0cfdb50d676605630171/tflint/loader.go#L97

After doing a bit of poking, I don't think this will break anything, since any path references Terraform has should be relative to the parent dir. However, it does mean that a runner can potentially return a file used outside of its module.

Rather than try to call [`parser.ConfigDirFiles`](https://pkg.go.dev/github.com/hashicorp/terraform/configs?tab=doc#Parser.ConfigDirFiles) and essentially mix in loader behavior to the runner, a simple fix here would be to continue to use the same map of files in every runner but validate that path provided has the same directory as the module.

wata727: Oops, I was misunderstanding. I thought this map would only contain root module files. As you say, it seems that all files are included.

Certainly, it's possible that we can refer to files outside the module, but I don't think there's a big problem. Thank you for the explanation.

bendrucker:
Yes feel free to merge!

wata727:
@bendrucker Thank you again!

By the way, are you interested in joining maintainers? I would be very happy if you help this project based on your deep experience with Terraform and HCL.

bendrucker:
Absolutely!

wata727:
Thank you! I added you to the maintainers team 🎉

PR #759, opened 2020-05-19
ignore lang.ReferencesInExpr errors when walking all expressions

bendrucker:
When visiting all expressions, we might encounter special attributes/blocks, where normal reference rules no longer apply:

https://www.terraform.io/docs/configuration/resources.html#meta-arguments

This currently returns an error for the `unused_declarations` and `workspace_remote` rules:

```
Error: Invalid reference: A reference to a resource type must be followed by at least one attribute access, specifying the resource name.
```

Detecting these attributes seems too brittle/awkward, so simply exiting if an expression cannot be evaluated for references seems like the best fix

### DIFF
--- a/rules/terraformrules/terraform_unused_declaration.go
+++ b/rules/terraformrules/terraform_unused_declaration.go
@@ -105,6 +105,7 @@ func (r *TerraformUnusedDeclarationsRule) declarations(module *configs.Module) *
 func (r *TerraformUnusedDeclarationsRule) checkForRefsInExpr(expr hcl.Expression, decl *declarations) error {
 	refs, diags := lang.ReferencesInExpr(expr)
 	if diags.HasErrors() {
+		log.Printf("[DEBUG] Cannot find references in expression, ignoring: %v", diags.Err())
 		return nil
 	}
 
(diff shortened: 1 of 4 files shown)


rules/terraformrules/terraform_unused_declaration.go
```diff
@@ -105,7 +105,7 @@ func (r *TerraformUnusedDeclarationsRule) declarations(module *configs.Module) *
 func (r *TerraformUnusedDeclarationsRule) checkForRefsInExpr(expr hcl.Expression, decl *declarations) error {
 	refs, diags := lang.ReferencesInExpr(expr)
 	if diags.HasErrors() {
-		return diags.Err()
+		return nil
```
wata727, line 108: Logging errors instead of just ignoring them may help in investigating when problems occur.

bendrucker: Adding a log message now

bendrucker:
Totally, I'm open to addressing this in `WalkExpressions`. Not sure we can deal with it once we're at the expression level, since we need to know that we're in a special meta-argument.

bendrucker:
Wanted to get this fixed ASAP but I can try to follow up so we don't have to ignore these errors in rules.

PR #762, opened 2020-05-20
add terraform_comment_syntax rule

bendrucker:
This PR adds a `terraform_comment_syntax` rule that enforces the canonical `#` comment style as described in the Terraform syntax guide:

https://www.terraform.io/docs/configuration/syntax.html#comments

This was requested by a colleague who noticed that @takescoop's Terraform configuration contains a mix of `#` and `//` comments but that Terraform clearly recommends `#`.

Since this is an aesthetic preference, the rule is disabled by default.

In order to accomplish this, I've added a `Files` method to the runner which returns all files in the current module. I'm interested in some other syntax-oriented rules that would depend on that method, e.g. enforcing canonical block padding:

```tf
resource "foo" "bar" {
  attribute = "value"

  block {
    
  }
}

resource "foo" "baz" {
  # ...
}
```

### DIFF
--- a/tflint/runner.go
+++ b/tflint/runner.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"path/filepath"
 	"strings"
 	"sync"
 
@@ -235,6 +236,17 @@ func (r *Runner) LookupIssues(files ...string) Issues {
 // or nil if there path does not match any configuration.
 func (r *Runner) File(path string) *hcl.File {
 	return r.files[path]
+}
+
+// Files returns the raw *hcl.File representation of all Terraform configuration in the module directory.
+func (r *Runner) Files() map[string]*hcl.File {
+	result := make(map[string]*hcl.File)
+	for name, file := range r.files {
+		if filepath.Dir(name) == r.TFConfig.Module.SourceDir {
+			result[name] = file
+		}
+	}
+	return result
 }
 
 // EnsureNoError is a helper for processing when no error occurs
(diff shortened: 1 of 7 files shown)


tflint/runner.go
```diff
@@ -237,6 +238,17 @@ func (r *Runner) File(path string) *hcl.File {
 	return r.files[path]
 }
 
+// Files returns the raw *hcl.File representation of all Terraform configuration in the module directory.
+func (r *Runner) Files() map[string]*hcl.File {
```
wata727, line 242: 👍

wata727:
According to Terraform documentation:

> Automatic configuration formatting tools may automatically transform // comments into # comments, since the double-slash style is not idiomatic.

After reading this, I wondered if `terraform fmt` would convert the comments automatically, but it doesn't seem so...
If Terraform convert comments with `terraform fmt`, we probably don't need this rule.

bendrucker:
Agreed, I think anything detectable by `terraform fmt` or `terraform validate` should be a non-goal for tflint.

I suspect that in the real world a lot of users (if not most) have a a mix of comment styles. Double slash is pretty much everywhere except for shells. If your editor has a syntax definition for Terraform/HCL, the keyboard shortcut for opening a comment should use `#` but without that definition it'll likely default to `//`.

Enforcing one comment style would be breaking for `terraform fmt -check` and currently `fmt` just modifies whitespace. So it seems like a big undertaking for visual consistency.

The language about auto-formatting in the docs is definitely interesting, and comes from here:

https://github.com/hashicorp/terraform/commit/39579e8d0f3cae0f13490cd8ae177643db39e0c1

Seems like Terraform 0.12 would have been the best time to introduce this, given that users were already significantly rewriting their configs. Very possible that was being considered at the time and it was cut from the release.

Issue #768, opened 2020-05-21
feat: support custom messages when failing tflint checks


nitrocode:
Would be nice to be able to define custom messages in every rule to either overwrite or append tflint's default message

bendrucker:
Since messages are assembled in code, just allowing a static replacement seems a bit confusing. Can you provide a real world example of how you'd use this?

nitrocode:
```lang-hcl
rule "aws_resource_missing_tags" {
  enabled = true
  tags    = ["Foo", "Bar"]
  exclude = ["aws_autoscaling_group"] # (Optional) Exclude some resource types from tag checks
  msg     = "bro bro bro you're missing some really important tags. see this confluence page for our mandatory tags page and why each tag is useful. also see this link for our custom terraform module that implements this to make life easier."
}
```

bendrucker:
Hmm, how about a custom URL that replaces the self-link to the TFLint docs? That way if you have custom rule configuration you can link to a custom document that describes how to fix the issue.

nitrocode:
¯\_(ツ)_/¯

That would be more helpful and customizable than the current implementation. I'd still want to append the output at the very least. Like the example above, I'd want to share multiple links.

bendrucker:
Sounds good, an optional message that is printed after `\n\n` and the current text seems ok. Not of interest to me personally but happy to review a PR.

wata727:
Will share my thoughts at the moment.

I believe that in many cases this is the case where you want to apply the policy. So instead of supporting custom messages, I think it should be better to make defining policies easy. This is tracked by #1358.

However, for some rules, it may be useful to be able to set custom messages. In that case, you can address it by changing the rule implementation. There are currently no plans to provide this as a feature of TFLint CLI.

Closing this issue based on the above. If you want to discuss more, please open a new discussion instead of an issue. Thanks.

PR #769, opened 2020-05-22
[WIP] - Adding new terraform_blacklisted_resources rule

mveitas:
This rule adds the ability to prevent a resource of a certain type from being used. Each type specified has a custom message that will be displayed to the user that can be used to provide guidance on an alternative.

Fixes https://github.com/terraform-linters/tflint/issues/700

### DIFF
--- a/docs/rules/terraform_blacklisted_resources.md
+++ b/docs/rules/terraform_blacklisted_resources.md
@@ -1,0 +1,50 @@
+# terraform_blacklisted_resources
+
+Disallow `resource` declarations of certain types. Each type has a custom message will be displayed.
+
+## Example
+
+```hcl
+rule "terraform_blacklisted_resources" {
+  enabled = true
+  types   = {
+    aws_iam_policy_attachment = "Consider aws_iam_role_policy_attachment, aws_iam_user_policy_attachment, or aws_iam_group_policy_attachment instead."
+  }
+}
+```
+
+```hcl
+resource "random_id" "server" {
+  byte_length = 8
+}
+
+resource "aws_iam_policy_attachment" "test-attach" {
+  name       = "test-attachment"
+  users      = [aws_iam_user.user.name]
+  roles      = [aws_iam_role.role.name]
+  groups     = [aws_iam_group.group.name]
+  policy_arn = aws_iam_policy.policy.arn
+}
+```
+
+```
+$ tflint
+1 issue(s) found:
+
+Warning: `aws_iam_policy_attachment` resource type is blacklisted
+
+Consider aws_iam_role_policy_attachment, aws_iam_user_policy_attachment, or aws_iam_group_policy_attachment instead. (terraform_blacklisted_resources)
+
+  on modules.tf line 1:
+   1: resource "aws_iam_policy_attachment" {
+
+Reference: https://github.com/terraform-linters/tflint/blob/v0.16.0/docs/rules/terraform_blacklisted_resources.md
+```
+
+## Why
+Organizations may want to disallow some resource types from being used in their organization and provide feedback to 
+alternative resources. An example is the `aws_iam_policy_attachment` and preferring to use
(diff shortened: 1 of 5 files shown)


mveitas:
Wanted to get some eyes on this approach before I go further. 

I was going to take a look at the additional attribute `allow_in_modules` that was specified in #700 in the example to ensure that this won't break when "deep" module check is enabled

rules/terraformrules/terraform_blacklisted_resources.go
```diff
@@ -0,0 +1,77 @@
...
+	log.Printf("[TRACE] Check `%s` rule for `%s` runner", r.Name(), runner.TFConfigPath())
+
+	config := terraformBlacklistedResourcesRuleConfig{}
+	config.Types = make(map[string]string)
+
+	if err := runner.DecodeRuleConfig(r.Name(), &config); err != nil {
+		return err
+	}
+
+	// Since the resources are stored as a map, there is no guarantee of the order in which
```
bendrucker, line 54: This seems like something we should address in `AssertIssues` rather than requiring a rule to sort like this

docs/rules/terraform_blacklisted_resources.md
```diff
@@ -0,0 +1,50 @@
...
+
+Disallow `resource` declarations of certain types. Each type has a custom message will be displayed.
+
+## Example
+
+```hcl
+rule "terraform_blacklisted_resources" {
+  enabled = true
+  types   = {
+    aws_iam_policy_attachment = "Consider aws_iam_role_policy_attachment, aws_iam_user_policy_attachment, or aws_iam_group_policy_attachment instead."
```
bendrucker, line 11: What about a `map[string][]string`, where the key is the forbidden resource and the values are the suggested alternatives? I think it's best that we avoid putting literal messages in rule configuration, unless it's a meta-parameter that all rules can provide (https://github.com/terraform-linters/tflint/issues/768).

mveitas:
Thoughts on naming this `terraform_alternative_resources`?

mveitas:
Going back to the idea of include a message as part of the rule, this is something that would be valuable to be able to provide some context for each type specified. The message would be something that is "optional"

What are you thoughts about a rule structure as such:
```
rule "terraform_alternative_resources" {
  enabled = true
  types = {
    myprovider_identity = {
      alternatives = ["myprovider_user"],
      message      = "This resource will be deprecated on Sept 1, 2020. See https://myprovider.com/docs/terraform_provider for more information"
    }
  }
}
```

rules/terraformrules/terraform_blacklisted_resources.go
mveitas: I agree, I just could not see a way to make this happen easily. The right thing to do is to resolve the issue in `AssertIssues` first instead of polluting the rule.

Terraform stores the resources as a `map`, so there is no guaranteed order in the case of evaluation of resources. 
```
type Issue struct {
	Rule    Rule
	Message string
	Range   hcl.Range
	Callers []hcl.Range
}
```

Thoughts?

bendrucker:
Yeah, I think suggested or alternative or something similar sets the right tone. 

And you're right re: the message, since it does need to be on a per-resource basis I think that makes sense.

What does `types` represent here? Was expecting something more like Terraform's structure, i.e. `map[string]map[string]*Resource`, a map of provider -> resources, where `data.` prefixes data sources.

nitrocode:
Thanks @mveitas for the PR. I think `terraform_alternative_resources` is a much better name for this rule.

Any update on this?

bendrucker:
See https://github.com/terraform-linters/tflint/issues/878 for discussion on linting child modules and a proposed behavior change. Otherwise, this rule would behave in unexpected ways with child modules, and only trigger errors when the disallowed resource included a `var.*` reference.

nitrocode:
@bendrucker wait how come this was closed ? Is there a way to do this now using a different rule ?

bendrucker:
This is not merge-able or complete. If someone else wants to take it up in a rule set plugin that's fine.

Issue #715, opened 2020-04-14
Allow overriding severity for individual checks


sergei-ivanov:
At the moment, severity is an immutable property of each check. In my configuration I would like to elevate it for some checks and suppress for the others. I looked into tflint codebase, hoping that it could be fixed with a small PR, but it looks like it is a major cross-cutting concern that is going to affect all existing rules. I believe that we need some common configuration for all rules, for a start it can be the `enabled` flag and the `severity` value only, but it should be possible to extend it with more common attributes later if needed.

wata727:
Interesting. By the way, what rule do you want to change the severity of? I have thought the severity was immutable (e.g. The incorrect instance type is clearly an error).

sergei-ivanov:
For example, there are some basic checks like `terraform_dash_in_output_name` that are currently at _notice_ level, but I'd like to elevate them to _warning_ or even _error_ level, in order to enforce consistent naming.

wata727:
Thank you for sharing your use case. It makes sense.

snypelife:
Would this sort of feature also fall under a severity level?

e.g. 
```
$ tflint --severity-level warn

[Proceeds to show only messages for warnings and above]
```

digitalfiz:
I would also like this as a feature. Currently we are a bit more strict with our linting and we want things like missing variable descriptions to be error instead of warn. We have a few other we would like to adjust depending on the situation.

wata727:
Will share my thoughts at the moment.

I believe that in many cases this is the case where you want to apply the policy. So instead of overriding severity, I think it should be better to make defining policies easy. This is tracked by #1358.

However, for some rules, it may be useful to be able to override the severity. In that case, you can address it by changing the rule implementation. There are currently no plans to provide this as a feature of TFLint CLI.

Closing this issue based on the above. If you want to discuss more, please open a new discussion instead of an issue. Thanks.

sergei-ivanov:
Our "policy" is effectively implemented through a set of already existing core rules. There is only one problem: we cannot enforce it, because most of those rules are at "notice" level, and as such don't trigger a linting failure. There is no compelling reason to rewrite exactly the same rules as custom rules, if the only change is overriding the severity. 

Maybe we can at least compromise on adding a command line option or configuration file option to set the minimum severity level that triggers overall linting failure (non-zero exit code). See for example `--minimum-severity` option in tfsec.

bendrucker:
That is tracked in #1556

PR #771, opened 2020-05-23
Allow rules to override the declared default severity level

mveitas:
This adds the ability for rules to override the severity. 

An example usage - this will override the rule's severity to `Error`. By default the severity is declared as "Notice"
```
  enabled  = true
  severity = "Error"
}
```

Fixes https://github.com/terraform-linters/tflint/issues/715

### DIFF
--- a/formatter/pretty.go
+++ b/formatter/pretty.go
@@ -36,7 +36,7 @@ func (f *Formatter) printIssueWithSource(issue *tflint.Issue, sources map[string
 	fmt.Fprintf(
 		f.Stdout,
 		"%s: %s (%s)\n\n",
-		colorSeverity(issue.Rule.Severity()), colorBold(issue.Message), issue.Rule.Name(),
+		colorSeverity(issue.Severity), colorBold(issue.Message), issue.Rule.Name(),
 	)
 	fmt.Fprintf(f.Stdout, "  on %s line %d:\n", issue.Range.Filename, issue.Range.Start.Line)
 
--- a/tflint/config.go
+++ b/tflint/config.go
@@ -56,9 +56,10 @@ type Config struct {
 
 // RuleConfig is a TFLint's rule config
 type RuleConfig struct {
-	Name    string   `hcl:"name,label"`
-	Enabled bool     `hcl:"enabled"`
-	Body    hcl.Body `hcl:",remain"`
+	Name      string   `hcl:"name,label"`
+	Enabled   bool     `hcl:"enabled"`
+	Serverity string   `hcl:"severity,optional"`
+	Body      hcl.Body `hcl:",remain"`
 }
 
 // PluginConfig is a TFLint's plugin config
--- a/tflint/issue.go
+++ b/tflint/issue.go
@@ -8,10 +8,11 @@ import (
 
 // Issue represents a problem in configurations
 type Issue struct {
-	Rule    Rule
-	Message string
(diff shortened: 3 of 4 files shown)


mveitas:
None of the test have been updated as these changes require ever test to now specify the `Serverity` as part of the assertions for the emitted issues.

Before updating all of the tests to check for the `Severity` field on the issue, I wanted to get some eyes on this approach. Most of the aws tests will be automatically be updated as they are generated via code. There are a large number of tests that will need to be manually updated

This is an example of the test failure:
```
 --- FAIL: Test_TerraformWorkspaceRemoteRule/terraform.workspace_in_locals_with_remote_backend (0.00s)
        testing.go:61: Expected issues are not matched:
               tflint.Issues{
                &{
                        Rule:     &terraformrules.TerraformWorkspaceRemoteRule{},
                        Message:  "terraform.workspace should not be used with a 'remote' backend",
            -           Severity: "",
            +           Severity: "Warning",
                        Range:    hcl.Range{Filename: "config.tf", Start: hcl.Pos{Line: 6, Column: 6}, End: hcl.Pos{Line: 6, Column: 25}},
                        Callers:  nil,
                },
              }
```

mveitas:
If this approach works the ability to add a custom message per rule would be able to be accomplished in the same manner: https://github.com/terraform-linters/tflint/issues/768

wata727:
If you want to extend the severity or message, you should probably extend the `Rule` rather than the `Issue`. We should also consider how to pass options such as severity to the plugin.

To be honest, I'm wondering if TFLint should allow for severity or message extensions, but even if I do, I have a lot to think about and I've been pending.

mveitas:
I'm still getting familiar with both Go and TFLint code (so excuse my asking so many questions). 

I started down that road wasn't able to find a way as all of the information from the configuration is stored in the `RuleConfig`.

mveitas:
> To be honest, I'm wondering if TFLint should allow for severity or message extensions, but even if I do, I have a lot to think about and I've been pending.

So here is a concrete use case for allowing to override the severity of messages:

Our company has a number of teams that are using Terraform and we recently started to enforce some basic standards with regards to formatting and style. We would love to have a consistent set of rules applied to our code, but this is going to take some time and different teams move at a different pace. We need the flexibility to set severity levels for different teams as they make this journey.

Any further thoughts?

wata727:
Umm...

In my opinion, if you want to extend this, I think it's better to write a ruleset plugin and use it. This is because my plan is that sometime I'd like to cut style rules as a plugin.

On the other hand, given that famous linters such as ESLint provide extensions of severity as configurations, I think that providing such extensions as a feature of TFLint is not such a bad idea.

However, some architectural changes are required to make this happen. First, when initializing the rule, we will need to accept the configuration as an argument. That seems like something we have to do, but nothing has been considered yet.

wata727:
This PR has been out of date since it was opened. I'll close this.

Messages and severity can be overridden by plugins. If you want to fine-tune the impact of style issues, I recommend extending the plugin. You may want to do this in a simpler way than a plugin, so we may consider a new feature in the future.

bendrucker:
Please don't start conversations on closed PRs. If you have new information to add, you can add a comment to #715. Otherwise, if you just want to indicate that it's a feature that would be important to you, you can 👍🏻 the original post.

Yes, modifying the plugin is not a simple intervention.

PR #776, opened 2020-05-27
Bump tf to v0.12.26

chenrui333:


lkysow:
@chenrui333 now I see why you're so on the ball with Atlantis 😂

PR #782, opened 2020-06-02
Update extend.md

jpreese:
After trying to get a local plugin dir to work, I noticed that the plugin dir should start with a period!


wata727:
Good catch 👍

PR #784, opened 2020-06-04
add terraform_deprecated_index (disallows foo.0)

bendrucker:
This adds a rule that disallows the legacy dot/attribute syntax for accessing indexes in a list. This syntax was officially recommended in v0.11:

https://www.terraform.io/docs/configuration-0-11/interpolation.html

And because most official providers still use 0.11 syntax in their docs, I still see this used quite a bit:

https://www.terraform.io/docs/providers/kubernetes/r/service_account.html
https://www.terraform.io/docs/providers/aws/r/acm_certificate_validation.html

It is no longer present in the 0.12 documentation and didn't even make the HCL spec:

https://github.com/hashicorp/hcl/issues/305

### DIFF
--- a/docs/rules/README.md
+++ b/docs/rules/README.md
@@ -68,6 +68,7 @@ These rules suggest to better ways.
 |Rule|Enabled by default|
 | --- | --- |
 |[terraform_deprecated_interpolation](terraform_deprecated_interpolation.md)|✔|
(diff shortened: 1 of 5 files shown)


wata727:
Good job!

I like this rule, but I'm a bit concerned about how much this syntax is used. If it's already widely used, it may be nonsense to encourage all to be fixed. What do you think?

PR #789, opened 2020-06-06
Allow no extensions on windows

jpreese:
Windows does not _require_ the `.exe` extension and this was causing some confusing when trying to create the `tflint` binary using `-o` with go build.


jpreese:
For kicks and giggles I tried tagging this on my own fork and pulled it in there to see if it truly solved the issue-- but I'm seeing the same behavior which was quite unexpected. Going to evaluate some more.

jpreese:
Don't code tired friends, this is fine. Wasn't even testing against these changes.

PR #801, opened 2020-06-17
Add terraform_standard_module_structure rule

bendrucker:
This PR creates a rule that enforces the Terraform standard module structure:

https://www.terraform.io/docs/modules/index.html#standard-module-structure

The rule itself ensures that main, variables, and outputs file are created. It also ensures that variable/output blocks are included in the corresponding file. File checks are skipped for JSON-only projects and block checks are skipped when the declaration was in a JSON file. 

Outside of the rule implementation, I've also added logic to `TestRunnerWithConfig` that detects the directory so that rules can test configuration that is in a directory other than `.`.

### DIFF
--- a/docs/rules/README.md
+++ b/docs/rules/README.md
@@ -78,4 +78,5 @@ These rules suggest to better ways.
 |[terraform_naming_convention](terraform_naming_convention.md)||
 |[terraform_required_version](terraform_required_version.md)||
 |[terraform_required_providers](terraform_required_providers.md)||
+|[terraform_standard_module_structure](terraform_standard_module_structure.md)||
 |[terraform_workspace_remote](terraform_workspace_remote.md)|✔|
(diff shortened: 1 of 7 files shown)


mveitas:
The standard module structure also enforces that a readme exist within the module (`README` or `README.md`). Thoughts on including this?

bendrucker:
I did think about that but chose to skip it:

* Validating non-Terraform files would be new and possibly unexpected
* Readmes are optional for nested modules and TFLint cannot distinguish a nested module if it is called directly on it (`cd modules/foo && tflint .`)

bendrucker:
Other related ideas here:

* Rule for community filename conventions for root modules (`providers.tf`, `backend.tf`)
* Config for `required_version`/`required_providers` rules allowing them to enforce the conventional filename (`versions.tf`)

Issue #788, opened 2020-06-05
error on commas in AWS Tag values


klauern:
There's a formatting specification for Tag Keys and Values.  One issue with applying some changes is that you are disallowed from using certain characters (notably the `,`) in a Tag Value, and this is not being caught by `tflint` currently.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-resource-tags.html#w2ab1c25c10d254c13c17

RyanDhuse:
I tripped on this one today with the "#" character.  I'd be happy to put some effort into a merge request.  Do the maintainers have some general direction on how this should be done?

bendrucker:
Hey, working on making tags rules easier to write. The existing tags rule (missing_tags) is in pretty rough shape. https://github.com/terraform-linters/tflint/pull/803 should lay the ground work. It adds a `tags` package within `awsrules` for dealing with tags that other tag-related rules can use.

https://github.com/terraform-linters/tflint/blob/master/rules/awsrules/aws_resource_missing_tags.go

I can see this being a `aws_invalid_tags` rule. In the mean time, might be helpful to start thinking about how to actually do this validation. If aws-sdk-go can do it (or even if not), you might also look into implementing this as field validator in terraform-provider-aws. TFLint should ideally suggest optional best practices and leave static validation to providers. The one exception is when a network connection would be required for validation, which Terraform validation forbids.

wata727:
Yes, I think static validation should ideally be included in providers.
On the other hand, I think there are some challenges in static validations by providers:

- Since it is often performed in a later process such as CI, it is difficult to try easily.
- Validation error cannot be ignored, so it is difficult to incorporate specifications that change frequently.

I think providers should be able to solve this kind of problem in the future, but until then TFLint doesn't seem to be a bad idea to tackle this problem. So I'm positive about implementing this kind of rule.

PR #803, opened 2020-06-18
awsrules: add tags package with generator

bendrucker:
This PR adds a generator as a child package in a similar style to terraform-provider-aws:

https://github.com/terraform-providers/terraform-provider-aws/tree/master/aws/internal/keyvaluetags

In responding to https://github.com/terraform-linters/tflint/issues/788 I realized that the working in aws_missing_tags is not very adapatable to other rules. It templates an entire file with several hundred lines to substitute one list of resources types.

Instead, I've created a package to handle tags and a sub-package to generate a file that exports the list of tagged resource types. From there I plan to pull out the tag evaluation functionality into the `tags` package. 

This may also pave the way for breaking apart the `tools` packages, moving generator packages to where they are used, and invoking them via `//go:generate` comments.

Wanted to open this early to check on interesting in breaking apart `tools` and whether there's a reason having separate dependencies is important/required.


bendrucker:
@wata727 Wanted to check in on this one before I go any further now that I have something working. If you're interested, I can expand this PR to similarly relocate the other `tools` packages.

wata727:
Looks great. The reason for defining another dependency under `tools` was to avoid a build error that occurs when adding `terraform-provider-aws` to the dependency. See https://github.com/terraform-linters/tflint/pull/337
At that time, we were using Azure Pipelines, so that might be the problem.

It would be nice if the dependencies were integrated and everything was generated by `go generate`.

client/aws.go
```diff
@@ -26,13 +26,13 @@ import (
...
-//go:generate mockgen -source ../vendor/github.com/aws/aws-sdk-go/service/rds/rdsiface/interface.go -destination aws_rds_mock.go -package client
-//go:generate mockgen -source ../vendor/github.com/aws/aws-sdk-go/service/ecs/ecsiface/interface.go -destination aws_ecs_mock.go -package client
+//go:generate mockgen -destination aws_ec2_mock.go -package client github.com/aws/aws-sdk-go/service/ec2/ec2iface EC2API
+//go:generate mockgen -destination aws_elasticache_mock.go -package client github.com/aws/aws-sdk-go/service/elasticache/elasticacheiface ElastiCacheAPI
+//go:generate mockgen -destination aws_elb_mock.go -package client github.com/aws/aws-sdk-go/service/elb/elbiface ELBAPI
+//go:generate mockgen -destination aws_elbv2_mock.go -package client github.com/aws/aws-sdk-go/service/elbv2/elbv2iface ELBV2API
+//go:generate mockgen -destination aws_iam_mock.go -package client github.com/aws/aws-sdk-go/service/iam/iamiface IAMAPI
+//go:generate mockgen -destination aws_rds_mock.go -package client github.com/aws/aws-sdk-go/service/rds/rdsiface RDSAPI
+//go:generate mockgen -destination aws_ecs_mock.go -package client github.com/aws/aws-sdk-go/service/ecs/ecsiface ECSAPI
```
wata727, line 35: Wow! It was a big problem for me that I had to run `go mod vendor` for mockgen. Great!

bendrucker:
Awesome! I'll make some time to do the same with the other generators. Figured it was a dependency issue.

go.mod
```diff
@@ -13,12 +13,16 @@ require (
...
 	github.com/hashicorp/terraform v0.12.26
+	github.com/hashicorp/terraform-plugin-sdk v1.14.0 // indirect
 	github.com/jessevdk/go-flags v1.4.0
 	github.com/mattn/go-colorable v0.1.6
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/sourcegraph/go-lsp v0.0.0-20181119182933-0c7d621186c1
 	github.com/sourcegraph/jsonrpc2 v0.0.0-20190106185902-35a74f039c6a
 	github.com/spf13/afero v1.2.2
 	github.com/terraform-linters/tflint-plugin-sdk v0.1.2-0.20200615160547-c1d3caf80fe0
+	github.com/terraform-providers/terraform-provider-aws v2.65.0+incompatible // indirect
```
wata727, line 24: @bendrucker I noticed this line is removed when run `go mod tidy` because the code required this dependency is ignored on the build phase. This seems a little inconvenient. Any thoughts?

bendrucker: Apparently this is a special case of `+build ignore` and any other constraint name will work.

https://github.com/golang/go/issues/29598
https://github.com/golang/go/issues/38061#issuecomment-603917218

bendrucker: Should be resolved by https://github.com/terraform-linters/tflint/pull/809

wata727: Awesome!

Issue #770, opened 2020-05-22
Support for Terraform 0.13


mveitas:
Beta releases of Terraform 0.13 are going to start shipping on June 3: https://github.com/hashicorp/terraform/issues/25016

wata727:
In my understanding,  the changes don't have a big impact on TFLint, but it seems good to start investigating early when the beta is released.

bendrucker:
It's possible that module instances could have some impact, but otherwise agreed that the work should be minimal. Since Terraform versions are for the CLI and don't apply to the go packages, there's always some possibility of breakage. Bits of 0.13 have been landing in master for a few months now so we don't have to wait for actual beta releases to try it out either.

danieldreier:
Beta 1 was just released today!

bendrucker:
👋 Hey Daniel, eager to try it out! Will open a draft PR upgrading TFLint soon.

bendrucker:
Yeah that’s expected. TFLint consumes Terraform as a library. The dependency will need to be updated to support new features.

That’s not to say everything will necessarily work straight away, but simple schema errors like you posted should be immediately addressable. 

I’ll spend some time later today getting a branch and draft PR open to track this, including build instructions.

gmazelier:
Great!

bendrucker:
Opened https://github.com/terraform-linters/tflint/pull/804 which will track the effort to update for 0.13. Still some additional work required to handle the fact that modules can have multiple instances. Some more guidance on the 0.13 process, copied from the PR:

<hr>

For specific issues (crashes, errors, etc.), please open a new issue and reference that you are testing with 0.13. Please include example configuration.

For general discussion, please comment on the main issue.

<hr>

This will apply once #804 is passing tests. For now please sit tight, but PRs are welcome!

bendrucker:
#804 now build/passes tests, please check that out and build locally if you're trying out Terraform 0.13. Now would be a good time to start reporting any unexpected errors for valid configurations (as new issues). As described in #804, TFLint is still not aware of multiple module instances. It won't try to evaluate the `for_each` or `count` attributes in module calls which means any attribute using `each.*` or `count.*` is unknown. Resources treat these values similarly and don't statically analyze for_each/count except to detect when a resource is skipped altogether.

wata727:
#804 was merged. It will be included in v0.19.

PR #804, opened 2020-06-18
Terraform v0.13

bendrucker:
This pull request updates TFLint for Terraform v0.13 compatibility and will serve as a tracking branch during the [0.13 beta](https://github.com/hashicorp/terraform/issues/25016). 

**Status:** Compatible with Terraform 0.12 configurations, and should run successfully with new 0.13 features (`module.{for_each,count}`). Attributes that use `each` or `count` will be unknown and will not be evaluated, even if the `for_each` or `count` expression is static.
 
<hr>

TFLint uses Terraform as a library and never calls the `terraform` binary. Because there are no breaking [changes](https://github.com/hashicorp/terraform/blob/master/CHANGELOG.md) in the language itself (only the CLI), TFLint should remain compatible with Terraform modules written for Terraform 0.12.x. 

## Building

Currently, you must build TFLint locally to use it with Terraform 0.13 modules. We may consider publishing release candidates when Terraform does the same.

```sh
make build
```

## Issues

For specific issues (crashes, errors, etc.), please open a new issue and reference that you are testing with 0.13. Please include example configuration.

For general discussion, please comment on the main issue (#770).

## Pull Requests

If you'd like to help prepare TFLint for Terraform 0.13, please target PRs against the `terraform-v0.13` branch (which this PR tracks).


bendrucker:
Left a TODO behind about next steps for getting this working. TFLint currently can't find any modules as a consequence of `for_each` changing around the structure.

bendrucker:
This is now in a working state. Somewhat predictably, the `each` and `count` objects are not defined in the parent context, so a module variable that uses either will be unknown. Evaluating `for_each`/`count` and expanding modules to module instances will take a more ambitious refactoring effort. I've added a test showing how using `for_each` prevents an invalid instance type from being detected. With more robust support for module iteration, that configuration should result in an issue referencing the full path, including the `each.key` value.

bendrucker:
Terraform 0.13 is now GA, I'll get this ready for release

tflint/runner.go
```diff
@@ -51,22 +51,25 @@ func NewRunner(c *Config, files map[string]*hcl.File, ants map[string]Annotation
...
-					Env: getTFWorkspace(),
-				},
-				Config:             cfg.Root,
-				VariableValues:     prepareVariableValues(cfg, variables...),
-				VariableValuesLock: &sync.Mutex{},
-			},
-		},
+		// TODO: As described in the godoc for UnkeyedInstanceShim,
+		// it will need to be replaced now that module.for_each is supported
+		ctx:         ctx.WithPath(cfg.Path.UnkeyedInstanceShim()),
```
wata727, line 72: 👍

PR #838, opened 2020-07-08
Add `Backend()` to plugin server

pd:


pd:
The failing checks are due to the corresponding tflint-plugin-sdk changes; I suppose those have to land first before this can successfully build.

wata727:
@pd Looks good. I have merged the changes to tflint-plugin-sdk. Can you upgrade the dependency to pass the CI?

pd:
@wata727 Updated the dependency. If you want to add a new tag to the SDK, I can update go.mod here to reference that instead, but this is green now either way.

wata727:
Thanks!

Issue #835, opened 2020-07-01
[tflint] output logging is not sufficient 


pduleba:
### Version
TFLint version 0.17.0

### Command
`TFLINT_LOG=debug tflint -c tflint.hcl --module --deep`

### Current
Using as example project listed below following problems/issues has been spotted:
- tflint is not printing path do `data.tf` which contains problem. It is harder to find problematic terraform data file.
- tflint is not able to detect problems inside of `service` module, even with `--deep` and `--module` flags enabled.

### Expected
- module `service` should be inspected
- tflint should provide anyway absolute path location to problematic `data.tf`

#### Project structure

**main.tf**
```
data "aws_caller_identity" "current" {}

variable "environment" { default = "test" }

module "service" {
  source = "./service"

  environment = var.environment
}

output "dummy_response" { value = module.service.dummy_response }
```

**service/main.tf**
```
data "aws_caller_identity" "current" {}

variable "environment" {}
variable "account_id" { default = 0 }

output "dummy_response" { value = "${var.environment}-${var.account_id}" }
```

**tflint.hcl**
```
config {
  module = true
  #deep_check = true # Executed for env using flag
  force = false
}

rule "terraform_deprecated_interpolation" {
  enabled = true
}
rule "terraform_deprecated_index" {
  enabled = true
}
rule "terraform_unused_declarations" {
  enabled = true
}
rule "terraform_comment_syntax" {
  enabled = false
}
rule "terraform_documented_outputs" {
  enabled = false
}
rule "terraform_documented_variables" {
  enabled = false
}
rule "terraform_typed_variables" {
  enabled = false
}
rule "terraform_module_pinned_source" {
  enabled = true
}
rule "terraform_naming_convention" {
  enabled = true
}
rule "terraform_required_version" {
  enabled = false
}
rule "terraform_required_providers" {
  enabled = false
}
rule "terraform_workspace_remote" {
  enabled = true
}
```

### Output
pduleba@home:~/workspace/repo/test$ TFLINT_LOG=debug tflint -c tflint.hcl --module --deep 
15:21:54 config.go:90: [INFO] Load config: tflint.hcl
15:21:54 config.go:267: [DEBUG] Config loaded
15:21:54 config.go:268: [DEBUG]   Module: true
15:21:54 config.go:269: [DEBUG]   DeepCheck: false
15:21:54 config.go:270: [DEBUG]   Force: false
[...]

PR #840, opened 2020-07-10
Fix module resolution when terraform init is invoked from another directory

mkielar:
This resolves the bug with running tflint with `--module` as described in https://github.com/terraform-linters/tflint/issues/835

- Removed code section in `tflint/loader.go` that caused the issue
- Added integration tests that cover running tflint with `--module` in two scenarios:
  1. Inside the folder where `terraform apply` is usually applied
      ```
      cd <dir>
      terraform init
      tflint --module
      ```  
  2. In the _"project root"_, by specifying the path to the folder where `terraform apply` is usually applied in command line args.
      ```
      terraform init <dir>
      tflint --module <dir>
      ```
- added `.idea/` to `.gitignore`, hope you don't mind ;-)


bendrucker:
Thank you for chasing this down, both for the issue comments and the fix!

Cleaned this up to represent the case you're tying to fix a bit more directly. 

Seems like `environments/` is perhaps something that came from your Terraform project and I went with `root` and `modules` to match official Terraform terminology. Flattening the directory structure and adjusting the names hopefully makes the difference between the two cases more apparent, whereas it took me a good bit of squinting to figure out how they differed initially. If I've changed your intent please let me know.

I will take a closer look at the change itself later today but it should be good to merge. Given that this is intended to fix #835, would be nice to see output before and after this change to help illustrate the improvement.

wata727:
I haven't checked this change in detail yet, but before merging it, we need to think about this change:  https://github.com/terraform-linters/tflint/pull/555

bendrucker:
Good catch, #555 has no test but this would definitely cause a regression. I think we need a test there before proceeding. To made a test case with absolute paths, we could presumably have a test that copies an existing case dir to a temp dir, rewrites the `modules.json` file to use absolute paths, and then runs the normal integration test process.

bendrucker:
Actually I take that back, #555 _reverts_ the value of `dir` to `record.Dir` if `Dir` is an absolute path. This change just means that would now happen always, for both relative and absolute paths. Doesn't seem like a regression.

bendrucker:
I've added support for passing `Dir` as an argument to TFLint rather than using `Chdir` for integration test cases. The case for running `terraform init` in a module but then invoking TFLint from a parent directory of that module seemed like a possible regression, but all looks good there.

wata727:
Thanks, @mkielar @bendrucker. Understood this change.
Certainly, this change doesn't seem a regression and makes sense.

mkielar:
Hi,
I've seen you've added this test case:
```
{
	Name:        "init from cwd, run from parent",
	Command:     "./tflint --format json --module",
	Dir:         "init-cwd/root",
	DirArgument: true,
},
```
and that it fails, but only if you comment all the other use cases. The reason it works is because you have the `init from cwd` just before it, and it changes the directory to `init-cwd/root` which is why the next one passes. 

I fixed the chdir inside the loop, but now the test fails. I think we also need to fix the loader for `modules.json`.

bendrucker:
Thanks for the fix there, was definitely expecting that to fail and should've looked closer...

I do agree that we shouldn't try to support configurations that Terraform itself would not create/support. I can adjust the tests a little and merge.

rajuvadarevu:
I still face this error:

I have modules defined in **examples/jenkins-sandbox/main.tf** 
I ran **terraform -chdir=examples/jenkins-sandbox init**
which initialized all the modules present in **examples/jenkins-sandbox/main.tf** file

But when I ran **tflint --module examples/jenkins-sandbox**, it still gives me error:
```
Failed to load configurations. 1 error(s) occurred:

Error: `azure-vm` module is not found. Did you run `terraform init`?

  on examples/jenkins-sandbox/main.tf line 7, in module "azure-vm":
   7: module "azure-vm" {
```

But if I,
```
cd examples/jenkins-sandbox/
tflint --module .
```

it works fine.

Any solution for this issue ?

PR #848, opened 2020-07-20
Expose `Server.ModuleCalls` for SDK

pd:
See the corresponding PR to the SDK this depends on:

https://github.com/terraform-linters/tflint-plugin-sdk/pull/50

Version constraints may be a bit different over the wire: the type exposed
by go-version can't necessarily be exactly reconstructed, as they've
been normalized under the hood. The original value is still accessible
through the usual hcl.Body.PartialContent or similar functions, if
needed.

In Terraform 0.12, I think the Count and ForEach struct members
will never actually be set, but the corresponding Terraform types have
them, so I've included them here. Maybe they'll Just Work once 0.13 is
out of beta and support is added to tflint.


wata727:
I have merged the changes in the SDK, so please update its dependency in this PR :)

wata727:
Thanks!

PR #850, opened 2020-07-20
terraform_required_providers: warn on provider.version

### LINKED ISSUE #839: terraform_required_providers errs with multiple copies of the same provider
(currently?) Terraform won't let you use multiple versions of the same provider for a project so the second provider is implicitly pinned to the version of the first provider. We have some instances of this where a Terraform project touches multiple AWS regions (like RDS replicas) and I think it's not necessarily desirable to pin the version multiple times since the first pinned version is always used anyway

If I pin to 2.64.0 & 2.65.0, respectively, `No provider "aws" plugins meet the constraint "2.64.0,2.65.0".`

```hcl
provider "aws" {
  assume_role {
    role_arn = "arn:aws:iam::123:role/admin"
  }

  version = "~> 2.0"
  region  = "us-east-2"ee

provider "aws" {
  assume_role {
    role_arn = "arn:aws:iam::123:role/admin"
  }

  alias  = "useast1"
  region = "us-east-1"
}
```

## Version
TFLint v0.17.0
Terraform v0.12.25

Maybe it some cases it's desirable to major version pin one provider and minor version pin another so possibly this could be a configurable option but in most cases I don't think someone would want to duplicate version requirements

bendrucker:
This PR updates the required providers rule to warn when a `provider` block specifies a `version` attribute. While this does have the effect of constraining the installed provider version, users should be moving to `required_providers` because declaring a new copy of the provider is likely to have unexpected effects for child modules.

New logic is:

* One issue is emitted for each provider _name_ that is missing from required_providers. If multiple providers for a given name are declared, the first text range will be used.
* One issue is emitted for each provider _instance_ that has a `version`, regardless of whether it also appears in `required_providers`.

This means that users moving from earlier versions with of Terraform with  `provider "foo" { version = "" }` in their configurations will see two issues that share an underlying fix. While it's possible to give more targeted messaging, there's a lot of possible cases here and keeping the rule simple seemed worthwhile.

Closes #839

### DIFF
--- a/docs/rules/terraform_required_providers.md
+++ b/docs/rules/terraform_required_providers.md
@@ -10,7 +10,7 @@ rule "terraform_required_providers" {
 }
 ```
 
-## Example
+## Examples
 
 ```hcl
 provider "template" {}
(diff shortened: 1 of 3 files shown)


docs/rules/terraform_required_providers.md
```diff
@@ -10,7 +10,7 @@ rule "terraform_required_providers" {
 }
 ```
 
-## Example
+## Examples
 
 ```hcl
 provider "template" {}
```
wata727, line 16: The issue message has been updated, so it would be nice to reflect it in the example output as well.

PR #851, opened 2020-07-21
-C throws error, should be -c

Stromweld:
-C throws error, should be -c

### DIFF
--- a/install_linux.sh
+++ b/install_linux.sh
@@ -66,7 +66,7 @@ if [[ $os == "windows"* ]]; then
 else
   echo "Installing /tmp/tflint to /usr/local/bin..."
   sudo mkdir -p /usr/local/bin
-  sudo install -b -C -v /tmp/tflint /usr/local/bin/
+  sudo install -b -c -v /tmp/tflint /usr/local/bin/
   retVal=$?
   if [ $retVal -ne 0 ]; then
     echo "Failed to install tflint"


bendrucker:
If there's an error, please share it and more about your setup. Perhaps `-C` (`--compare`) should be removed for broader compatibility, but it is assuredly not correct to replace it with `-c`:

https://linux.die.net/man/1/install

bendrucker:
Closing b/c of no response, if you can share more background in support of the change we can reconsider

Stromweld:
sorry for the late response I was trying to run this in our CI/CD pipeline on the hashicorp/terraform docker image. After adding curl,unzip,sudo I ran `curl -L https://raw.githubusercontent.com/terraform-linters/tflint/master/install_linux.sh | sh` and got error: 
```
/ # curl -L https://raw.githubusercontent.com/terraform-linters/tflint/master/install_linux.sh | sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0os=linux_amd64


100  2097  100  2097    0     0   5268      0 --:--:-- --:--:-- --:--:--  5255
Downloading TFLint v0.18.0
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   633  100   633    0     0   1935      0 --:--:-- --:--:-- --:--:--  1929
100 11.0M  100 11.0M    0     0   263k      0  0:00:43  0:00:43 --:--:--  219k
Download was successfully


Unpacking /tmp/tflint.zip ...
Archive:  /tmp/tflint.zip
Installing /tmp/tflint to /usr/local/bin...
install: unrecognized option: C
BusyBox v1.31.1 () multi-call binary.

Usage: install [-cdDsp] [-o USER] [-g GRP] [-m MODE] [-t DIR] [SOURCE]... DEST

Copy files and set attributes

	-c	Just copy (default)
	-d	Create directories
	-D	Create leading target directories
	-s	Strip symbol table
	-p	Preserve date
	-o USER	Set ownership
	-g GRP	Set group ownership
	-m MODE	Set permissions
	-t DIR	Install to DIR
Failed to install tflint
```

Stromweld:
alpine must use an older version of `install` command. I found this on my mac in the man pages for install:
```
 -C      Copy the file.  If the target file already exists and the files are the same, then don't change the modification time of the target.

     -c      Copy the file.  This is actually the default.  The -c option is only included for backwards compatibility.
```
lowercase -c would work the same and be backwards compatible with older versions of install.

bendrucker:
Thanks, looking into this and will get that Docker image added to the tests to be sure.

Stromweld:
I was running latest hashicorp/terraform docker image which looks to be running on alpine 3.12.

bendrucker:
Merged as https://github.com/terraform-linters/tflint/commit/88f5a01be415f9e7bbf2b2f83a539049bb88c171, along with testing

PR #875, opened 2020-08-15
Capability to only run explicitly-provided rules

markliederbach:
# Context
We have a use case where we only want to run specific rules against our terraform repositories, and leave further linting to other workflows. As such, we need a way to disable all default rules, and only allow the ones we've specifically enabled via the config or CLI arguments.

# What I Changed
A new argument/config option called `--only` (boolean) has been added, which is used in the `NewRules` part of the `rules/provider.go` to determine whether a rule should be enabled if it wasn't found in the provided config.

## CLI example usage
```zsh
tflint --only --enable-rule aws_instance_invalid_type
```
- This will _only_ run the provided rule, and not the other 770 default rules.

## Config example usage
```hcl
config {
  only = true
  # other options here...
}

rule "aws_instance_invalid_type" {
  enabled = true
}
```
- Again, will _only_ run the provided rule, and not the other 770 default rules.

Additionally, I've added a warning message when this mode is enabled and no rules were provided.

Tests have also been updated to reflect the new option.


wata727:
Makes sense, but I think the `--explicit-rules-mode` option feels a bit redundant. How about passing a rule name with an option like `--only`?

markliederbach:
I like that, making the change.

markliederbach:
Updated, let me know if there's anything else.

markliederbach:
Whoops, sorry. I misread. You wanted it to replace the use of `--enable-rule`.

markliederbach:
@wata727  I've modified the CLI option to accept rules like the following:
```zsh
tflint --only aws_instance_invalid_type
```
I've also added a warning message when `--enable-rule` or `--disable-rule` are used in addition to `--only`, because those options will be ignored as a result.

Let me know what you think.

docs/guides/advanced.md
```diff
@@ -57,3 +57,28 @@ You can use the `--ignore-module` option if you want to skip inspection for a pa
...
+```
+$ tflint --only aws_instance_invalid_type --only aws_instance_invalid_ami
+```
+**Note:** usage of `--only` will ignore any other rules defined via command line via `--enable-rule` or `--disable-rule`.
+
+You can also set `only = true` in the config file. Using this method, any rules enabled in your config file will implicitly become `--only` rules, and all other defaults will be ignored.
+
+```hcl
+config {
+  only = true
```
wata727, line 77: In the config file, it seems more natural to be able to set with an alias like `disabled_by_default = true`. What do you think?

rules/provider.go
```diff
@@ -104,12 +108,17 @@ func NewRules(c *tflint.Config) []Rule {
 				log.Printf("[DEBUG] `%s` is disabled", rule.Name())
 			}
 			enabled = r.Enabled
+		} else if c.Only {
+			enabled = false
```
wata727, line 112: We need to do the same when applies config to the plugin because rules by a plugin are not ignored:

https://github.com/terraform-linters/tflint/blob/70c764983a1ccdde853d3eac2085702aa5e4023e/tflint/config.go#L151-L161

wata727:
I also noticed that using the `--only` option and config file leads to a bit strange behavior. For example:

**ec2.tf**
```hcl
  instance_type = "t4.micro"
}

resource "aws_instance" "bar" {
  instance_type = "t1.micro"
}
```

**.tflint.hcl**
```hcl
rule "aws_instance_invalid_type" {
  enabled = true
}
```

```
$ tflint --only aws_instance_previous_type
2 issue(s) found:

Error: "t4.micro" is an invalid value as instance_type (aws_instance_invalid_type)

  on ec2.tf line 2:
   2:   instance_type = "t4.micro"

Warning: "t1.micro" is previous generation instance type. (aws_instance_previous_type)

  on ec2.tf line 6:
   6:   instance_type = "t1.micro"

Reference: https://github.com/terraform-linters/tflint/blob/v0.18.0/docs/rules/aws_instance_previous_type.md
```

In this case, it may be useful to ignore the `enabled = true` of the config file. What do you think?

In order to do that, we have to change the `Config` generated from the config file by CLI options, rather than just merging two `Config` (config file and CLI options). This may require a slightly more complicated implementation.

rules/provider.go
markliederbach: I've created https://github.com/terraform-linters/tflint-plugin-sdk/pull/54 which will allow us to pass `DisabledByDefault` down the the plugin. And with the SDK, it will perform the same check as in providers to disable rules if not provided by the config directly.

Once that's merged and bumped, I'll commit my changes to pass that in here (as well as bumping `go.mod` for the package).

PR #888, opened 2020-08-24
Rule - Terraform resources have required providers 

mcuckson:
Adds functionality for ensuring all resources have the provider declared in required_providers

### DIFF
--- a/docs/rules/terraform_resouces_have_required_providers.md
+++ b/docs/rules/terraform_resouces_have_required_providers.md
@@ -1,0 +1,55 @@
+# terraform_resouces_have_required_providers
+
+Require that all resources have version constraints declared through `required_providers`.
+
+## Configuration
+
+```hcl
+rule "terraform_resources_have_required_providers" {
+  enabled = true
+}
+```
+
+## Examples
+
+```hcl
+terraform {
+ required_providers {
+ }
+}
+
+resource "template_test" "example" {
+}
+
+```
+
+```
+$ tflint
+1 issue(s) found:
+
+Warning: `Missing version constraint for provider "template" in "required_providers"
+
+  on main.tf line 6:
+   1: provider "template" {}
+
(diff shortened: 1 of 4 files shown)


bendrucker:
Hi, the `terraform_required_providers` rule already exists:

https://github.com/terraform-linters/tflint/blob/master/docs/rules/terraform_required_providers.md
https://github.com/terraform-linters/tflint/blob/master/rules/terraformrules/terraform_required_providers.go

It may be instructive to compare your PR to the actual rule implementation, particularly with respect to how provider `alias` values relate to `required_providers` (they don't).

Happy to help with future PRs if you have other rule ideas.

mcuckson:
Thanks for the feedback, much appreciated. 

I think that the rule which you mentioned is for a different scenario. In my PR the intention is to generate the issue when a provider has been used in a resource block but is not pinned to a version in `required_providers` . As an example, when given the snippet below, `terraform_required_providers` will not raise a warning unless the provider line is uncommented.

```hcl
terraform {
  required_providers {
  }
}

//provider "azuread" {}

resource "azuread_application" "example" {
  name = "ExampleApp"
}
```

With regards to alias, as you suggested, I have misunderstood the Terraform documentation and have updated the PR accordingly.

Please let me know if you have any more feedback. I have a need for this rule on my current project so can maintain it in the fork but it would be great to get it into upstream if you think it would be useful.

bendrucker:
Ok, in that case that should be a fix to the existing rule. The intent is definitely to enforce required versions for all providers in use, not just those those with explicit blocks/configuration. Would've thought Terraform smooths over the implicit provider configuration, but I guess not.

Start with adding failing test cases and work from there. I would not expect any existing code to be removed/changed, just extra loops to build up the map of provider names -> declared range. Keep in mind that any logic operating on resources must be repeated for data sources.

bendrucker:
Since this was a very quick fix I took care of it in #889.

PR #910, opened 2020-09-11
Adding a rule to check aws_s3_bucket names match a common regex

sam-burrell:
We would like to follow convention of bucket names in our distributed terraform system.
This seems to be a good quick way to do that.

I've tested it in our local multi state file terraform system. 
Happy to do anything else to get this released


bendrucker (commented):
Few general comments:

* The rule name can be `aws_s3_bucket_name` and take multiple options for enforcing bucket naming patterns, rather than describing exactly how the rule is implemented
* It seems like `prefix` should be an option for the common case and then `regex` can be available for anything else. If both are defined both should be validated.
* In examples, keep things generic. Avoid using `prod` or anything hinting at your own or even expected usage. Foo/bar is best for generic rules.

Will come back through and do an actual code review, but wanted to give you an opportunity to address these high level comments first.

sam-burrell:
Thanks good feedback makes sense.

Seeing as prefix is easily done with a regex for now. 
Could we try this first and see how people use it as is?

I will changed the docs to be more generic.

sam-burrell:
now renamed to aws_s3_bucket_name

thoughts on the option name for the regex?

bendrucker:
Thanks! `regex` is good as the attribute name.

> Could we try this first and see how people use it as is?

The feedback cycle isn't that tight for usability improvements. If a user wants more flexibility, they'll open issues or PRs. If a feature is too generic or low level, users will suffer through. "Seeing how people use" features isn't really a thing we can do. 

The use case you have in mind seems to be prefixes, and that would seem to be the common case by a significant margin. So I'd like to start with the expected use case being simpler rather than pushing the burden onto future maintainers.

sam-burrell:
I have now added a prefix option and made both optional. 
We might want to consider order of preferences of rules.

bendrucker:
Awesome, thanks! Will get to this for a thorough review of the code this evening. Looks good at a glance so I should be able to just take care of anything I do find. Re preference, I mentioned above:

> If both are defined both should be validated.

I think that should work, you could define a prefix as well as a regex adding additional specificity (e.g. restricting characters).

bendrucker:
Sorry, took an extra day, but this is merged in eebaf5a. Wasn't able to push my commits to this PR b/c it came from master, but you can see the difference (plus some dependency noise in between) here:

https://github.com/signal-ai/tflint/compare/77fb676..terraform-linters:eebaf5a

Aside from some all around cleanup, the rule was panicking when an invalid regex was passed, b/c of `MustCompile`. That is now returned as an error before any resources are evaluated and compiled once per configuration.

PR #928, opened 2020-09-23
docker: add curl + tfenv

### LINKED ISSUE #926: Feature request: Docker image to run against multiple directories
Since `tflint` doesn't support a recursive mode, for use in CI, it would useful if there were a variant of the Docker container that contains some way to check multiple files (e.g., a shell, such that something like `find . -type f -name "*.tf" -exec dirname {} \; | uniq | xargs -n1  tflint` can be run). I can see why it might not be ideal to build that into the default Dockefile, but some other lint tools (e.g., shellcheck) provide both options to support situations like this.

Or even an alternate entrypoint that supports checking a recusive filesystem hierarchy natively.

bendrucker:
This adds some utilities to the Dockerfile per discussion in #926 

* `curl`: for downloading plugin binaries
* `tfenv`: for easily installing Terraform, including installing/switching among multiple versions in a monorepo. This requires bash.

Closes #926

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,10 @@ FROM golang:1.15-alpine3.12 as builder
 
 RUN apk --no-cache add git make gcc musl-dev zip
 
+WORKDIR /tfenv
+ARG TFENV_VERSION=v2.0.0
+RUN git clone https://github.com/tfutils/tfenv.git . && git checkout --quiet ${TFENV_VERSION}
+
 WORKDIR /tflint
 ADD . /tflint
 RUN make build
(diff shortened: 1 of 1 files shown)


wata727:
Umm... I think it's not bad to pre-install tfenv and curl.

However, initially, I intended to provide this Docker image as a way to check the behavior without installing TFLint locally, rather than an image that can be used in a production workflow. Perhaps even if this image contains the utilities, it may not be enough for the workflow.

On the other hand, I'm interested in providing another image that can be used for such a real workflow. In my opinion, I would like to manage such an image in another repository and contains only tflint binary in this image. Any thoughts?

bendrucker:
Sure, that sounds good to me. I do agree with the "is this enough?" sentiment you expressed. 

And also with keeping the Dockerfile suitable for local use. Most users could just `terraform init` on the host and then mount the working directory into the container.

Once you start adding tools you open up choice, versioning, and other problems. I don't think this is a big enough need to put any more work in for now, but this discussion could serve as a reference if someone wanted to work on a dedicated tools repo/image, perhaps under the terraform-linters org.

PR #963, opened 2020-10-27
escape language server paths

filiptepper:
When testing the Language Server feature in Windows I have discovered, that VSCode is sending file path as encoded URLs.

This results in paths like `C%3A/` instead of expected `C:/`, hence Language Server is yielding errors during `chdir`.


filiptepper:
One of the checks is failing with `GitHub Actions has encountered an internal error when running your job.`.

wata727:
👍

PR #995, opened 2020-12-07
fix: exclude tfvars files from pre-commit hook

andyjp94:
tflint does not support tfvars files which causes
pre-commit to fail when there are repository changes
to tfvars files

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,4 +3,5 @@
   description: TFLint is a Terraform linter focused on possible errors, best practices, and so on.
   language: golang
   entry: tflint
+  exclude: .*.tfvars$
   types: [terraform]


bendrucker:
TFLint has a `--var-file` flag. Will need a more complete reproduction/explanation of your problem to consider merging this.

andyjp94:
Hi Ben, I think this is a pretty straightforward scenario so I think I'm likely confused as to how this hook is intended to be used.

Lets say there is a scenario where I edit a tfvars file called `x.tfvars`, stage it and try to commit it. The pre-commit hook will run against that file and it will fail as running `tflint x.tfvars` also fails.

The error I get in such a scenario is:
```

bendrucker:
Hmm, ok, was not familiar with these hooks at all, thanks for filling in the background. So it looks like the following means "when any Terraform file changes, pass the file path to `tflint`":

https://github.com/terraform-linters/tflint/blob/master/.pre-commit-hooks.yaml#L5-L6

Is that right?

If so I'm inclined to just remove that config, as that behavior is incorrect. The scope of Terraform evaluation is a directory and at a file level the only checking that's safe to do is `fmt`. 

It seems like this is actively developed and perhaps the config in this repo can just be deleted:

https://github.com/antonbabenko/pre-commit-terraform

bendrucker:
To wrap this up, this seems to be an issue with https://github.com/gruntwork-io/pre-commit/blob/master/hooks/tflint.sh. If there's something we can change in the config here happy to take a PR, otherwise it doesn't seem like this is behavior we control.

PR #1007, opened 2020-12-19
rule: terraform_unused_required_providers

bendrucker:
As a complement to the "required providers" rule, this adds a rule that detects unused required providers. This ends up being slightly more complicated in the wild, since technically it's valid to put all your provider requirements in the root module, but it means `terraform init` may not work in child modules.

In theory this rule could also walk down the module tree looking for provider use, but I haven't implemented that here. This rule is disabled by default and users who specify required providers directly in each module can just turn it on.

### DIFF
--- a/rules/terraformrules/terraform_unused_required_providers.go
+++ b/rules/terraformrules/terraform_unused_required_providers.go
@@ -4,10 +4,11 @@ import (
 	"fmt"
 	"log"
 
+	"github.com/hashicorp/terraform/configs"
 	"github.com/terraform-linters/tflint/tflint"
 )
 
-// TerraformUnusedRequiredProvidersRule checks whether Terraform sets version constraints for all configured providers
+// TerraformUnusedRequiredProvidersRule checks whether required providers are used in the module
 type TerraformUnusedRequiredProvidersRule struct{}
 
 // NewTerraformUnusedRequiredProvidersRule returns new rule with default attributes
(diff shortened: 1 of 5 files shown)


rules/terraformrules/terraform_unused_required_providers.go
```diff
@@ -0,0 +1,79 @@
...
+	for _, required := range runner.TFConfig.Module.ProviderRequirements.RequiredProviders {
+		r.checkProvider(runner, required)
+	}
+
+	return nil
+}
+
+func (r *TerraformUnusedRequiredProvidersRule) checkProvider(runner *tflint.Runner, required *configs.RequiredProvider) {
+	for _, resource := range runner.TFConfig.Module.ManagedResources {
+		if required.Name == resource.Provider.Type {
```
wata727, line 57: If you don't use a preferred local name, this condition may not work as expected.
https://www.terraform.io/docs/configuration/provider-requirements.html#local-names

Example:

```hcl
terraform {
  required_providers {
    mycloud = {
      source = "hashicorp/aws"
      version = "3.20.0"
    }
  }
}

  provider = mycloud

  ami = "ami-12345678"
  instance_type = "t2.micro"
}
```

bendrucker: Good catch, handled in 3a3b486

wata727 (approved):
👍

PR #1036, opened 2021-01-16
plugin: Fix segfault when accessing Config

wata727:
Fixes https://github.com/terraform-linters/tflint-plugin-sdk/issues/95


wata727:
Merged #1040 instead.

PR #1040, opened 2021-01-21
Fix panic on empty backend in Config()

syndicut:
Config() method causes panic when used in plugins if backend config is non-existent

Related:
terraform-linters/tflint-plugin-sdk#95


syndicut:
sorry, didn't see #1036 already addressing it. My fix includes test, though =) And I think returning empty structure is more "in line" with other methods.

syndicut:
> sorry, didn't see #1036 already addressing it. My fix includes test, though =) And I think returning empty structure is more "in line" with other methods.

changed my mind about nil, empty structure causes errors

wata727 (approved):
👍

PR #1041, opened 2021-01-21
Fix gob encoder error on unknown value

syndicut:
When using Config() method with `--module` option you can get cty.unknownType inside module variables without default values and following error:
```
13:39:46 server.go:418: rpc: gob error encoding body: error encoding cty.Value: gob: type not registered for interface: cty.unknownType
```
This fix forces cty.Value{} in case of unknown value

### DIFF
--- a/plugin/encode.go
+++ b/plugin/encode.go
@@ -8,6 +8,7 @@ import (
 	"github.com/terraform-linters/tflint-plugin-sdk/terraform/experiments"
 	tfplugin "github.com/terraform-linters/tflint-plugin-sdk/tflint/client"
 	"github.com/zclconf/go-cty/cty/msgpack"
+	"github.com/zclconf/go-cty/cty/json"
 	"github.com/terraform-linters/tflint/tflint"
 )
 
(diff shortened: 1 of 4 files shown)


wata727:
Umm, I understand this bug, but this fix makes it impossible for the plugin to handle variables as unknown.

Maybe gob is not a good way to transfer `cty.Value`. go-cty seems to support encoding/decoding with msgpack, so I think this might be a good idea, but I haven't started investigating anything yet. If you are interested, I would like to support you as much as possible.
https://pkg.go.dev/github.com/zclconf/go-cty@v1.7.1/cty/msgpack

syndicut:
> Umm, I understand this bug, but this fix makes it impossible for plugins to handle variables as unknown.
> 
> Maybe gob is not a good way to transfer `cty.Value`. go-cty seems to support encoding/decoding with msgpack, so I think this might be a good idea, but I haven't started investigating anything yet. If you are interested, I would like to support you as much as possible.
> https://pkg.go.dev/github.com/zclconf/go-cty@v1.7.1/cty/msgpack

Ok, good point, will look at it

syndicut:
> Umm, I understand this bug, but this fix makes it impossible for plugins to handle variables as unknown.
> 
> Maybe gob is not a good way to transfer `cty.Value`. go-cty seems to support encoding/decoding with msgpack, so I think this might be a good idea, but I haven't started investigating anything yet. If you are interested, I would like to support you as much as possible.
> https://pkg.go.dev/github.com/zclconf/go-cty@v1.7.1/cty/msgpack

Think those changes and terraform-linters/tflint-plugin-sdk#96 should do the trick

wata727:
Merged https://github.com/terraform-linters/tflint-plugin-sdk/pull/96. Could you update the sdk version in this PR?

syndicut:
> Merged [terraform-linters/tflint-plugin-sdk#96](https://github.com/terraform-linters/tflint-plugin-sdk/pull/96). Could you update the sdk version in this PR?

Done!

plugin/encode.go
```diff
@@ -360,11 +362,14 @@ func (s *Server) encodeVariable(variable *tfconfigs.Variable) *tfplugin.Variable
 		validations[i] = s.encodeVariableValidation(v)
 	}
 
+	defaultVal, _ := msgpack.Marshal(variable.Default, variable.Default.Type())
+	// We need to use json because cty.DynamicPseudoType cannot be encoded by gob
+	typeVal, _ := json.MarshalType(variable.Type)
```
wata727, line 367: We may need to extend `encodeVariable` to properly return this error to the caller.

wata727 (approved):
👍

PR #1042, opened 2021-01-22
Added option to enable plugins from the cli

janritter:
This adds a CLI option to enable plugins.

https://github.com/terraform-linters/tflint/pull/1039#issuecomment-764896952

### DIFF
--- a/cmd/option.go
+++ b/cmd/option.go
@@ -18,6 +18,7 @@ type Options struct {
 	EnableRules    []string `long:"enable-rule" description:"Enable rules from the command line" value-name:"RULE_NAME"`
 	DisableRules   []string `long:"disable-rule" description:"Disable rules from the command line" value-name:"RULE_NAME"`
 	Only           []string `long:"only" description:"Enable only this rule, disabling all other defaults. Can be specified multiple times" value-name:"RULE_NAME"`
+	EnablePlugins  []string `long:"enable-plugin" description:"Enable plugins from the command line" value-name:"PLUGIN_NAME"`
 	Varfiles       []string `long:"var-file" description:"Terraform variable file name" value-name:"FILE"`
 	Variables      []string `long:"var" description:"Set a Terraform variable" value-name:"'foo=bar'"`
 	Module         bool     `long:"module" description:"Inspect modules"`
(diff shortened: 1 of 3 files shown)


cmd/option.go
```diff
@@ -85,6 +87,15 @@ func (opts *Options) toConfig() *tflint.Config {
 		log.Printf("[WARN] Usage of --only will ignore other rules provided by --enable-rule or --disable-rule")
 	}
 
+	plugins := map[string]*tflint.PluginConfig{}
+	for _, plugin := range opts.EnablePlugins {
+		plugins[plugin] = &tflint.PluginConfig{
+			Name:    plugin,
+			Enabled: true,
+			Body:    nil,
```
wata727, line 95: I'm a little worried about the design of this feature. I find it convenient to be able to enable plugins with the CLI flag, but in that case, the `Body` will always be nil, which will result in incomprehensible errors for plugins that require configuration.

janritter: One option would be to change the name of the cli option to something like `--enable-plugin-without-config`, then it would be clear that the config is always nil and plugin errors for missing config would be easy to understand. The second option would be to pass the config with the cli option, but I think this is not easy to use or implement.

wata727: > One option would be to change the name of the cli option to something like --enable-plugin-without-config

Umm... I would like to avoid the descriptive long option name...

> The second option would be to pass the config with the cli option, but I think this is not easy to use or implement.

Agreed.

It's not a problem now, so we have the option to accept this change and consider the design once it's actually a problem. Now I am willing to accept this change.

wata727: However, since the plugin configuration will be overridden when using the CLI flag, it is necessary to take the same ways as the rules.

https://github.com/terraform-linters/tflint/blob/382e3cc7a5f3d7320c816525caa0769f63bd9943/tflint/config.go#L360-L365

janritter: I'm not fully sure what you mean. 
I just manually tested my change with  a `.tflint.hcl` config:

```hcl
plugin "codestyle" {
    enabled = false
}

plugin "aws" {
    enabled = true
}
```
The codestyle plugin can be enabled via cli option, this will override the `enabled = false`. Setting the cli option has no effect on the aws plugin block.

wata727: For instance, if you enable the AWS plugin from the CLI with the following configuration, the value of the `deep_check` option will be overridden.

```hcl
plugin "aws" {
  enabled = false
  deep_check = true
}
```

tflint/config.go
```diff
@@ -377,6 +377,10 @@ func mergePluginMap(a, b map[string]*PluginConfig) map[string]*PluginConfig {
 	}
 	for k, v := range b {
 		ret[k] = v
+		if prevConfig, exists := ret[k]; exists && v.Body == nil {
+			// Use the plugin config from the config file when the plugin is enabled via CLI
+			ret[k].Body = prevConfig.Body
```
wata727, line 382: It seems that `file` needs to be kept as well. When transferring the configuration to a plugin, it will be ignored if the file is nil.

wata727 (approved):
👍

PR #1043, opened 2021-01-23
Call ApplyConfig before ValidateRules

richardTowers:
I'm working on a plugin which generates a list of rules dynamically,
based on its config:

https://github.com/richardTowers/tflint-ruleset-workspaces/

At the moment, it's not possible to refer to config in RuleNames because
that gets called before ApplyConfig. As a workaround, I've made my
plugin return the empty list for RuleNames:

https://github.com/richardTowers/tflint-ruleset-workspaces/blob/master/ruleset.go#L28

Naïvely, it looks like making the ApplyConfig a bit earlier would fix
this.

The tests pass, but I'm not sure if this will have unintended
consequences. Careful review would be much appreciated!

### DIFF
--- a/cmd/inspect.go
+++ b/cmd/inspect.go
@@ -63,7 +63,12 @@ func (cli *CLI) inspect(opts Options, dir string, filterFiles []string) int {
 	defer plugin.Clean()
 
 	rulesets := []tflint.RuleSet{&rules.RuleSet{}}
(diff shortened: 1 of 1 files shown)


wata727 (approved):
Hmm, originally, I didn't expect to generate rules dynamically, but I don't think this change will cause any major problems.

Issue #880, opened 2020-08-19
Error installing latest version


rspowell:

bendrucker:
We'll need more information, including the script you're running and which OS. Happen to reopen with details.

rspowell:
Thanks Ben. I’m using the provided install_linux.sh script and I’m running RHEL 7.7 with selinux enabled and in permissive mode.

bendrucker:
That would seem to indicate that the ZIP file isn't downloaded successfully, and `unzip` is being called on something that is not a ZIP. Can you inspect the file and see if it's empty or something? Perhaps something went wrong with the `curl` command that was non-error (e.g. a 4xx response) but resulted in a bad file.

The full output of running the install script would also be helpful

rspowell:
./install_linux.sh
os=linux_amd64


====================================================
Looking up the latest version ...
Downloading TFLint
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     9  100     9    0     0     26      0 --:--:-- --:--:-- --:--:--    26
Download was successfully


====================================================
Unpacking /tmp/tflint.zip ...
Archive:  /tmp/tflint.zip
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of /tmp/tflint.zip or
        /tmp/tflint.zip.zip, and cannot find /tmp/tflint.zip.ZIP, period.
Installing /tmp/tflint to /usr/local/bin...
install: cannot stat '/tmp/tflint': No such file or directory
Failed to install tflint

[cloud-user@devestation tmp]$ sudo sum tflint.zip
31950     1

bendrucker:
Quite apparent from that output that the right file is not downloaded, assuming the curl output showing a very small number of bytes is accurate. Add `set -x` to the script to see each how command is evaluated as it runs.

PR #1045, opened 2021-01-26
Versions are stored with a preceeding "v"

cdw:
I was getting the same error as #880 when installing as part of a docker build. Taking a look, it appears that a missing 'v' was causing a 404.


cdw:
Nope, should have been solved by adding a 'v' to my TFLINT_VERSION env variable.

PR #1047, opened 2021-01-27
add module/build caching to e2e + gen

bendrucker:
Brings the same changes in #1046 to the E2E and generated code workflows.


bendrucker:
[Before](https://github.com/terraform-linters/tflint/runs/1773406258?check_suite_focus=true): 1m 10s
[After](https://github.com/terraform-linters/tflint/pull/1047/checks?check_run_id=1779116944): 22s

PR #1060, opened 2021-02-15
docker: remove unused build tools

pujan14:
Signed-off-by: Pujan Shah <pujan14@gmail.com>

Removed unused build tools from dockerfile and added default argument to image with CMD

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
 FROM golang:1.15-alpine3.12 as builder
 
-RUN apk --no-cache add git make gcc musl-dev zip
+RUN apk add --no-cache make
 
 WORKDIR /tflint
-ADD . /tflint
+COPY . /tflint
 RUN make build
 
 FROM alpine:3.12 as prod


Dockerfile
```diff
@@ -12,7 +12,14 @@ LABEL maintainer=terraform-linters
 
 RUN apk add --no-cache ca-certificates
 
-COPY --from=builder /tflint/dist/tflint /usr/local/bin
+# copy tflint from builder
```
bendrucker, line 15: Remove these comments, they are unnecessarily re-stating things that can be readily inferred from the code. Everything happening here is basic and well-documented upstream. You can assume basic understanding of Dockerfiles here and not try to explain how they work.

Dockerfile
```diff
@@ -12,7 +12,14 @@ LABEL maintainer=terraform-linters
...
+COPY --from=builder /tflint/dist/tflint /usr/local/bin/tflint
 
-ENTRYPOINT ["tflint"]
+# set working directory
 WORKDIR /data
+
+# set the default entrypoint -- when this container is run, use this command
+ENTRYPOINT [ "tflint" ]
```
bendrucker, line 22: Keep the formatting consistent with Docker's docs, meaning no array padding:

https://docs.docker.com/engine/reference/builder/

Dockerfile
```diff
@@ -12,7 +12,14 @@ LABEL maintainer=terraform-linters
...
+
+# as we specified an entrypoint, this is appended as an argument (i.e., `tflint --help`)
+CMD [ "--help" ]
```
bendrucker, line 25: Is this really the best default command, as opposed to invoking TFLint with no arguments, which is implicitly the same as `tflint .`?

bendrucker (changes requested):
I don't agree with the user-facing change proposed here, but if those APK packages in the builder phase are unnecessary, happy to accept that change on its own.

Dockerfile
pujan14: reverting back to original behaviour

PR #1070, opened 2021-02-24
pass --loglevel to plugins as TFLINT_LOG

bendrucker:
Per https://github.com/terraform-linters/tflint-ruleset-aws/issues/69#issuecomment-784701244, when users set a log level via `--loglevel`, plugin processes do not reflect this input. This passes it via `TFLINT_LOG`.

Longer term, removing `--loglevel` altogether would prevent this from being an issue. Terraform does not have a log level flag, perhaps in part in recognition that environment variables fit better with external plugin processes.

### DIFF
--- a/cmd/cli.go
+++ b/cmd/cli.go
@@ -63,6 +63,8 @@ func (cli *CLI) Run(args []string) int {
 	level := os.Getenv("TFLINT_LOG")
 	if opts.LogLevel != "" {
 		level = opts.LogLevel
+		// Pass log level to plugin processes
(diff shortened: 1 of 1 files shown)


wata727 (approved):
Good catch!

> Longer term, removing --loglevel altogether would prevent this from being an issue. Terraform does not have a log level flag, perhaps in part in recognition that environment variables fit better with external plugin processes.

Agreed. Initially, I didn't understand this concern, so it was implemented in https://github.com/terraform-linters/tflint/pull/844, but for this reason, I don't think there's any reason to continue to support it.

PR #1123, opened 2021-05-19
feat(rules): detect branch names "main", "develop"

jabielecki:
It's true that the repo's default branch is mostly named "master", but
two notable additional names are "main" and "develop". All three are
equally unwanted when used in a ref=.

In most git implementations (including GitHub) you are free to set any
branch name as default, but these 3 names cover probably 99% of cases
as far as I've seen personally in the wild.

This change is backward-compatible with the TF modules based on these
two branches.

How do I know I should expect URL-like behavior such as
`example.com?ref=a&depth=5` here? Peeking at Terraform source code, it is
using go-getter library. Checking its docs confirms the format:

https://pkg.go.dev/github.com/hashicorp/go-getter

### DIFF
--- a/docs/rules/terraform_module_pinned_source.md
+++ b/docs/rules/terraform_module_pinned_source.md
@@ -8,11 +8,13 @@ Name | Default | Value
 --- | --- | ---
 enabled | true | Boolean
 style | `flexible` | `flexible`, `semver`
+default_ref_regexp | `master` | A regular expression
 
 ```hcl
 rule "terraform_module_pinned_source" {
   enabled = true
   style = "flexible"
+  default_ref_regexp = "master|main|develop"
 }
 ```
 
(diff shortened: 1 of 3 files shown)


bendrucker (commented):
Nice! Seems like these branch names should be moved to config for this rule rather than hardcoded in a regex. That way users can pick their own forbidden branch names if desired. The three you've picked are good defaults.

jabielecki:
Thanks @bendrucker! 😄 Please clarify whether what you are saying is a nitpick or a blocking comment.

Could you link or point me to a place in the codebase where any other rule is using a config setting?

bendrucker:
All review comments are blocking. Our goal is to merge something maintainable and not just minimally address your needs. 

The job of the Terraform rules is to implement official, documented recommendations. Avoiding specific branches is not an official recommendation. 

https://www.terraform.io/docs/language/modules/sources.html

TFLint has already overstepped a bit by having opinions here and your desire to adjust those opinions reflects the fact that these are user space concerns that should be configurable, even if TFLint provides agreeable defaults. 

Look at the naming convention (casing) rule for an example of config parsing and handling.

jabielecki:
You probably mean [this part](https://github.com/terraform-linters/tflint/blob/a50db0dbaf8f38a1d4ee3a2ef1ad9cff15dccd3e/rules/terraformrules/terraform_naming_convention.go#L83-L92), thank you, it's doesn't seem immediately doable for me. I'll be looking into that as time allows.

bendrucker:
Yep! If you're not interested that's totally ok too.

jabielecki:
I've added a config option. The change seems backwards compatible now.

@bendrucker Could you take a look please?

bendrucker:
Awesome, thank you! I may tweak this a bit but I can handle those revisions directly before merging rather than turning them back to you.

PR #1132, opened 2021-06-11
Expose Files() runner method to Server

jonathansp:
**Abstract**

This PR exposes `Files()` method to [tflint-plugin-sdk](https://github.com/terraform-linters/tflint-plugin-sdk/issues/118). Current implementation already has the `File(filename string)` method exposed.

**Rationale**

Linting is the process of spotting potential errors. These errors can cause problems not only during the code execution but also to the business logic. Clean and organised files are often mandatory to maintainability. Some rules, however, make sense to one specific project or company. [tflint-plugin-sdk](https://github.com/terraform-linters/tflint-plugin-sdk) helps to extend _**tflint**_ in a way where each company can implement their own set of rules. 

**Use cases**

This PR exposes the `Files()` method in order to allow the developer to create more low-level/business-specific rules. A few examples are:
 - Checks if the body of a provider is empty, where the company does not allow it.
 - Keeps blocks in alphabetical order, when needed, for cleansing aspects.
 - Helps to stick to business-defined naming conventions.
 - Helps to block cross-environment actions, dev to prod and vice-versa.

Arguably, some of these rules can be achieved in other ways, but once we can extend _tflinf_ in plugins, I think its utilisation is correct in these cases. This PR depends on https://github.com/terraform-linters/tflint-plugin-sdk/pull/122/files already merged.

**Closes:** 
- https://github.com/terraform-linters/tflint-plugin-sdk/issues/118
 - https://github.com/terraform-linters/tflint-plugin-sdk/issues/119

Any improvement that you guys think to this PR, I will be happy to change.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/sourcegraph/go-lsp v0.0.0-20181119182933-0c7d621186c1
 	github.com/sourcegraph/jsonrpc2 v0.0.0-20190106185902-35a74f039c6a
 	github.com/spf13/afero v1.2.2 // matches version used by terraform
-	github.com/terraform-linters/tflint-plugin-sdk v0.8.2
+	github.com/terraform-linters/tflint-plugin-sdk v0.8.3-0.20210614125323-8364139f3745
 	github.com/terraform-linters/tflint-ruleset-aws v0.4.1
 	github.com/zclconf/go-cty v1.8.3
 	golang.org/x/crypto v0.0.0-20210513164829-c07d793c2f9a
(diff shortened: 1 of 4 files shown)


bendrucker:
Have a look at the failing checks. Also, a PR description indicating how you plan to use this would be helpful.

jonathansp:
cc @wata727

jonathansp:
> Have a look at the failing checks. Also, a PR description indicating how you plan to use this would be helpful.

PR description updated.

wata727:
👍

jonathansp:
Thanks @wata727 

Now we can write rules like:

```go


func (rule *MyProjectOnlyAllowedFilesRule) Check(runner tflint.Runner) error {
	files, _ := runner.Files()
	allowedFiles := map[string]bool{"providers.tf": true, "main.tf": true}

	for name := range files {
		_, filename := path.Split(name)

		if _, exists := allowedFiles[filename]; !exists {
			message := fmt.Sprintf("File %s is not allowed here.", filename)

			return runner.EmitIssue(rule, message, hcl.Range{Start: hcl.InitialPos})
		}
	}
	return nil
}
```

will output:

```bash

Error: File backend.tf is not allowed here. (my_project_only_allowed_files_rule)

  on  line 1:
   (source code not available)

```

PR #1160, opened 2021-07-11
plugin: Deprecate bundled plugins

wata727:
For historical reasons, the "aws" plugin was available even if you didn't install the plugin, unlike other plugins. It is called "bundled plugins".

This PR will display a deprecation warning to end support for bundled plugins. The warning will be displayed if the following conditions are met:

- tflint-ruleset-aws was not installed
- Terraform configurations have declarations of aws resources (such as `aws_instance`)

This warning is output to stderr at runtime, etc. in the following format:

```console
$ tflint
WARNING: The plugin `aws` is not explicitly enabled. The bundled plugin will be enabled instead, but it is deprecated and will be removed in a future version. Please see https://github.com/terraform-linters/tflint/pull/1159 for details.
1 issue(s) found:

Error: "invalid" is an invalid value as instance_type (aws_instance_invalid_type)

  on template.tf line 21:
  21:   instance_type = "invalid"

```

You must explicitly install tflint-ruleset-aws to get rid of this warning. Add the following declaration to `.tflint.hcl` and install it with `tflint --init`.

```hcl
plugin "aws" {
    enabled = true
    version = "0.5.0"
    source  = "github.com/terraform-linters/tflint-ruleset-aws"
}
```

```console
$ tflint --init
Installing `aws` plugin...
Installed `aws` (source: github.com/terraform-linters/tflint-ruleset-aws, version: 0.5.0)
```

We will proceed step by step to end support for the bundled plugin. Please let us know if you have any operational issues at the end of support. Thank you.

PR #1164, opened 2021-07-15
Update README about Docker images

wata727:
This PR adds a link to the newly created [tflint-bundle](https://github.com/terraform-linters/tflint-bundle) image.

This image is one of the alternatives to bundled plugins and is mainly intended for use in super-linter.

See also https://github.com/terraform-linters/tflint/pull/1160

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,13 +21,13 @@ TFLint is a framework and each feature is provided by plugins, the key features 
 Bash script (Linux):
 
 ```console
-$ curl https://raw.githubusercontent.com/terraform-linters/tflint/master/install_linux.sh | bash
+curl https://raw.githubusercontent.com/terraform-linters/tflint/master/install_linux.sh | bash
 ```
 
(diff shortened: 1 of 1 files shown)


bcdady:
I'm just starting out converting my workflow to use docker images with `tflint` (instead of installed from homebrew), as preparation for including in CI (but not GitHub Actions).

When I run the tflint-bundle image, I get this error message, even though my local .tfling.hcl file includes `plugin "aws" { enabled = true }`

Do the comments in [PR 1160](https://github.com/terraform-linters/tflint/pull/1164) mean that the `tflint-bundle` image will also be deprecated soon? If so, I think that should be clarified in the README.

```txt
WARNING: The plugin `aws` is not explicitly enabled.
The bundled plugin will be enabled instead, but it is deprecated and will be removed in a future version.
Please see https://github.com/terraform-linters/tflint/pull/1160 for details.
```

Issue #1168, opened 2021-07-22
Error fork/exec when directory exists with plugin name


josh-barker-coles:
This is re-raising the issue from https://github.com/terraform-linters/tflint-ruleset-azurerm/issues/101

Looking at this function, it looks like there is no check to determine if the path is a directory or a file.
Hence, there is the error about fork/exec.

https://github.com/terraform-linters/tflint/blob/master/plugin/discovery.go#L119

PR #1169, opened 2021-07-22
fix: return error if plugin is not a file

josh-barker-coles:
Addresses #1168

### DIFF
--- a/plugin/discovery.go
+++ b/plugin/discovery.go
@@ -117,15 +117,24 @@ func getPluginDir() (string, error) {
 // Only in the case of Windows, the pattern with the `.exe` is also considered,
 // and if it exists, the extension is added to the argument.
 func findPluginPath(path string) (string, error) {
-	_, err := os.Stat(path)
-	if os.IsNotExist(err) && runtime.GOOS != "windows" {
-		return "", os.ErrNotExist
-	} else if !os.IsNotExist(err) {
-		return path, nil
+	if runtime.GOOS != "windows" {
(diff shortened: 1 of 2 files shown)


plugin/discovery.go
```diff
@@ -117,15 +117,17 @@ func getPluginDir() (string, error) {
 // Only in the case of Windows, the pattern with the `.exe` is also considered,
 // and if it exists, the extension is added to the argument.
 func findPluginPath(path string) (string, error) {
-	_, err := os.Stat(path)
+	info, err := os.Stat(path)
 	if os.IsNotExist(err) && runtime.GOOS != "windows" {
 		return "", os.ErrNotExist
-	} else if !os.IsNotExist(err) {
+	} else if !os.IsNotExist(err) && !info.IsDir() {
```
wata727, line 123: If you add a condition here, it will cause unintended behavior in the following cases:

- `runtime.GOOS == "linux"`
- path exists, but it's a directory

In this case, `findPluginPath` searches for the `.exe` file, which should only be handled on Windows.
In order to organize the implementation, it may be better to first branch the process according to the runtime OS.

wata727 (commented):
Nice!

plugin/discovery.go
josh-barker-coles: Hi @wata727, I've refactored it into a separate function to tidy up the intent and understanding.

wata727:
Thank you!

josh-barker-coles:
No worries @wata727  :-)

Issue #1053, opened 2021-01-31
feature request: module source must point to git repository


m00lecule:
Hello Guys!

After a while of using terraform I spotted a tendency to push/commit changes, that for testing purposes points to my local path instead of git repo. I believie this rule is missing and might be useful in `TFLint`, forcing that all sources must point to remote repository.

Following code would be evaluated as invalid:

```
module "my-module" {
  source = "/home/user/documents/directory"
}
```
Example of valid module:
```
module "my-module" {
  source = "git::ssh://git@github.com/user/repo.git"
}
```

Also property `clone_protocol` that check if source repo is cloned using `https` or `ssh` might be a valid feature

m00lecule:
@bendrucker  - I agree, verification if source path is not absolute is enough. There might be cases, when we would like to keep make submodules, to gather related resources. If this feature (absolute path verification) is implemented please close this issue.

abuxton:
it would be useful instead of pinning to a just git based repo if you could variable-ize the URL? 
working with TF you run into situations where you want to ensure the module source 

-   on prem git/tfe or if it is sourcing externally
-  confined to TFC PMRs `https://app.terraform.io/orgname`
- using tfe and confined to generic hostname `localterraform.com`  [terraform](https://www.terraform.io/cloud-docs/registry/using#generic-hostname-terraform-enterprise) 

i wanted to suggest this before I attempted to create a new rule just for validating the uri based on a configurable variable.

PR #1182, opened 2021-08-13
add rule: terraform_module_version

### LINKED ISSUE #1176: rule: require `module` blocks to specify `version`
Hi Community,

In the current version, tflint does not have rule check is module version pinned.
Alternatively, you can use `terraform_module_pinned_source` however, that does not work together with modules from Terraform registry.

Thanks for considering this feature.

Best,
Andor

bendrucker:
Adds a `terraform_module_version` rule that Terraform modules from the registry should set a source. Similar to `terraform_module_pinned_source`, this is enabled by default.

Users can optionally specify `exact = true` to require that `version` be an exact 3 part version.

This brings back some Terraform files in `addrs` as well as dependent packages. They were necessary to determine whether a `source` resolves to a registry:

https://github.com/hashicorp/terraform/blob/120629848761f5434c75b9876a030f8145abf043/internal/addrs/module_source.go#L63-L73

This could also be useful for #1053 which would want to check whether a `source` is a local one before applying some rules to it.

Fixes #1176


bendrucker:
Whoops, tagged the wrong # above

terraform/copy/copy_dir.go
```diff
@@ -0,0 +1,146 @@
+package copy
```
wata727, line 1: Do you need these codes? I have confirmed that the `CopyDir` is referenced from the `getmodules` package, but it seems that `getWithGoGetter` and `FetchPackage` that use it are not used from there either.

Whenever possible, I recommend that you remove unused code. This should reduce the code under the `terraform` directory that must be added with this change.

rules/terraformrules/terraform_module_version.go
```diff
@@ -0,0 +1,118 @@
...
+
+	if len(module.Version.Required) > 1 {
+		runner.EmitIssue(
+			r,
+			fmt.Sprintf("module %q should specify an exact version, but multiple constraints were found", module.Name),
+			module.Version.DeclRange,
+		)
+	}
+
+	if !exactVersionRegexp.MatchString(module.Version.Required[0].String()) {
```
wata727, line 109: If `exact = true` and no version is specified, panic will occur here.
If the version is nil or if multiple constraints are found, it is recommended to return early.

rules/terraformrules/terraform_module_version.go
```diff
@@ -0,0 +1,118 @@
...
+	"fmt"
+	"log"
+	"regexp"
+
+	"github.com/terraform-linters/tflint/terraform/addrs"
+	"github.com/terraform-linters/tflint/terraform/configs"
+	"github.com/terraform-linters/tflint/tflint"
+)
+
+var exactVersionRegexp = regexp.MustCompile(`^=?\s*\d+\.\d+\.\d+$`)
```
wata727, line 13: I'm a little worried if this regex is enough. For instance, go-version uses the following regexp:
https://github.com/hashicorp/go-version/blob/v1.3.0/version.go#L18-L31

Ideally, it would be nice to be able to verify that the `f` in the `Constraint` is the `constraintEqual`, but this struct doesn't seem to expose the logic...
https://github.com/hashicorp/go-version/blob/v1.3.0/constraint.go#L30-L31

rules/terraformrules/terraform_module_version.go
bendrucker: Good catch, forgot those returns

rules/terraformrules/terraform_module_version.go
bendrucker: Agree. Spent a bunch of time looking at constraints and saw that the constraint type was not exposed. Was thinking of this more complicated regexp compilation:

https://github.com/hashicorp/go-version/blob/c2de2312d363381e4254abec260c6df2b1226483/constraint.go#L45-L49

But the version one is easy to reuse.

bendrucker: After trying this, the go-version regex doesn't require major/minor/patch and the test ensuring `1.0` is rejected as exact fails. Grabbed a regex from the semver.org docs and added a test ensuring that prerelease suffixes work.

terraform/copy/copy_dir.go
bendrucker: Removed `package copy`. I get the desire to reduce LOC from Terraform but the flip side is that modifying packages to remove dependencies on others makes it harder to copy in the latest version of a package from upstream.

bendrucker:
Revised per your comments, ready for another look

wata727 (approved):
👍

PR #1183, opened 2021-08-13
install: allow passing `$TFLINT_INSTALL_PATH`

bendrucker:
Supports a custom install path, `$TFLINT_INSTALL_PATH`, with the previous path as the default. 

Fixes #1182

### DIFF
--- a/install_linux.sh
+++ b/install_linux.sh
@@ -85,7 +85,7 @@ else
 fi
 
 echo "Cleaning /tmp/tflint.zip and /tmp/tflint ..."
-rm /tmp/tflint.zip /tmp/tflint
+rm -f /tmp/tflint.zip /tmp/tflint
 
 echo -e "\n\n===================================================="
 echo "Current tflint version"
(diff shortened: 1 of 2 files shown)


install_linux.sh
```diff
@@ -54,35 +54,38 @@ echo -e "\n\n===================================================="
...
   
   if [[ "$(id -u)" == 0 ]]; then SUDO=""; else
     SUDO="sudo";
   fi
 
-  $SUDO mkdir -p /usr/local/bin
-  $SUDO install -b -c -v /tmp/tflint /usr/local/bin/
+  
+  $SUDO mkdir -p "$dest"
+  $SUDO install -b -c -v /tmp/tflint "$dest"
```
williamboman, line 77: I think it'd be neat to also control whether it should be installed as root or not. My use case is having a fully non-interactive, isolated (as in, installs to a custom path and doesn't tamper with anything system-wide/global), installation process. Wdyt?

williamboman: I opened https://github.com/terraform-linters/tflint/pull/1184 with my suggestions, thanks for looking into this!

PR #1184, opened 2021-08-14
update install script to better support non-PATH installations

williamboman:


install_linux.sh
```diff
@@ -68,13 +70,13 @@ else
...
   
-  if [[ "$(id -u)" == 0 ]]; then SUDO=""; else
+  if [[ "$(id -u)" == 0 ]] || [[ $TFLINT_INSTALL_NO_ROOT == 1 ]]; then SUDO=""; else
     SUDO="sudo";
   fi
 
   
   $SUDO mkdir -p "$dest"
-  $SUDO install -b -c -v /tmp/tflint "$dest"
+  $SUDO install --backup=none -c -v /tmp/tflint "$dest"
```
williamboman, line 79: Perhaps scope creeping a but - but I suggest we change this to not create any backup, to avoid polluting the PATH. What happens now if you run the install script more than once is that you'll end up with a `tflint` and `tflint~` (the backup) in in your install destination.

install_linux.sh
```diff
@@ -68,13 +70,13 @@ else
   dest="${TFLINT_INSTALL_PATH:-/usr/local/bin}/"
   echo "Installing /tmp/tflint to ${dest}..."
   
-  if [[ "$(id -u)" == 0 ]]; then SUDO=""; else
+  if [[ "$(id -u)" == 0 ]] || [[ $TFLINT_INSTALL_NO_ROOT == 1 ]]; then SUDO=""; else
```
bendrucker, line 73: Seems like the script should just test whether it can access the directory as the non-root user and use sudo if not rather than adding an option

install_linux.sh
```diff
@@ -4,6 +4,8 @@ processor=$(uname -m)
 
 if [ "$processor" == "x86_64" ]; then
   arch="amd64"
+elif [ "$processor" == "arm64" ]; then
```
bendrucker, line 7: If you could submit this separately happy to merge it right away

williamboman: :+1: https://github.com/terraform-linters/tflint/pull/1185

install_linux.sh
williamboman: Good call :+1: https://github.com/terraform-linters/tflint/pull/1184/commits/a80c7639772e978647cb0a34b5eb2d3425b165de

PR #1185, opened 2021-08-15
install: support arm64 arch

williamboman:


install_linux.sh
```diff
@@ -41,7 +43,7 @@ else
 fi
 
 echo "Downloading TFLint $version"
-curl -L -o /tmp/tflint.zip "https://github.com/terraform-linters/tflint/releases/download/${version}/tflint_${os}.zip"
+curl --fail --silent -L -o /tmp/tflint.zip "https://github.com/terraform-linters/tflint/releases/download/${version}/tflint_${os}.zip"
```
williamboman, line 46: I added `--fail` because now it'd silently proceed, even when met with 404 (the subsequent `unzip` would fail instead). Also added `--silent` to reduce some noise. This both seems to be in line with other `curl` calls in this script

PR #1191, opened 2021-08-19
cmd: Change exit status

### LINKED ISSUE #528: 'Warning' exit status is 3 instead of 2
Hi,

According to [the README](https://github.com/terraform-linters/tflint#exit-statuses), my understanding is that a warning from the linter should return the exit status 3 (error is 2). 

Consider the following terraform code that should issue an error (bad instance type), with a return code of 2:

```hcl
variable "instance_type" {
  description = "ec2 instance type"
  default     = "m10.small"
}
```

Using tflint 0.13.1 (also 0.12), error code should be 2 and not 3: 

```shell
$ tflint
1 issue(s) found:

Error: instance_type is not a valid value (aws_instance_invalid_type)

  on ec2.tf line 3:
   3:   instance_type = var.instance_type

$ echo $?
3
```

Warnings do work correctly, though!
 
## Version

TFLint v0.13.1

wata727:
Fixes #528 

Change the exit status to what I originally intended.

Before:

- 0: No issues found
- 2: Errors occurred
- 3: No errors occurred, but issues found

After:

- 0: No issues found
- 1: Errors occurred
- 2: No errors occurred, but issues found

This is clearly a breaking change, but due to its natural behavior, it may be worth accepting.

### DIFF
--- a/cmd/cli.go
+++ b/cmd/cli.go
@@ -19,8 +19,8 @@ import (
 
 // Exit codes are int values that represent an exit code for a particular error.
 const (
-	ExitCodeOK    int = 0
-	ExitCodeError int = 1 + iota
+	ExitCodeOK int = iota
+	ExitCodeError
 	ExitCodeIssuesFound
 )
 
--- a/docs/user-guide/config.md
+++ b/docs/user-guide/config.md
@@ -52,8 +52,8 @@ CLI flag: `--force`
 Return zero exit status even if issues found. TFLint returns the following exit statuses on exit by default:
 
 - 0: No issues found
-- 2: Errors occurred
-- 3: No errors occurred, but issues found
+- 1: Errors occurred
+- 2: No errors occurred, but issues found
 
 ## `disabled_by_default`
 


bendrucker:
Seems like the original submission in #528 is focused on setting the exit code based on the issue with the highest severity level. I don't see that as a particularly compelling feature, FWIW. 

I'm 👍 for this, few users depend on specific non-zero exit codes anyway. You could hang on to it for a bit and release it with other breaking changes.

bendrucker:
Spotted something related to this:

https://golangci-lint.run/usage/configuration/

`golangci-lint` has an `--issues-exit-code` flag which defaults to 1. It's been in there for multiple years. Not really sure making this configurable is warranted but it did make me think of this discussion.

wata727:
Interesting. I'm not sure why golangci-lint offers this as a configurable option, but at least in my opinion, it doesn't seem necessary to do that. I can't find any problems that cannot be solved unless they can be configurable.

PR #1192, opened 2021-08-22
build: update generated code

chenrui333:
run `make code` to update the generated code.


bendrucker:
Hmm there's a check that does this and `git diff`, and idea why this didn't get picked up?

chenrui333:
yeah, I found it out when I tried go1.17 update (when I run `make code`). It looks like something got missed out before?

bendrucker:
Bug in the workflow, running `git diff` won't find newly generated files. Running `git add --intent-to-add .` first ensures they're tracked.

PR #1206, opened 2021-09-03
Build Docker image for linux/aarch64

ivy:
I noticed there's only a Docker image for `linux/amd64` and wanted to propose adding support for `linux/aarch64`. It's pretty useful to me as an Apple M1 and Docker Desktop user since I can get native performance out of tflint and avoid potential hiccups with x86_64 emulation. 😄

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -42,6 +42,7 @@ jobs:
       uses: docker/build-push-action@v2
       with:
         platforms: ${{ matrix.platform }}
+        load: true
         push: ${{ github.event_name != 'pull_request' }}
         tags: ${{ steps.meta.outputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -15,7 +15,7 @@ jobs:
       matrix:
         platform:
         - linux/amd64
-        - linux/aarch64
+        - linux/arm64/v8
     steps:
     - name: Checkout
       uses: actions/checkout@v2
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -11,11 +11,10 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        platform:
-        - linux/amd64
-        - linux/arm64/v8
+    env:
+      # Comma-delimited list of platforms to build for.
+      # See https://github.com/docker-library/official-images#multiple-architectures
+      DOCKER_PLATFORMS: linux/amd64,linux/arm64/v8
     steps:
     - name: Checkout
       uses: actions/checkout@v2
@@ -24,7 +23,7 @@ jobs:
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v1
       with:
-        platforms: ${{ matrix.platform }}
+        platforms: ${{ env.DOCKER_PLATFORMS }}
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@v1
     - name: Docker meta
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -11,11 +11,23 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform:
+        - linux/amd64
+        - linux/aarch64
```
bendrucker, line 18:
```suggestion
        - linux/arm64
```

Unless I'm missing something, seems like this should be arm64

https://stackoverflow.com/questions/31851611/differences-between-arm64-and-aarch64

`arm64` is self-documenting whereas with `aarch64` I might leave an `arm` comment next to it. Seems like they'd be interchangeable, possible even that only `arm64` works:

https://github.com/docker/setup-qemu-action/runs/3505773201?check_suite_focus=true 

Platforms:

```
linux/amd64,linux/arm64,linux/riscv64,linux/ppc64le,linux/s390x,linux/386,linux/mips64le,linux/mips64,linux/arm/v7,linux/arm/v6
```

.github/workflows/docker.yml
```diff
@@ -11,11 +11,23 @@ on:
...
     - name: Checkout
       uses: actions/checkout@v2
       with:
         submodules: true
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+      with:
+        platforms: ${{ matrix.platform }}
+    - name: Set up Docker Buildx
+      id: buildx
```
bendrucker, line 29:
```suggestion
```

Include an `id` only for steps that are referenced elsewhere in the job

.github/workflows/docker.yml
```diff
@@ -30,8 +42,9 @@ jobs:
     - name: Build and push
       uses: docker/build-push-action@v2
       with:
+        platforms: ${{ matrix.platform }}
```
bendrucker, line 45: Is this right to be doing a matrix build/push for each platform? Do new platforms push into existing images? 

If not, remove the matrix from this job and add a `platforms` output. Then, split `docker run` out into a `test` job. That job will be a matrix job, using the `platforms` output from the build job to set the `platform` list for the matrix.

bendrucker (changes requested):
Neat, first time seeing this!

https://github.com/docker/build-push-action/blob/v2/docs/advanced/multi-platform.md

I've approved the workflow run, take a look at the failure and my comments, happy to get this merged if you can get it working. FWIW I don't think the performance is much of an issue here (benchmarks appreciated if you've found otherwise) compared to the inconvenience of adding `--platform=linux/amd64`.

.github/workflows/docker.yml
ivy: Good callout! I'm more familiar with the `aarch64` identifier but it looks like the more correct name would be [`arm64v8`](https://github.com/docker-library/official-images#multiple-architectures) and more verbosely/explicitly, [`linux/arm64/v8`](https://github.com/docker-library/bashbrew/blob/v0.1.2/architecture/oci-platform.go#L14-L27) (following a format of `:os/:architecture/:variant`).

.github/workflows/docker.yml
ivy: Good question. After looking more closely with `docker image inspect`, this might not be working the way I'd assumed. 😅

ivy:
@bendrucker Thanks for your feedback! Unfortunately, it's a bit difficult to test these GitHub Actions myself but I think it _should_ work.

> FWIW I don't think the performance is much of an issue here

Definitely! In this particular project's case, the performance difference is negligible. If you're interested in offering native support like the rest of the binary releases, I'd still consider it valuable. I'd understand though if it wasn't something you or other maintainers wanted to worry about. 🤷‍♀️

bendrucker:
Seems like Actions doesn't consider you a "contributor" until a commit is associated with you in the repository's default branch. I thought workflow run approvals applied on a user's first PR/workflow run for a repository, but I guess we have to approve each run. 

https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/

bendrucker:
Speaking of performance...

The arm64 build is ~5x slower than amd64. To be expected I guess, goreleaser takes ~20 mins to cross-compile and upload release binaries. Not really an issue considering we're also building for Windows.

> necessary to "re-run" the build (everything's already cached) to load

Doesn't seem that way (cached), the `Test` step is taking as long as the previous build.

.github/workflows/docker.yml
```diff
@@ -30,8 +40,23 @@ jobs:
...
       uses: docker/build-push-action@v2
       with:
+        platforms: ${{ env.DOCKER_PLATFORMS }}
         push: ${{ github.event_name != 'pull_request' }}
         tags: ${{ steps.meta.outputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
-    - name: docker run
-      run: docker run ghcr.io/${{ github.repository }}:${{ steps.meta.outputs.version }} -v
+    - name: Test
+      # NOTE(ivy): Docker Buildx doesn't support the --load flag when multiple
```
bendrucker, line 48: Is there any issue we can link to on this?

wata727:
@ivy What's the status now? I'm interested in merging this feature.

wata727:
I've confirmed the comments and I think this pull request is ready to merge. Revert if there is any problem.
Thank you for your contributions!

PR #1207, opened 2021-09-03
ci(lint) : setup golangci-lint-action

mmorel-35:
Hi! I have set up golangci-lint-action. I also disabled several linters that where failing. You can re-enable them when you agree with them and they are fixed


bendrucker:
Thanks! Went ahead and got all the default linters enabled.

Issue #1245, opened 2021-11-02
Error: Failed to load '' : Multiple files in different directories are not allowed


hcylia:
Hello,

I am facing this error here but I don't understand the origin of it. I would like to ignore it or ignore the whole file. How can I proceed?

```
Terraform Lint...........................................................Failed
- hook id: tflint
- exit code: 1

Failed to parse CLI arguments. An error occurred:

Error: Failed to load ``: Multiple files in different directories are not allowed
```

bendrucker:
Without clear info on how to reproduce we can't really help. One sentence and output isn't enough for an issue. 

Seems like you're using a git hook there. Terraform files must be in the same directory so if the hook is calling otherwise that would be invalid.

hcylia:
@bendrucker In fact, we can't put all our Terraform files in the same repository, we kind of have this structure :

```
├── terraform
│   ├── .pre-commit-config.yaml
└── modules
    ├── module1
    │   ├── script1.1.tf
    │   └── script1.2.tf
    ├── module2
    │   └── script2.1.tf
    └── module3
        ├── script3.1.tf
        ├── script3.2.tf
        └── script3.3.tf
```

bendrucker:
Alright, well this is not a TFLint issue

hcylia:
My question is just how can I ignore this error/file. Thing like : « tflint:ignore » or I don’t know…

bendrucker:
You can't, this is an invalid way to call Terraform or TFLint. The scope is a directory. If you have multiple directories it is your responsibility to invoke the tool in each one. 

This is an issue with the git hook you're using.

fabasoad:
I use `pre-commit` tool for that and basically `tflint` provides a possibility to be used with `pre-commit` tool ([.pre-commit-hooks.yaml](https://github.com/terraform-linters/tflint/blob/master/.pre-commit-hooks.yaml) file). But in fact I can't use it there as this is the way as `pre-commit` works - scans all directories and run hook on each file (when I run `pre-commit run tflint --all-files`). So, I don't think that it's _an issue_ in the `pre-commit` tool but actually - by design.

Am I understand you correctly that the only solution here is to change files structure to store all `*.tf` files in a single folder? Another question - do you have any plans to give people a possibility to use multiple directories (by supporting multiple directories or by having a possibility to ignore the error)?

bendrucker:
[...]

PR #1247, opened 2021-11-04
remove precommit hooks

bendrucker:
* Having a `go fmt` hook isn't that important, most users integrate this somewhere in their editor
* TFLint does not need to offer this if it's not straightforward and well-maintained. Someone contributed this in 2019 and it's not been touched.

Meanwhile, there are actively maintained suites of hooks out there with good support for TFLint, users should head there.

https://github.com/antonbabenko/pre-commit-terraform

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,0 @@
-repos:
-- repo: git://github.com/dnephin/pre-commit-golang
-  rev: master
-  hooks:
-    - id: go-fmt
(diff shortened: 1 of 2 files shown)


asottile:
👋 hi -- this broke integrations -- can you restore the `.pre-commit-hooks.yaml` file?

bendrucker:
What integrations? See https://github.com/terraform-linters/tflint/issues/1245, this hook has been unmaintained for a while and is broken for a non-trivial fraction of users. Removing it is intentional. I'd assume users can specify a git ref if they want to continue using the existing one.

asottile:
the integration with `pre-commit`, it was still providing value (though doesn't work for some unconventional repository setups)

bendrucker:
A monorepo is not so unconventional. Look, I get this is a bummer for people who were happily using this. We have to make things right at some point and delaying doesn't actually help anyone. If there's a listing somewhere I'm happy to send a PR to remove the entry.

PR #1188, opened 2021-08-17
build: bump to use go1.17

chenrui333:
Update to the [latest go1.17](https://blog.golang.org/go1.17)

PR #1253, opened 2021-11-08
refactor: move from io/ioutil to io and os packages

Juneezee:
The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). Since tflint has been upgraded to Go 1.17 (#1188), this PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.

### DIFF
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -1,13 +1,13 @@
 package cmd
 
 import (
-	"io/ioutil"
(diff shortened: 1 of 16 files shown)


bendrucker (approved):

wata727 (approved):
Thanks!

PR #1267, opened 2021-12-02
enable errorlint linter

mmorel-35:


formatter/pretty.go
```diff
@@ -87,11 +88,12 @@ func (f *Formatter) printIssueWithSource(issue *tflint.Issue, sources map[string
 }
 
 func (f *Formatter) printErrors(err *tflint.Error, sources map[string][]byte) {
-	if diags, ok := err.Cause.(hcl.Diagnostics); ok {
+	var diags *hcl.Diagnostics
+	if errors.As(err.Cause, &diags) {
```
bendrucker, line 92: This should be `errors.As(err, &diags)` which may require amending the implementation of `tflint.Error` to play nicely with error wrapping. 

The `Cause` field in a `tflint.Error` is its own form of error chaining. `tflint.Error` should implement `As` to express the fact that `Cause` is considered part of the error chain.

> An error type might provide an As method so it can be treated as if it were a different error type.

formatter/json.go
```diff
@@ -76,7 +77,8 @@ func (f *Formatter) jsonPrint(issues tflint.Issues, tferr *tflint.Error) {
 
 	if tferr != nil {
 		var errs []error
-		if diags, ok := tferr.Cause.(hcl.Diagnostics); ok {
+		var diags *hcl.Diagnostics
```
bendrucker, line 80: Why is this a pointer and not a value? From `errors.As`:

> An error matches target if the error's concrete value is assignable to the value pointed to by target

bendrucker (changes requested):

formatter/json.go
mmorel-35: Otherwise, you'll meet this error :
```
errorsas: second argument to errors.As must be a non-nil pointer to either a type that implements error, or to any interface type (govet)
```

formatter/json.go
```diff
@@ -77,8 +77,8 @@ func (f *Formatter) jsonPrint(issues tflint.Issues, tferr *tflint.Error) {
 
 	if tferr != nil {
 		var errs []error
-		var diags *hcl.Diagnostics
-		if errors.As(tferr.Cause, &diags) {
+		var diags hcl.Diagnostics
+		if errors.As(tferr.Cause, diags) {
```
bendrucker: Per the error here, this still should be a pointer, `&diags`. Before, you had a pointer to a pointer. Now, you'll have a pointer to a value.

bendrucker (commented):
Thanks, I'll keep this open for more work and then merge it when complete. Replacing `%s` with `%w` isn't inherently meaningful/valuable unless you're actually using the error chaining APIs to check for those wrapped error types. 

I'd want to remove the `nolint` and actually implement wrapping for `tflint.Error`.

Some of these changes also need test coverage since they are currently uncovered. I don't see any coverage against errors caused by `hcl.Diagnostics`. Some of my initial feedback about incorrect pointer to pointer types should have resulted in test failures but didn't because of lack of coverage. Certainly no need to test `%w` but any `errors.As` branches I'd want to be covered.

wata727:
`tflint.Error` is removed in https://github.com/terraform-linters/tflint/pull/1292. I think this has improved the error wrapping and made it easier to introduce errorlint.
However, this PR seems to be out of date, so I'll close it. You can rework it by reopening it or opening a new PR.
Thank you for your contribution.

PR #1284, opened 2021-12-24
docs: linking terraform_module_version rule docs

PatMyron:
added in https://github.com/terraform-linters/tflint/pull/1182


wata727 (approved):
Thanks!

PR #1291, opened 2021-12-27
enable more linters and code coverage 

mmorel-35:


.github/workflows/build.yml
```diff
@@ -41,6 +41,12 @@ jobs:
           ${{ runner.os }}-go-
     - name: Run test
       run: make test
+    - name: Upload Test Coverage
+      uses: codecov/codecov-action@v2
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
+        flags: '${{ runner.os }},unit'
+        fail_ci_if_error: true
```
bendrucker, line 49: Overriding the default such that this workflow fails anytime Codecov is unavailable seems unnecessary

.github/workflows/build.yml
```diff
@@ -41,6 +41,12 @@ jobs:
           ${{ runner.os }}-go-
     - name: Run test
       run: make test
+    - name: Upload Test Coverage
```
bendrucker, line 44: Sending data to Codecov is more than just "enabling code coverage" and needs explanation. To continue reviewing these pull requests I need you to include some accompanying description of intent and other considerations and not just code.

This step depends on a secret that's not defined. Presumably you expect us to go sign up add it before merging this?

Have you reviewed alternatives? Is there a decent GitHub Action that might handle this without the additional service dependency, or is a fully hosted coverage tool worth it?

Code spells out _how_ something will happen but not necessarily _what_ or _why_. A PR's description should always speak to the questions of _what_ and _why_ so the review can consider whether the code accomplishes those goals effectively.

bendrucker (changes requested):
Please split this stuff up, "more linters" is not necessarily desirable. Formatting enhancements like grouping vars/constants don't really require much explanation/thought. But changing function signatures certainly does. Batching changes makes pull requests harder to review and ultimately lowers the chance that a contribution will be accepted. Small, focused PRs are always preferred.

We're also not interesting in style thrashing in the `terraform` package, since that's forked from the `hashicorp/terraform` repo.

langserver/initialize.go
```diff
@@ -1,13 +1,10 @@
 package langserver
 
 import (
-	"context"
-
 	lsp "github.com/sourcegraph/go-lsp"
-	"github.com/sourcegraph/jsonrpc2"
 )
 
-func initialize(ctx context.Context, conn *jsonrpc2.Conn, req *jsonrpc2.Request) (result interface{}, err error) {
```
mmorel-35, line 10: the unparam linter found that none of the parameters are being used in the function

langserver/workspace_did_change_watched_files.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/terraform-linters/tflint/tflint"
 )
 
-func (h *handler) workspaceDidChangeWatchedFiles(ctx context.Context, conn *jsonrpc2.Conn, req *jsonrpc2.Request) (result interface{}, err error) {
+func (h *handler) workspaceDidChangeWatchedFiles(ctx context.Context, conn *jsonrpc2.Conn) (result interface{}, err error) {
```
mmorel-35, line 15: the unparam linter found that the req parameter is not being used in the function

formatter/sarif.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/terraform-linters/tflint/tflint"
 )
 
-func (f *Formatter) sarifPrint(issues tflint.Issues, tferr *tflint.Error, sources map[string][]byte) {
```
mmorel-35, line 10: the unparam linter found that tferr and sources are not being used in the function

.github/workflows/build.yml
mmorel-35: I've removed everything concerning code coverage from this PR to make it focused on the linters. 
Codecov is the most commonly used coverage tool on the project I've seen on github. There are probably other solutions but to my knowledge all of them require to send a file summarizing the result of the coverage of your files.

> This step depends on a secret that's not defined. Presumably you expect us to go sign up add it before merging this?

I took it from [here](https://github.com/codecov/codecov-action#usage), It seems unnecessary when your repository is public and it doesn't seems to fail when this token is missing but if it's not used it's better to not include it

PR #1297, opened 2022-01-15
json,sarif: print structured error data

### LINKED ISSUE #1163: Terraform parser errors should be reported in formatted outputs
When using the checkstyle or junit formatter, users may run TFLint with invalid, un-parseable configuration. Invalid configuration can result from subtle mistakes like typos (`variabl "foo" {}`) or errant quotes (`variable "foo" { type = "string" }`).

Currently, TFLint treats all errors that result from loading configuration as fatal and prints them and exits with a non-zero code. While this is still reasonably correct, users often do not understand the "Failed to load configurations" prefix and believe TFLint or specific rules to be broken. 

These errors are reported back with diagnostics and could be returned as JUnit or Checkstyle output. In cases where failure to load configuration does return diagnostics, we should pass them through.

sxlijin:
Teach `--format json` and `--format sarif` to print structured error data. The former currently discards errors with severity warning, and the latter completely ignores `tferr`.

This PR is still in a draft stage (code complete, but missing unit/e2e tests) and is meant to solicit feedback on whether y'all think this makes sense or if you'd prefer to go in a different direction on this.

----

The context behind this PR is that I work on [`trunk`](https://trunk.io/products/check), a universal linter which makes it easy to set up and run linters for all the different types of tooling that developers use, and while looking into setting up `tflint` I was surprised that `--format default`, `--format json`, and `--format sarif` were returning wildly different things on an example repo (I was using [mozilla/partinfra-terraform](https://github.com/mozilla/partinfra-terraform) to experiment).

A bit of yak shaving later, and here we are!

Closes #1163

### DIFF
--- a/tflint/errors.go
+++ b/tflint/errors.go
@@ -1,6 +1,7 @@
 package tflint
 
 import (
+	"errors"
 	"fmt"
 )
 
@@ -52,6 +53,11 @@ func (e *Error) Error() string {
 	}
 
 	return e.Message
+}
+
+// As allows the error to be used with errors.As, detecting whether the underlying Cause has an error of the given type in its chain.
+func (e *Error) As(target interface{}) bool {
+	return errors.As(e.Cause, target)
 }
 
 // NewContextError makes a new context error
--- a/tflint/errors.go
+++ b/tflint/errors.go
@@ -2,6 +2,7 @@ package tflint
 
 import (
 	"fmt"
+
 	hcl "github.com/hashicorp/hcl/v2"
 )
 
--- a/tflint/errors.go
+++ b/tflint/errors.go
@@ -1,6 +1,9 @@
 package tflint
 
-import "fmt"
+import (
+	"fmt"
+	hcl "github.com/hashicorp/hcl/v2"
+)
 
 const (
 	// EvaluationError is an error when interpolation failed (unexpected)
@@ -22,7 +25,7 @@ const (
 	// ContextError is pseudo error code for propagating runtime context.
 	ContextError string = "I:Context"
 
-	// FatalLevel is a recorverable error, it cause panic
+	// FatalLevel is an unrecoverable error, it causes a panic
 	FatalLevel string = "Fatal"
 	// ErrorLevel is a user-level error, it display and feedback error information
 	ErrorLevel string = "Error"
@@ -50,6 +53,14 @@ func (e *Error) Error() string {
 	}
 
 	return e.Message
+}
+
+type ConfigParseError struct {
+	Detail *hcl.Diagnostics
+}
+
+func (e ConfigParseError) Error() string {
+	return (*e.Detail).Error()
 }
 
 // NewContextError makes a new context error
--- a/tflint/errors.go
+++ b/tflint/errors.go
@@ -2,8 +2,6 @@ package tflint
 
 import (
 	"fmt"
-
-	hcl "github.com/hashicorp/hcl/v2"
 )
 
 const (
@@ -54,15 +52,6 @@ func (e *Error) Error() string {
 	}
 
 	return e.Message
-}
-
-// TerraformConfigParseError is an error encountered by TFLint when parsing the user's Terraform configuration.
-type TerraformConfigParseError struct {
-	Diags hcl.Diagnostics
-}
-
-func (e TerraformConfigParseError) Error() string {
-	return e.Diags.Error()
 }
 
 // NewContextError makes a new context error
--- a/tflint/errors.go
+++ b/tflint/errors.go
@@ -56,12 +56,13 @@ func (e *Error) Error() string {
 	return e.Message
 }
 
-type ConfigParseError struct {
-	Detail hcl.Diagnostics
+// TerraformConfigParseError is an error encountered by TFLint when parsing the user's Terraform configuration.
+type TerraformConfigParseError struct {
+	Diags hcl.Diagnostics
 }
 
-func (e ConfigParseError) Error() string {
-	return e.Detail.Error()
+func (e TerraformConfigParseError) Error() string {
+	return e.Diags.Error()
 }
 
 // NewContextError makes a new context error
--- a/tflint/errors.go
+++ b/tflint/errors.go
@@ -57,11 +57,11 @@ func (e *Error) Error() string {
 }
 
 type ConfigParseError struct {
-	Detail *hcl.Diagnostics
+	Detail hcl.Diagnostics
 }
 
 func (e ConfigParseError) Error() string {
-	return (*e.Detail).Error()
+	return e.Detail.Error()
 }
 
 // NewContextError makes a new context error
--- a/formatter/checkstyle.go
+++ b/formatter/checkstyle.go
@@ -7,18 +7,18 @@ import (
 	"github.com/terraform-linters/tflint/tflint"
 )
 
-type checkstyleError struct {
+type checkstyleIssue struct {
 	Rule     string `xml:"rule,attr"`
 	Line     int    `xml:"line,attr"`
 	Column   int    `xml:"column,attr"`
 	Severity string `xml:"severity,attr"`
 	Message  string `xml:"message,attr"`
-	Link     string `xml:"link,attr"`
+	Link     string `xml:"link,attr,omitempty"`
 }
 
 type checkstyleFile struct {
 	Name   string             `xml:"name,attr"`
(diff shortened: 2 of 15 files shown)


bendrucker:
See #1163 which explains this exact issue. All the formatters print the exact same data. When `tflint` encounters configuration parsing errors, these could in theory be returned back by formatters too, but are not currently. 

However, we're probably not inclined to include arbitrary TFLint errors as formatted output, only those arising from parsing user configuration.

wata727 (commented):
Thank you for opening this PR. I'm not sure why this change is needed for trunk, so let me ask you questions.

sxlijin:
Thanks for the feedback!

Re trunk: this PR does not block integrating tflint. Originally I thought there was a bug in the implementation of `--format sarif` because when running on a repo with HCL errors, I thought the HCL errors were lint findings, e.g.:

```
sxlijin@sxlijin-devbox:2022-01-14T12:39:32-08:00:~/repos/partinfra-terraform (master) [0]
$ ~/playground/tflint . 2>&1 | tail
++ 2022-01-14T12:43:14-08:00 ++ tail
++ 2022-01-14T12:43:14-08:00 ++ /home/sxlijin/playground/tflint .

Terraform 0.11 and earlier required type constraints to be given in quotes, but that form is now deprecated and will be removed in a future version of Terraform. Remove the quotes around "map" and write map(string) instead to explicitly indicate that the map elements are strings.

Error: Invalid quoted type constraints

  on variables.tf line 61, in variable "unmanaged_role_arns":
  61:     type = "map"

Terraform 0.11 and earlier required type constraints to be given in quotes, but that form is now deprecated and will be removed in a future version of Terraform. Remove the quotes around "map" and write map(string) instead to explicitly indicate that the map elements are strings.
```

When I finally figured out that these were HCL parse(?) errors, I was so deep into the tflint code that I figured I would just put this up as a PR.

The other thing that was really confusing is that these were only showing up with the default formatter; `--format json` was hiding the `warning` parse errors and `--format sarif` was showing nothing.

> However, we're probably not inclined to include arbitrary TFLint errors as formatted output, only those arising from parsing user configuration.

I can get behind this; so every formatter will do something like 

```
func (f *Formatter) printErrors(err *tflint.Error, sources map[string][]byte) {
        if diags, ok := err.Cause.(hcl.Diagnostics); ok {
                printAsFormattedOutput
        } else {
                fmt.Fprintf(f.Stderr, "%s. An error occurred:\n\n", err.Message)
                fmt.Fprintf(f.Stderr, "%s: %s\n\n", colorError("Error"), err.Cause.Error())
        }
}
```

> I think printing it as an issue can be confusing. There is a clear distinction between issues and errors, and issues have a mechanism that can be ignored by annotations and settings, but this is not available in this way. Also, it will surprise users that the rule names are automatically generated.

Agreed that printing as an issue can be confusing; this PR deliberately puts them in a different section deliberately (`ret['errors']` as opposed to `ret['issues']` in `json`, and `tool.driver = "tflint-errors"` as opposed to `tool.driver = "tflint"` in `sarif`, see below for examples).

Re ignoring: I don't think the default formatter allows ignoring errors though?

json:

```
{'errors': [
            {'callers': None,
             'message': 'Terraform 0.11 and earlier required type constraints '
                        'to be given in quotes, but that form is now '
                        'deprecated and will be removed in a future version of '
                        'Terraform. Remove the quotes around "map" and write '
                        'map(string) instead to explicitly indicate that the '
                        'map elements are strings.',
             'range': {'end': {'column': 17, 'line': 61},
                       'filename': 'variables.tf',
                       'start': {'column': 12, 'line': 61}},
             'rule': {'link': '',
                      'name': 'Invalid quoted type constraints',
                      'severity': 'error'}},
              [...],
              ],
 'issues': []}
```

sarif:

```
{
  "version": "2.1.0",
  "$schema": "https://json.schemastore.org/sarif-2.1.0-rtm.5.json",
  "runs": [
    {
      "tool": {
        "driver": {
          "name": "tflint",
          "informationUri": "https://github.com/terraform-linters/tflint"
        }
      },
      "results": []
    },
    {
      "tool": {
        "driver": {
          "name": "tflint-errors",
          "informationUri": "https://github.com/terraform-linters/tflint"
        }
      },
      "results": [
        {
          "ruleId": "Invalid argument name",
          "level": "error",
          "message": {
            "text": "Argument names must not be quoted."
          },
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "bitergia.tf"
                },
                "region": {
                  "startLine": 34,
                  "startColumn": 9,
                  "endLine": 34,
                  "endColumn": 10,
                  "byteOffset": 686,
                  "byteLength": 1
                }
              }
            }
          ]
        },
        [...]
      ],
    },
  ],
}
```

Let me know if there's anything I missed.

bendrucker:
Re `Diagnostics`, I think we'd be better of wrapping the site where we call TF to parse the config. For example, return a new `ConfigParseError` type or similar. Otherwise the formatter depends on the implementation of the parser. Both the `hcl` and `terraform` packages implement `Diagnostics` IIRC so it can be particularly tricky to try to assert/unwrap these errors correctly far from where they were returned.

sxlijin:
> Re `Diagnostics`, I think we'd be better of wrapping the site where we call TF to parse the config. For example, return a new `ConfigParseError` type or similar. Otherwise the formatter depends on the implementation of the parser. Both the `hcl` and `terraform` packages implement `Diagnostics` IIRC so it can be particularly tricky to try to assert/unwrap these errors correctly far from where they were returned.

Makes sense. `ConfigParseError` I think precludes surfacing parse errors for `arbitrary_file_in_repo.tf` though, so maybe `StructuredParseError`?

bendrucker:
Yeah, `Config` there refers to [Terraform config](https://www.terraform.io/docs/glossary#terraform-configuration). Hopefully the package should help clarify that b/c parsing Terraform config dirs and TFLint config files would be handled by separate packages.

sxlijin:
Hmm. So right now, the pretty printer is pretty intrinsically tied to `hcl.Diagnostics` since it delegates to `hcl.DiagnosticWriter`, and `tfdiags.Diagnostics` has a `toHCL` method on it (I'm assuming that's what you were referring to when you said `hcl` and `terraform` were separate packages... if not I'm not sure what you were referring to).

Based on this, it seems like `hcl.Diagnostic` is actually the common format we want to pass around- is there something I'm missing?

bendrucker:
It's a fine way to pass around the error data, just perhaps as a field or embedded struct so that TFLint is only matching its own wrapper type and not one supplied by another module.

sxlijin:
OK, updated per @bendrucker's feedback to use a `ConfigParseError` and added test coverage. (No integration test coverage for sarif, just unit tests.)

nvuillam:
This PR is highly awaited by MegaLinter, because today tflint only outputs empty SARIF files :)

image

wata727 (approved):
Thank you for your reply. Understood. It looks good to me.
Once @bendrucker approved, please merge this PR.

tflint/errors.go
```diff
@@ -52,6 +56,14 @@ func (e *Error) Error() string {
 	return e.Message
 }
 
+type ConfigParseError struct {
+	Detail *hcl.Diagnostics
```
bendrucker, line 60: Go slices (and maps) are implemented with pointers, sort of like this:

```go
type StringSlice struct {
  items []*string
}
```

You can point a slice to a new list of items, but when you pass it around you're not copying the items themselves, just pointers. This is why you'll often see `[]*Widgets` in Go code but rarely `*[]Widgets`.

I'll take care of converting this, just wanted to leave a note for myself and an explanation if you're interested.

bendrucker:
I have time planned to review this later today. I see a small tweak or two I want to make but will just handle that myself and get this merged.

tflint/errors.go
sxlijin: Ahh, I missed that `hcl.Diagnostics` is just a typedef. Fixed!

bendrucker:
I'd pictured `ConfigParseError` being usable in `tflint.Loader`, directly wrapping sites where Terraform configuration is parsed. After looking at the numbered of returned `diags` there are across many methods on a `Loader` as well as the formatter dependency on `hcl` for severity mapping, I think the code's better off without it. Thanks for trying it.

sxlijin:
Hah, yeah, that makes a lot of sense. I definitely sympathize with the desire to isolate dependencies :)

PR #1309, opened 2022-02-08
Add change to readme to map tflint.d directory so we can init

babyhuey:
If we don't do this, then it is more difficult to run init and then run tflint. This maps that so that it gets added to your home directory so all projects can use the initialized tflint

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Instead of installing directly, you can use the Docker images:
 Example:
 
 ```console
-docker run --rm -v $(pwd):/data -t ghcr.io/terraform-linters/tflint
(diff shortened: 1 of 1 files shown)


bendrucker (changes requested):
This might be the right choice for you but it's not necessarily a suitable default. Relying on host mounts to share data across commands run in different containers is viable but not ideal. 

We'd accept an example of how to run multiple commands in a single container via a shell if you'd like to add one.

babyhuey:
Out of curiosity, what is your current recommended way of doing an init inside of the container? Up to the user to map/figure out?

babyhuey:
Or what about changing the entrypoint to automatically run the init?

wata727:
I think there is room for improvement in the Docker image as a pluggable linter. The current usage is designed for TFLint running as a single binary and may not be suitable for the current architecture. However, I haven't thought about what to do yet.

Issue #1056, opened 2021-02-02
Is there a way to use all existing rules


shchowdhury:
Hi,
I am total newbie of tflint. trying out its features. Looking into https://github.com/terraform-linters/tflint/tree/master/docs/rules, so these rules are already built-in in the tflint binary. However, 3 is enabled by default. is there a shortcut that I use says to tflint, use all those (14 i believe) during execution. 

Dont want to individually add each rule in the config file and enable them! 

rule "terraform_naming_convention" {
  enabled = true
}

rule "terraform_documented_variables" {
  enabled = true
}

.....

wata727:
Since these rules will be cut out into plugins in the future, enabling/disabling each plugin may solve this problem.

tgdfool2:
Any news on this one? Any timeframe when the default rules will be packed in plugins?

Zialus:
So, is there an official answer to this question?

Zialus:
So, is there an official answer to this question?

alexanderdavide:
Awesome, [enabling all terraform plugin rules](https://github.com/terraform-linters/tflint/pull/1496) is now possible by upgrading to TFLint v0.40.0 and adding.

```
plugin "terraform" {
  enabled = true
}
```

Note that `preset = "recommended"`, which is now the default when you don't have an explicit "terraform" plugin configuration, already has [more rules](https://github.com/terraform-linters/tflint-ruleset-terraform/pull/3) enabled than < v0.40.0.

PR #1311, opened 2022-02-11
Add option --enable-all to enable all rules

dvob:
It would be nice to have a way to enable all available rules.
It seems that there are also other people requesting this feature #1056

### DIFF
--- a/cmd/option.go
+++ b/cmd/option.go
@@ -18,6 +18,7 @@ type Options struct {
 	IgnoreModules  []string `long:"ignore-module" description:"Ignore module sources" value-name:"SOURCE"`
 	EnableRules    []string `long:"enable-rule" description:"Enable rules from the command line" value-name:"RULE_NAME"`
 	DisableRules   []string `long:"disable-rule" description:"Disable rules from the command line" value-name:"RULE_NAME"`
+	EnableAllRules bool     `long:"enable-all" description:"Enable all rules" value-name:"ENABLE_ALL"`
 	Only           []string `long:"only" description:"Enable only this rule, disabling all other defaults. Can be specified multiple times" value-name:"RULE_NAME"`
 	EnablePlugins  []string `long:"enable-plugin" description:"Enable plugins from the command line" value-name:"PLUGIN_NAME"`
 	Varfiles       []string `long:"var-file" description:"Terraform variable file name" value-name:"FILE"`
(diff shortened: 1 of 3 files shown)


wata727:
Thank you for your contribution!
Unfortunately, there are still many things to consider in order to achieve this goal.

First, this change only enables some of the Terraform rules built into the TFLint core, not rules provided by plugins. To enable all plugin rules, you need to add an RPC endpoint to plugin and call it from TFLint core.

Second, we can't enable rules that require settings like `aws_s3_bucket_name` automatically.
https://github.com/terraform-linters/tflint-ruleset-aws/blob/v0.12.0/docs/rules/aws_s3_bucket_name.md

For these reasons, I don't think we should add an option to enable all rules uniformly.
It is a historical reason that Terraform rules are included in the core, and I believe that it is better to cut this out into a ruleset so that it can be controlled in units of rulesets.

dvob:
Hi @wata727 thank you for maintaining tflint.

I was looking for a way to find unused variables in Terraform files and found out that this should be possible with tflint via this issue https://github.com/hashicorp/terraform/issues/11412#issuecomment-624452494
I downloaded tflint but I did not understand why tflint did not report the unused variables. It was not clear to me that only certain rules are enabled by default. Then looked for a way to enable all rules. Whats actually  the reason that not all builtin rules are enabled?

In my opinion it would be nice to have the the most common and provider independent checks built into the TFLlint core. This way using tflint would be super simple. You could download tflint and run it in the directory where you have your Terraform files without further configuration. And then, only if you have unwanted warnings or if you want to add additional checks, you can start to configure tflint to disable rules or add plugins.

For me it would be ok if this option would only enable the built in rules but I understand that this could be confusing for people which use plugins. So I understand if you don't want this change in tflint and its OK if you close this PR.

gtirloni:
I'm just trying tflint on a large project and I was surprised that it did not detect _anything_ wrong. It took me some time to realize that rules were not being enabled at all and I had to specify them one by one (either in the config file or command line). Had I not dug deeper into the issue I would have thought tflint didn't do anything at all.

It might be worth to enable at least all core rules that don't require customization from users.

dvob:
@wata727 Somehow i missed your last reply. 

How do I install the strict rule set?

I think your design decision is perfectly fine. But if you want to solve it like this i suggest that all rules are disabled by default and a warning is presented to the user if no rules are enabled or if no configuration was found.

This way if a user just downloads tflint and expect it to check certain basic rules they get a warning that they have to configure the rules themself. At least for me such a hint would have been very helpful.

wata727:
Thank you for the advice. To be sure, the current UX is not ideal. The biggest problem is that the terraform rules have been built into the core for historical reasons, and I believe that we can improve it to design a better UX.

wata727:
The current design cannot accept this feature and will close it. Thank you for your contribution.

PR #1321, opened 2022-03-03
feat: get architecture for install

techsolx:
* add bash function to get architecture
    * match arch to available download assets

I noted that `uname -m` will return aarch64 or aarch,
for arm based machines these changes will allow `os` var to match
multiple architectures and existing download assets.

Tested on darwin x86_64 (amd64), and linux amd64 and arm64.

### DIFF
--- a/install_linux.sh
+++ b/install_linux.sh
@@ -1,14 +1,18 @@
 #!/bin/bash -e
 
-processor=$(uname -m)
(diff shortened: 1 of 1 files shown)


bendrucker:
Would be nice to have test coverage, attempted that and left notes in #1322

PR #1322, opened 2022-03-03
test: run intall script on multiple container images

bendrucker:
Runs the installer test on multiple container images to help cover alternate OS configurations


bendrucker:
Kind of surprising to see `actions/checkout` fail. In any case https://github.com/uraimo/run-on-arch-action seems like a better direction.

PR #1359, opened 2022-04-17
feat: add new rule to check comparisons with empty list

bmbferreira:
As explained by @jbardin on [this issue](https://github.com/hashicorp/terraform/issues/23562#issuecomment-564071042) "..._the preferred way to check for an empty set or series of any type would be to check its length, rather than compare it against a literal value_". 
I've had a bug on my terraform code that took me almost 2 days to find out the root cause, just to find out that we had a check for `[]` in our code that was returning always false. 😓 
Since we are already using tflint, having a check for it would be a really good way to avoid doing the same stupid mistake again! 😃 Hopefully, you will also find this new rule useful and everyone can save a lot of headaches with it as well.

More info about why this rule is "needed":
https://discuss.hashicorp.com/t/conditional-for-empty-list/26958/2
https://github.com/hashicorp/terraform/issues/25798#issuecomment-675721098


test.tf
```diff
@@ -0,0 +1,7 @@
+variable "my_list" {
```
bendrucker, line 1: This is a leftover

docs/rules/terraform_empty_list_check.md
```diff
@@ -0,0 +1,36 @@
+# terraform_empty_list_check
```
bendrucker, line 1: `check` is unclear/generic. Seems like `equality` could work here.

rules/terraformrules/terraform_empty_list_check_test.go
```diff
@@ -0,0 +1,85 @@
...
+		Expected tflint.Issues
+	}{
+		{
+			Name: "comparing with [] is not recommended",
+			Content: `
+variable "my_list" {
+	type = list(string)
+}
+resource "aws_db_instance" "mysql" {
+	count = var.my_list == [] ? 0 : 1
```
bendrucker, line 23: Make sure to test the inverted order, `[] == var.my_list`

rules/terraformrules/terraform_empty_list_check.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+		return nil
+	})
+}
+
+func checkEmptyList(tupleConsExpr *hclsyntax.TupleConsExpr, runner *tflint.Runner, r *TerraformEmptyListCheckRule, binaryOpExpr *hclsyntax.BinaryOpExpr) {
+	if len(tupleConsExpr.Exprs) == 0 {
+		runner.EmitIssue(
+			r,
+			"List is compared with [] instead of checking if length is 0.",
```
bendrucker, line 84:
```suggestion
			"Comparing a collection with an empty list is invalid. To detect an empty collection, check its length.",
```

Seems like this would apply to checking an empty set as well, thus collection over list. If not list is fine.

Since the error is recommending a fix to the user, use active voice to make that clear.

bendrucker (commented):
Nice, few minor comments

bmbferreira:
@bendrucker thank you for reviewing it so quickly! 🙇  

Just fixed all the issues reported.

Have a great day!

bendrucker:
Will get back on this tomorrow

bmbferreira:
> Will get back on this tomorrow

Thank you!

rules/terraformrules/terraform_empty_list_equality.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"log"
+
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/terraform-linters/tflint/tflint"
+)
+
+// TerraformEmptyListCheckRule checks whether is there a comparison with an empty list
+type TerraformEmptyListCheckRule struct{}
```
bendrucker, line 12: This and other related identifiers should be EmptyListEqualityRule

rules/terraformrules/terraform_empty_list_equality.go
```diff
@@ -0,0 +1,88 @@
...
+	for filename := range files {
+		if err := r.checkEmptyList(runner, filename); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *TerraformEmptyListCheckRule) checkEmptyList(runner *tflint.Runner, filename string) error {
```
bendrucker, line 62: `filename` is received but not used here

rules/terraformrules/terraform_empty_list_equality.go
```diff
@@ -0,0 +1,88 @@
...
+func (r *TerraformEmptyListCheckRule) Check(runner *tflint.Runner) error {
+	if !runner.TFConfig.Path.IsRoot() {
+		// This rule does not evaluate child modules.
+		return nil
+	}
+
+	log.Printf("[TRACE] Check `%s` rule for `%s` runner", r.Name(), runner.TFConfigPath())
+
+	files := make(map[string]*struct{})
+	for _, variable := range runner.TFConfig.Module.Variables {
```
bendrucker, line 49: Not sure how this relates, i.e. iterating through all variable declarations and using those filenames. Doesn't seem like that's needed?

rules/terraformrules/terraform_empty_list_equality_test.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+func Test_TerraformEmptyListCheckRule(t *testing.T) {
+	cases := []struct {
+		Name     string
+		Content  string
+		Expected tflint.Issues
+	}{
+		{
+			Name: "comparing with [] is not recommended",
```
bendrucker, line 17: A simpler test would just be `count = [] == [] ? 0 : 1`. The rule only works on files with variables in them and because that is also true in the test cases that isn't caught by the tests. Keeping the test minimal helps prevent that situation.

bendrucker (changes requested):

rules/terraformrules/terraform_empty_list_equality.go
bmbferreira: bad copy pasta. removed

rules/terraformrules/terraform_empty_list_equality.go
bmbferreira: done

rules/terraformrules/terraform_empty_list_equality.go
bmbferreira: again, bad copy pasta from a different rule. thank you

rules/terraformrules/terraform_empty_list_equality_test.go
bmbferreira: done. Replaced the two tests to just one with `count = [] == [] ? 0 : 1` that returns two results

bmbferreira:
sorry to bother you @bendrucker, but please let me know if there's still something missing on the PR 😃 

thank you! and have an awesome week!

wata727 (approved):
It looks good to me. If @bendrucker approves, merge this pull request.

bmbferreira:
Thanks for the feedback @sergei-ivanov! Can you report this as an issue so I can tackle it when I have some time please? 

🙇

sergei-ivanov:
@bmbferreira I have created #1424 -- thank you very much for your contribution!

Issue #1325, opened 2022-03-07
Please add a way to force colors in the output


mbyio:
We run tflint in a monorepo containing many terraform workspaces and modules. I want to run one invocation of tflint for each workspace and module. I use `find` to get paths to all the workspaces and modules, and then I use `parallel` to invoke tflint for each of them. However, when I use `parallel` to invoke `tflint`, I think `tflint` detects that it isn't a TTY, and it turns off colors. I actually want colors because it makes the output much easier to read, so I'd like to force enable them using a flag like `--color`.

PatMyron:
related for reference: https://github.com/terraform-linters/tflint/pull/415

PR #1370, opened 2022-05-03
cmd: Add `--color` option

wata727:
Fixes https://github.com/terraform-linters/tflint/issues/1325

This PR adds a new `--color` option. This option forces colorized output to be enabled. Colored output is enabled by default, but is automatically disabled if a tty is not attached. The `--color` option always enables colorized output, even if the tty is not attached. You can see the behavior by output via a pipe.

```bash
$ tflint | less -R
# Disabled color

$ tflint --color | less -R
# Enabled color
```


mbyio:
Thank you!!

Issue #1388, opened 2022-05-19
Add support for Terraform v1.2 syntax


lostplan:
Changes: https://github.com/hashicorp/terraform/blob/v1.2/CHANGELOG.md

joe-a-t:
Specific issues to call out where the linter fails include `precondition` and `postcondition` as well as [using variables in variable validation error messages](https://github.com/hashicorp/terraform/pull/30613)

onematchfox:
Adding `replace_triggered_by` to the list please.

PR #1401, opened 2022-05-27
Add support for `replace_triggered_by` in `lifecycle`

lonegunmanb:
As #1388 said we need to support new feature in v1.2.0.

Hello @wata727 , this is the first time I submit pr to tflint repo, I'm not sure the whether this patch meets tflint's requirement, would you please give it a review since I have pending issue depends on `replace_triggered_by`? Thanks a lot!

### DIFF
--- a/terraform/configs/resource.go
+++ b/terraform/configs/resource.go
@@ -42,6 +42,7 @@ type ManagedResource struct {
 	PreventDestroy      bool
 	IgnoreChanges       []hcl.Traversal
 	IgnoreAllChanges    bool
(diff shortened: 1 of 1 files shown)


bendrucker (approved):
The entire `terraform` package should be updated for 1.2, happy to merge that if you want to give it a shot. The intent of that package is to match TF and we won't accept individual patches.

bendrucker (changes requested):
Accidentally set the previous review as approving

wata727:
This individual patch is not accepted by itself, but we can start supporting v1.2 based on this patch. I will start working after merging this. Thanks.

PR #1407, opened 2022-06-02
feat: added support for codeclimate output format

ST-Apps:
This is mostly meant to be used in GitLab CI as a way to provide a code quality report from TFLint.

As of today, getting a code quality report requires building a Docker image that adds `jq` to `tflint` and to manually manipulate the provided json output to match the expected format.

EDIT: as a side not, I'm not a Go expert so please provide feedback on what can be improved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,26 +115,26 @@ Usage:
   tflint [OPTIONS] [FILE or DIR...]
 
 Application Options:
-  -v, --version                                                 Print TFLint version
-      --init                                                    Install plugins
-      --langserver                                              Start language server
-  -f, --format=[default|json|checkstyle|junit|compact|sarif]    Output format
-  -c, --config=FILE                                             Config file name (default: .tflint.hcl)
-      --ignore-module=SOURCE                                    Ignore module sources
-      --enable-rule=RULE_NAME                                   Enable rules from the command line
-      --disable-rule=RULE_NAME                                  Disable rules from the command line
-      --only=RULE_NAME                                          Enable only this rule, disabling all other defaults. Can be specified multiple times
-      --enable-plugin=PLUGIN_NAME                               Enable plugins from the command line
-      --var-file=FILE                                           Terraform variable file name
-      --var='foo=bar'                                           Set a Terraform variable
-      --module                                                  Inspect modules
-      --force                                                   Return zero exit status even if issues found
-      --color                                                   Enable colorized output
-      --no-color                                                Disable colorized output
-      --loglevel=[trace|debug|info|warn|error]                  Change the loglevel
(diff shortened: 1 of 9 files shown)


bendrucker:
Please start with better documentation of the need for this. If the purpose is just to extract a single key from the existing JSON output we are unlikely to accept a proprietary output format for a commercial product.

ST-Apps:
Hi @bendrucker, let me try to explain what's the need behind this.

GitLab provides a free code quality scan feature that is available on all tiers, as explained [here](https://docs.gitlab.com/ee/user/project/merge_requests/code_quality.html#implementing-a-custom-tool), so I don't believe this is strictly related to a commercial product.

The format they use is described [here](https://docs.gitlab.com/ee/user/project/merge_requests/code_quality.html#implementing-a-custom-tool) and it's a subset of the Code Climate one (which is commercial indeed but it's not the reason why I started this PR).

Their code scanning feature allows people to write custom scanners that are able to populate the Merge Request widget with the results from code analysis, as long as this returns an artifact with that specific format.
While the current `json` formatter can still be adapted to provide a valid input for GitLab, this requires a lot of manipulation and it's not a straight-forward process.

Having this formatter allows people to just plug `tflint` in their GitLab pipeline and effortlessly see the results in their Merge Requests.

Again, this is something beneficial to GitLab users with any license tier, free one included, so I believe it makes sense to have a native way to output this kind of report.

I hope this clarifies some of the doubts, if not I'm available to provide further details.

bendrucker:
Code Climate is the commercial product here. Note that all existing formatters are open, community-defined formats.

ST-Apps:
I'm sorry but I'm not sure to understand what you're saying.

The implemented format is GitLab's one which is a subset of CodeClimate, but the goal of the PR is to support GitLab not CodeClimate. If the issue is having references to CodeClimate in the code I can just rename everything to GitLab and the use case still stands.

bendrucker:
You're hitting on the problem with re-using proprietary formats. Naming, compatibility, etc. become a problem.

https://docs.codeclimate.com/docs/building-a-code-climate-engine

Guessing this is what you were alluding to with:

> As of today, getting a code quality report requires building a Docker image that adds jq to tflint and to manually manipulate the provided json output to match the expected format

An Engine seems like the appropriate path here rather than burdening TFLint with maintenance of a proprietary output format.

wata727:
Thank you for your contribution.

Unfortunately, as Ben says, like major linters like ESLint, I think it should be supported by a Code Climate engine. If there is any inconvenience caused by it, please tell me the details. If necessary, consider integrating it into the core.

Issue #1424, opened 2022-06-24
terraform_empty_list_equality rule does not cover non-trivial expressions


sergei-ivanov:
## Introduction

Thanks for adding this rule, but unfortunately it only picks up the most trivial cases.

For example, it picks up this:
```hcl
resource "aws_security_group_rule" "outbound_proxy" {
  count =var.proxy_subnets != [] ? 1 : 0
```

but not this:
```hcl
resource "aws_security_group_rule" "outbound_proxy" {
  count = var.enabled && var.proxy_subnets != [] ? 1 : 0
```

nor this:
```hcl
templatefile("${path.module}/jenkins.yaml.tmpl", {
    JENKINS_INGRESS_ENABLED = var.jenkins_controller_alb_allow_list != []
})
```

I guess it needs to implement more complex recursion for parsing into expressions.

It may also need to check for comparison with empty maps, e.g. `var.peering_connections != {}`.

## Expected Behavior

All cases of comparison to empty list (or map) are highlighted by the rule.

bendrucker:
The first example would be worth attempting. Short term, the primary goal should be to support expressions commonly passed to attributes like `count`.

Recursive evaluation of complex attribute values is a deeper issue than this rule, e.g. see https://github.com/terraform-linters/tflint/issues/1257. Providing a helper function that visits each value expression within collection types is its own exercise with potential for use in many rules.

wata727:
Fixed by https://github.com/terraform-linters/tflint/pull/1426

Issue #1257, opened 2021-11-16
`terraform_deprecated_interpolation`: deeply inspect map/list attributes


morremeyer:
terraform_deprecated_interpolation does not detect interpolation everywhere in a configuration. Take the following two examples:

```hcl
data "terraform_remote_state" "remote_state" {
  backend = "remote"

  config = {
    organization = "Organization"
    workspaces = {
      name = "${var.environment}"
    }
  }
}
```

and 

```hcl
provider "aws" {
  region = var.region
  default_tags {
    tags = {
      Owner          = "owner"
      Application    = "application"
      Environment    = "${var.environment}"
      Provisioned-by = "terraform"
    }
  }
}
```

In both, the `"${var.environment}"` is not detected as deprecated interpolation-only variable.

Interestingly, the same applies to `terraform fmt`, which does also not detect this. This might be related as tflint uses terraform built-in as far as I’m aware.

## Version

```console
❯ tflint -v
TFLint version 0.33.0
+ ruleset.aws (0.8.0)
❯ terraform -v
Terraform v1.0.11
on darwin_amd64
```

bendrucker:
There is nothing special about either of these two block types so I have a hard time seeing why this would be. That rule attempts to walk all expressions:

https://github.com/terraform-linters/tflint/blob/a18d0944774d34ed10a86ff1c80c97e1e9e47000/tflint/runner_walk.go#L143-L186

That certainly includes walking `provider` and `data` blocks. It's conceivable that Terraform parses them differently I guess.

bendrucker:
Ok, did a bit of debugging here. The key characteristic of the examples is not their block types, it's the fact that it's an object expression. 

The rule is looking for expressions of type `*hclsyntax.TemplateWrapExpr`. That applies when an _attribute_ is set to an unnecessarily quoted expression directly. But if the attribute is assigned to an object or another collection type that contains the expression, there's no match.

I seem to recall the logic for that rule was lifted from Terraform. `terraform fmt` could do better here and making the TFLint rule more comprehensive than `fmt` is also ok, as long as the rule is faithful to the written guidance in the Terraform docs.

In this case, the expression is actually a `*hclsyntax.ObjectConsExpr`. And `expr.(*hclsyntax.ObjectConsExpr).Items[0].ValueExpr` is a `*hclsyntax.TemplateWrapExpr`. Don't have time to take this any further at the moment, but in theory it would be doable to traverse the data in known collection expression types.

[...]

PR #1426, opened 2022-07-01
`terraform_empty_list_equality`: improve expression detection

bmbferreira:
Fixes partially https://github.com/terraform-linters/tflint/issues/1424 to be able to also detect empty lists on more complex conditional expressions such as `var.enabled && var.proxy_subnets != [] ? 1 : 0`.

As @bendrucker referred [here](https://github.com/terraform-linters/tflint/issues/1424#issuecomment-1165980739), the evaluation of complex attribute values is a deeper issue (https://github.com/terraform-linters/tflint/issues/1257) and this PR does not fix that part.

### DIFF
--- a/rules/terraformrules/terraform_empty_list_equality.go
+++ b/rules/terraformrules/terraform_empty_list_equality.go
@@ -55,27 +55,33 @@ func (r *TerraformEmptyListEqualityRule) Check(runner *tflint.Runner) error {
 func (r *TerraformEmptyListEqualityRule) checkEmptyList(runner *tflint.Runner) error {
 	return runner.WalkExpressions(func(expr hcl.Expression) error {
 		if conditionalExpr, ok := expr.(*hclsyntax.ConditionalExpr); ok {
-			if binaryOpExpr, ok := conditionalExpr.Condition.(*hclsyntax.BinaryOpExpr); ok {
-				if binaryOpExpr.Op.Type.FriendlyName() == "bool" {
-					if right, ok := binaryOpExpr.RHS.(*hclsyntax.TupleConsExpr); ok {
-						checkEmptyList(right, runner, r, binaryOpExpr)
-					}
-					if left, ok := binaryOpExpr.LHS.(*hclsyntax.TupleConsExpr); ok {
-						checkEmptyList(left, runner, r, binaryOpExpr)
-					}
-				}
-			}
+			searchEmptyList(conditionalExpr.Condition, runner, r, conditionalExpr.Range())
 		}
 		return nil
 	})
 }
 
-func checkEmptyList(tupleConsExpr *hclsyntax.TupleConsExpr, runner *tflint.Runner, r *TerraformEmptyListEqualityRule, binaryOpExpr *hclsyntax.BinaryOpExpr) {
-	if len(tupleConsExpr.Exprs) == 0 {
-		runner.EmitIssue(
-			r,
-			"Comparing a collection with an empty list is invalid. To detect an empty collection, check its length.",
-			binaryOpExpr.Range(),
-		)
+func searchEmptyList(expr hcl.Expression, runner *tflint.Runner, r *TerraformEmptyListEqualityRule, exprRange hcl.Range) {
+	if binaryOpExpr, ok := expr.(*hclsyntax.BinaryOpExpr); ok {
+		if binaryOpExpr.Op.Type.FriendlyName() == "bool" {
+			searchEmptyList(binaryOpExpr.RHS, runner, r, binaryOpExpr.Range())
+			searchEmptyList(binaryOpExpr.LHS, runner, r, binaryOpExpr.Range())
+		}
+	} else if binaryOpExpr, ok := expr.(*hclsyntax.BinaryOpExpr); ok {
+		searchEmptyList(binaryOpExpr, runner, r, binaryOpExpr.Range())
+	} else if parenthesesExpr, ok := expr.(*hclsyntax.ParenthesesExpr); ok {
+		searchEmptyList(parenthesesExpr.Expression, runner, r, parenthesesExpr.Range())
+	} else if tupleConsExpr, ok := expr.(*hclsyntax.TupleConsExpr); ok {
+		if len(tupleConsExpr.Exprs) == 0 {
+			emitIssue(exprRange, runner, r)
+		}
 	}
+}
+
+func emitIssue(exprRange hcl.Range, runner *tflint.Runner, r *TerraformEmptyListEqualityRule) {
+	runner.EmitIssue(
+		r,
+		"Comparing a collection with an empty list is invalid. To detect an empty collection, check its length.",
+		exprRange,
+	)
 }
--- a/rules/terraformrules/terraform_empty_list_equality.go
+++ b/rules/terraformrules/terraform_empty_list_equality.go
@@ -58,17 +58,17 @@ func (r *TerraformEmptyListEqualityRule) checkEmptyList(runner *tflint.Runner) e
 	return runner.WalkExpressions(func(expr hcl.Expression) error {
 		if binaryOpExpr, ok := expr.(*hclsyntax.BinaryOpExpr); ok && binaryOpExpr.Op.Type == cty.Bool {
 			if tupleConsExpr, ok := binaryOpExpr.LHS.(*hclsyntax.TupleConsExpr); ok && len(tupleConsExpr.Exprs) == 0 {
-				r.emitEmptyListEqualityIssue(binaryOpExpr.Range(), runner)
+				r.emitIssue(binaryOpExpr.Range(), runner)
 			} else if tupleConsExpr, ok := binaryOpExpr.RHS.(*hclsyntax.TupleConsExpr); ok && len(tupleConsExpr.Exprs) == 0 {
-				r.emitEmptyListEqualityIssue(binaryOpExpr.Range(), runner)
+				r.emitIssue(binaryOpExpr.Range(), runner)
 			}
 		}
 		return nil
 	})
 }
 
-// emitEmptyListEqualityIssue emits issue for comparison with static empty list
-func (r *TerraformEmptyListEqualityRule) emitEmptyListEqualityIssue(exprRange hcl.Range, runner *tflint.Runner) {
+// emitIssue emits issue for comparison with static empty list
+func (r *TerraformEmptyListEqualityRule) emitIssue(exprRange hcl.Range, runner *tflint.Runner) {
 	runner.EmitIssue(
 		r,
 		"Comparing a collection with an empty list is invalid. To detect an empty collection, check its length.",
--- a/rules/terraformrules/terraform_empty_list_equality.go
+++ b/rules/terraformrules/terraform_empty_list_equality.go
@@ -6,6 +6,7 @@ import (
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/terraform-linters/tflint/tflint"
+	"github.com/zclconf/go-cty/cty"
 )
 
 // TerraformEmptyListEqualityRule checks whether is there a comparison with an empty list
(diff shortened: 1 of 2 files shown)


rules/terraformrules/terraform_empty_list_equality.go
```diff
@@ -55,27 +55,33 @@ func (r *TerraformEmptyListEqualityRule) Check(runner *tflint.Runner) error {
...
+	} else if parenthesesExpr, ok := expr.(*hclsyntax.ParenthesesExpr); ok {
+		searchEmptyList(parenthesesExpr.Expression, runner, r, parenthesesExpr.Range())
+	} else if tupleConsExpr, ok := expr.(*hclsyntax.TupleConsExpr); ok {
+		if len(tupleConsExpr.Exprs) == 0 {
+			emitIssue(exprRange, runner, r)
+		}
 	}
 }
+
+func emitIssue(exprRange hcl.Range, runner *tflint.Runner, r *TerraformEmptyListEqualityRule) {
```
bendrucker, line 81: This is not a suitable function name, note that Go is package scoped

rules/terraformrules/terraform_empty_list_equality_test.go
```diff
@@ -41,6 +41,34 @@ resource "aws_db_instance" "mysql" {
 				},
 			},
 		},
+		{
+			Name: "comparing with [] is not recommended (mixed with other conditions)",
+			Content: `
+resource "aws_db_instance" "mysql" {
+	count = true == true || false != true && (false == false || [] == []) ? 1 : 0
```
bendrucker, line 48: This test is unclear and seems to be trying to do to much. The goal here should be to have unit tests where the case being tested is clear and unique. That likely means more tests to achieve the same coverage, but the ability for a reader to understand the test's intent is paramount.

rules/terraformrules/terraform_empty_list_equality_test.go
```diff
@@ -41,6 +41,34 @@ resource "aws_db_instance" "mysql" {
...
+	instance_class = "m4.2xlarge"
+}`,
+			Expected: tflint.Issues{
+				{
```
bendrucker, line 52: This doesn't seem like a correct result. The expression that is invalid is the comparison, with both left, right, and a comparator. In the case of `[] == []` that is only one invalid expression and should be one issue. Emitting duplicate issues should be a hint that something's off.

bendrucker (changes requested):

rules/terraformrules/terraform_empty_list_equality_test.go
bmbferreira: separated this test in two (one to check parentheses and another one to check multiple conditions that check for empty lists)

rules/terraformrules/terraform_empty_list_equality_test.go
bmbferreira: Just used a map to store the ranges and to avoid emitting multiple issues with the same range. Now it will just report once in case we have `[] == []`

rules/terraformrules/terraform_empty_list_equality.go
bmbferreira: renamed it to `emitEmptyListEqualityIssues`. Now it also receives a map (actually, a Set) with all the issues to emit.

bmbferreira:
Hi @bendrucker , thanks for the swift review. 🙇 
Addressed all your comments, please let me know if there's anything else to fix/improve. 

Have a great weekend!

rules/terraformrules/terraform_empty_list_equality.go
```diff
@@ -10,6 +10,7 @@ import (
 
 // TerraformEmptyListEqualityRule checks whether is there a comparison with an empty list
 type TerraformEmptyListEqualityRule struct{}
+type void struct{}
```
bendrucker, line 13: Better to just use anonymous empty structs rather than inventing a `void` type like this

rules/terraformrules/terraform_empty_list_equality.go
```diff
@@ -55,27 +56,37 @@ func (r *TerraformEmptyListEqualityRule) Check(runner *tflint.Runner) error {
...
-					if right, ok := binaryOpExpr.RHS.(*hclsyntax.TupleConsExpr); ok {
-						checkEmptyList(right, runner, r, binaryOpExpr)
-					}
-					if left, ok := binaryOpExpr.LHS.(*hclsyntax.TupleConsExpr); ok {
-						checkEmptyList(left, runner, r, binaryOpExpr)
-					}
-				}
-			}
+			var issuesRangeSet = make(map[hcl.Range]void)
+			searchEmptyList(conditionalExpr.Condition, runner, r, conditionalExpr.Range(), issuesRangeSet)
```
bendrucker, line 60: Making your helper functions into methods on `*TerraformEmptyListEqualityRule` would eliminate the naming issues

rules/terraformrules/terraform_empty_list_equality_test.go
```diff
@@ -30,6 +30,25 @@ resource "aws_db_instance" "mysql" {
 						End:      hcl.Pos{Line: 3, Column: 18},
 					},
 				},
+			},
+		},
+		{
+			Name: "comparing with [] multiple times in the same statement is not recommended",
```
bendrucker, line 36: You can tighten up these names to be shorter/clearer for Go test output. For example, `"multiple empty comparisons"`.

bendrucker (commented):
Nice, getting closer!

rules/terraformrules/terraform_empty_list_equality.go
bmbferreira: made it a method but kept the name. Also added comments to all the `*TerraformEmptyListEqualityRule` methods to be more clear on what they're doing

bendrucker:
Thanks, will be away for a few days, can give this another review in about a week.

bmbferreira:
> Thanks, will be away for a few days, can give this another review in about a week.

Ok! Thank you! 🙇

wata727:
While working on another issue, added a change to walk expressions inside expressions in `WalkExpressions` https://github.com/terraform-linters/tflint/pull/1432
Perhaps this issue has also been resolved. Can you confirm it?

wata727:
I checked it again on the latest master branch, but it didn't fix the issue. This fix is still needed.

rules/terraformrules/terraform_empty_list_equality.go
```diff
@@ -52,30 +52,43 @@ func (r *TerraformEmptyListEqualityRule) Check(runner *tflint.Runner) error {
...
+	} else if binaryOpExpr, ok := expr.(*hclsyntax.BinaryOpExpr); ok {
+		r.searchEmptyList(binaryOpExpr, runner, binaryOpExpr.Range(), issuesRangeSet)
+	} else if parenthesesExpr, ok := expr.(*hclsyntax.ParenthesesExpr); ok {
+		r.searchEmptyList(parenthesesExpr.Expression, runner, parenthesesExpr.Range(), issuesRangeSet)
+	} else if tupleConsExpr, ok := expr.(*hclsyntax.TupleConsExpr); ok {
+		if len(tupleConsExpr.Exprs) == 0 {
+			issuesRangeSet[exprRange] = struct{}{}
+		}
+	}
+}
```
wata727, line 83: Perhaps this search logic can be rewritten fairly easily by #1432.

bmbferreira: Hi @wata727 ! Thanks for the suggestion and sorry for being silent in the last couple of weeks. Would you mind to ellaborate on your suggestion please? 😕 I'm calling `WalkExpressions` [here](https://github.com/terraform-linters/tflint/pull/1426/files/dd24efe8c505b32e3d15b093295eddeb85d55605#diff-c4c162621b4beba8a29758aed0555eb9d6a2e326ae4b2f537f3fad72592432c4R57) but as you noticed the latest change didn't fix the issue, therefore I think I still need to process the expression recursively here.

wata727: I think the pattern that this rule essentially wants to detect is `any == []`. In other words, it might be better to check against `BinaryOpExpr` while walking through all the expressions, rather than checking for `ConditionalExpr` first.

Imagine something like:

```go
runner.WalkExpression(func(expr hcl.Expression) error {
  if binaryOpExpr, ok := expr.(*hclsyntax.BinaryOpExpr); ok && binaryOpExpr.Op.Type == cty.Bool {
    if tupleConsExpr, ok := binaryOpExpr.RHS.(*hclsyntax.TupleConsExpr); ok && len(tupleConsExpr.Exprs) == 0 {
      runner.EmitIssue(...)
    }

    if tupleConsExpr, ok := binaryOpExpr.LHS.(*hclsyntax.TupleConsExpr); ok && len(tupleConsExpr.Exprs) == 0 {
      runner.EmitIssue(...)
    }
  }
})
```

What do you think?

bmbferreira: ahh! got it 🙇 Just tried it and it is much better now! I could even remove the map to avoid duplicated ranges 👌 thank you!

bendrucker (changes requested):
One small style change, would apply myself but I don't seem to have edit access to push to your PR branch

```diff
diff --git a/rules/terraformrules/terraform_empty_list_equality.go b/rules/terraformrules/terraform_empty_list_equality.go
index 25baf37a..83bbfd7a 100644
--- a/rules/terraformrules/terraform_empty_list_equality.go
+++ b/rules/terraformrules/terraform_empty_list_equality.go
@@ -58,17 +58,17 @@ func (r *TerraformEmptyListEqualityRule) checkEmptyList(runner *tflint.Runner) e
 	return runner.WalkExpressions(func(expr hcl.Expression) error {
 		if binaryOpExpr, ok := expr.(*hclsyntax.BinaryOpExpr); ok && binaryOpExpr.Op.Type == cty.Bool {
 			if tupleConsExpr, ok := binaryOpExpr.LHS.(*hclsyntax.TupleConsExpr); ok && len(tupleConsExpr.Exprs) == 0 {
-				r.emitEmptyListEqualityIssue(binaryOpExpr.Range(), runner)
+				r.emitIssue(binaryOpExpr.Range(), runner)
 			} else if tupleConsExpr, ok := binaryOpExpr.RHS.(*hclsyntax.TupleConsExpr); ok && len(tupleConsExpr.Exprs) == 0 {
-				r.emitEmptyListEqualityIssue(binaryOpExpr.Range(), runner)
+				r.emitIssue(binaryOpExpr.Range(), runner)
 			}
 		}
 		return nil
 	})
 }
 
-// emitEmptyListEqualityIssue emits issue for comparison with static empty list
-func (r *TerraformEmptyListEqualityRule) emitEmptyListEqualityIssue(exprRange hcl.Range, runner *tflint.Runner) {
+// emitIssue emits issue for comparison with static empty list
+func (r *TerraformEmptyListEqualityRule) emitIssue(exprRange hcl.Range, runner *tflint.Runner) {
 	runner.EmitIssue(
 		r,
 		"Comparing a collection with an empty list is invalid. To detect an empty collection, check its length.",
```

You can copy the above and pipe to `git apply`.

bmbferreira:
> One small style change, would apply myself but I don't seem to have edit access to push to your PR branch
> 
> ```diff
> diff --git a/rules/terraformrules/terraform_empty_list_equality.go b/rules/terraformrules/terraform_empty_list_equality.go
> index 25baf37a..83bbfd7a 100644
> --- a/rules/terraformrules/terraform_empty_list_equality.go
> +++ b/rules/terraformrules/terraform_empty_list_equality.go
> @@ -58,17 +58,17 @@ func (r *TerraformEmptyListEqualityRule) checkEmptyList(runner *tflint.Runner) e
>  	return runner.WalkExpressions(func(expr hcl.Expression) error {
>  		if binaryOpExpr, ok := expr.(*hclsyntax.BinaryOpExpr); ok && binaryOpExpr.Op.Type == cty.Bool {
>  			if tupleConsExpr, ok := binaryOpExpr.LHS.(*hclsyntax.TupleConsExpr); ok && len(tupleConsExpr.Exprs) == 0 {
> -				r.emitEmptyListEqualityIssue(binaryOpExpr.Range(), runner)
> +				r.emitIssue(binaryOpExpr.Range(), runner)
>  			} else if tupleConsExpr, ok := binaryOpExpr.RHS.(*hclsyntax.TupleConsExpr); ok && len(tupleConsExpr.Exprs) == 0 {
> -				r.emitEmptyListEqualityIssue(binaryOpExpr.Range(), runner)
> +				r.emitIssue(binaryOpExpr.Range(), runner)
>  			}
>  		}
>  		return nil
>  	})
>  }
>  
> -// emitEmptyListEqualityIssue emits issue for comparison with static empty list
> -func (r *TerraformEmptyListEqualityRule) emitEmptyListEqualityIssue(exprRange hcl.Range, runner *tflint.Runner) {
> +// emitIssue emits issue for comparison with static empty list
> +func (r *TerraformEmptyListEqualityRule) emitIssue(exprRange hcl.Range, runner *tflint.Runner) {
>  	runner.EmitIssue(
>  		r,
>  		"Comparing a collection with an empty list is invalid. To detect an empty collection, check its length.",
> ```
> 
> You can copy the above and pipe to `git apply`.
> 
> Otherwise LGTM

thanks again for the review @bendrucker. just applied the suggested change!

bendrucker (approved):

bendrucker:
Getting some 500s right now from GitHub for merge status. I'll revisit this and merge tomorrow.

bmbferreira:
> Getting some 500s right now from GitHub for merge status. I'll revisit this and merge tomorrow.

yup, np. Seems they're having some issues: https://www.githubstatus.com/
image

PR #1454, opened 2022-07-28
`terraform_required_providers`: fix regression with `configuration_aliases`

### LINKED ISSUE #1450: `terraform_required_providers` Error: Variables not allowed
Hi. Looks like TFLint version `0.39.0` has a bug in `terraform_required_providers` rule:

```
Failed to check `terraform_required_providers` rule; versions.tf:6,32-39: Variables not allowed; Variables may not be used here.:

Error: Variables not allowed

  on versions.tf line 6, in terraform:
   6:       configuration_aliases = [azurerm.core]

Variables may not be used here.
```

However, the rule works absolutely fine on `0.38.1`. 

I use the following command to launch TFLint:

```
docker run --entrypoint /bin/sh --rm -v /path/to/tf/code/:/data -v /path/to/.tflint.hcl:/.tflint.hcl ghcr.io/terraform-linters/tflint:v0.39.0 -c 'tflint --init --config /.tflint.hcl > /dev/null && tflint --force --config /.tflint.hcl'
```

You can reproduce the issue by creating a provider with an alias and run TFLint against it:

```
terraform {
  required_providers {
    azurerm = {
      source                = "hashicorp/azurerm"
      version               = "~> 3.0.2"
      configuration_aliases = [azurerm.core]
    }
  }
}
```

Content of .tflint.hcl:

```
config {
  plugin_dir = "~/.tflint.d/plugins"

  module = true
}

plugin "azurerm" {
  enabled = true
  version = "0.15.0"
  source  = "github.com/terraform-linters/tflint-ruleset-azurerm"
}

rule "terraform_unused_declarations" {
  enabled = true
}

rule "terraform_unused_required_providers" {
  enabled = true
}

rule "terraform_typed_variables" {
  enabled = true
}

rule "terraform_required_version" {
  enabled = true
}

rule "terraform_required_providers" {
  enabled = true
}

  enabled = true

  format = "snake_case"
}

rule "terraform_documented_variables" {
  enabled = true
}

rule "terraform_deprecated_index" {
  enabled = true
}
```

bendrucker:
Fixes a regression, described in #1450. When evaluating an entry in `required_providers`, this adds a variable (dynamic value) with the provider's name, allowing the evaluation of `configuration_aliases`. This evaluation is looser than Terraform would perform, e.g., `configuration_aliases = [aws.foo.bar]` would presumably be considered valid. References to other providers should still fail.

Closes #1450

### DIFF
--- a/rules/terraformrules/terraform_required_providers.go
+++ b/rules/terraformrules/terraform_required_providers.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 
+	"github.com/hashicorp/hcl/v2"
 	tfaddr "github.com/hashicorp/terraform-registry-address"
 	"github.com/terraform-linters/tflint-plugin-sdk/hclext"
 	sdk "github.com/terraform-linters/tflint-plugin-sdk/tflint"
@@ -126,7 +127,14 @@ func (r *TerraformRequiredProvidersRule) Check(runner *tflint.Runner) error {
 			continue
 		}
 
-		val, diags := provider.Expr.Value(nil)
+		val, diags := provider.Expr.Value(&hcl.EvalContext{
+			Variables: map[string]cty.Value{
+				// configuration_aliases can declare additional provider instances
+				// required provider "foo" could have: configuration_aliases = [foo.a, foo.b]
+				// @see https://www.terraform.io/language/modules/develop/providers#provider-aliases-within-modules
+				name: cty.DynamicVal,
+			},
+		})
 		if diags.HasErrors() {
 			return diags
 		}
--- a/rules/terraformrules/terraform_required_providers_test.go
+++ b/rules/terraformrules/terraform_required_providers_test.go
@@ -206,6 +206,26 @@ provider "template" {
 			},
 		},
 		{
+			Name: "version set with configuration_aliases",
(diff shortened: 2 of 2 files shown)


rules/terraformrules/terraform_required_providers.go
```diff
@@ -126,7 +127,14 @@ func (r *TerraformRequiredProvidersRule) Check(runner *tflint.Runner) error {
...
 
-		val, diags := provider.Expr.Value(nil)
+		val, diags := provider.Expr.Value(&hcl.EvalContext{
+			Variables: map[string]cty.Value{
+				// configuration_aliases can declare additional provider instances
+				// required provider "foo" could have: configuration_aliases = [foo.a, foo.b]
+				// @see https://www.terraform.io/language/modules/develop/providers#provider-aliases-within-modules
+				name: cty.DynamicVal,
+			},
+		})
```
wata727, line 137: Interesting. FWIW, Terraform core uses `hcl.ExprMap`, so it's probably best to do so.
https://github.com/hashicorp/terraform/blob/v1.2.6/internal/configs/provider_requirements.go#L78

wata727 (approved):
👍

PR #1466, opened 2022-08-03
workflow(homebrew): update to use personal fork

chenrui333:
adopted [from the comment ](https://github.com/Homebrew/homebrew-core/pull/106670)

### DIFF
--- a/.github/workflows/homebrew.yml
+++ b/.github/workflows/homebrew.yml
@@ -9,9 +9,10 @@ jobs:
     name: Bump Homebrew formula
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/homebrew.yml
```diff
@@ -9,9 +9,10 @@ jobs:
     name: Bump Homebrew formula
     runs-on: ubuntu-latest
     steps:
-      - uses: mislav/bump-homebrew-formula-action@v1
+      - uses: mislav/bump-homebrew-formula-action@v2
```
chenrui333, line 12: not sure why the dependabot did not pick up on this.

Issue #1328, opened 2022-03-21
`terraform_unused_required_providers` errors with implicit module passing


tmatilai:
#1225 fixed the case where a provider was explicitly passed to a module, but stated:

> It remains a lint error to declare a required provider when it is implicitly inherited by the child module, as that declaration is the child's responsibility.

But this in conflict with the [Terraform best practices](https://www.terraform.io/language/expressions/version-constraints#best-practices):

> **Terraform Core and Provider Versions**
> 
> - Reusable modules should constrain only their minimum allowed versions of Terraform and providers, such as `>= 0.12.0`. This helps avoid known incompatibilities, while allowing the user of the module flexibility to upgrade to newer versions of Terraform without altering the module.
>
> - Root modules should use a `~>` constraint to set both a lower and upper bound on versions for each provider they depend on.

I.e. the root modules should specify a (stricter) version requirement. But `tflint` complains about it.

bendrucker:
Yes, I've also run into this. Your notes/citations are accurate, this should be allowed. Short of disabling the rule entirely, it seems like walking the module tree and accumulating all required providers will be necessary.

This is only an issue when you use environment variables to configure your providers. When a `provider` block is used that will serve as a usage of the provider in the root module.

tmatilai:
> This is only an issue when you use environment variables to configure your providers.

And not all providers even need configuration. For example `null` or `random`.

PR #1477, opened 2022-08-10
`terraform_unused_required_providers`: check child modules for required_providers

bendrucker:
Currently, `terraform_unused_required_providers` has a tricky bug that conflicts with official module development recommendations, as detailed in #1328. This is a hacky fix, which, when built, correctly detects that a child module declares the required provider and that the parent is setting a more specific requirement. I'm putting this up for discussion around creating the supporting APIs so that #1328 can be solved without resorting to hacks.

Issues include:

* This relies on calling `NewModuleRunners`, which a plugin using the SDK would be unable to do. More generally, rules right now are expected to evaluate a single module and broadcast issues up, not inspect children directly.
* Rules cannot directly detect whether inspection was invoked with module mode on or off. Seems like `NewModuleRunners` returns empty when the modules aren't loaded so perhaps `len(moduleRunners) == 0 && len(moduleBlocks) > 0` is enough to indicate that module mode could have but wasn't used. This rule, if it's going to attempt to detect required providers in child modules, should presumably include a warning when those modules are unavailable.
* `TestRunner` expects all files to be in the same dir, and we'd probably want to add a separate method for testing configs across multiple modules.

### DIFF
--- a/rules/terraformrules/terraform_unused_required_providers.go
+++ b/rules/terraformrules/terraform_unused_required_providers.go
@@ -50,7 +50,45 @@ func (r *TerraformUnusedRequiredProvidersRule) Check(runner *tflint.Runner) erro
 		return diags
 	}
 
+	moduleRunners, err := tflint.NewModuleRunners(runner)
+	if err != nil {
+		return err
+	}
+
+	requiredProviders, diags := getRequiredProviders(runner)
+	if diags.HasErrors() {
+		return diags
+	}
+
+RequiredProvidersLoop:
+	for key, required := range requiredProviders {
+		if _, exists := providerRefs[required.Name]; !exists {
+			for _, runner := range moduleRunners {
+				moduleRequiredProviders, diags := getRequiredProviders(runner)
+				if diags.HasErrors() {
+					return diags
+				}
+
+				if _, exists := moduleRequiredProviders[key]; exists {
+					continue RequiredProvidersLoop
+				}
+			}
+
+			runner.EmitIssue(
+				r,
+				fmt.Sprintf("provider '%s' is declared in required_providers but not used by the module", required.Name),
+				required.Range,
+			)
+		}
+	}
+
+	return nil
+}
+
+func getRequiredProviders(runner *tflint.Runner) (hcl.Attributes, hcl.Diagnostics) {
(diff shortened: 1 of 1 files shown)


wata727:
> More generally, rules right now are expected to evaluate a single module and broadcast issues up, not inspect children directly.

Correct. We need to be able to get the complete module tree in the context of the root module to detect this issue.
There's probably a way to provide an option to get the child module with `GetModuleContent`, but the issue is that you can't always get the contents of the child module.

Another issue with extending the `GetModuleContent` API is that we don't know which child module to get.
If it returns the contents of all child modules, you may need the module name or hierarchy of the retrieved content, so it might be better to add a new API such as `GetChildModuleContent` to include this metadata in the response.

> Rules cannot directly detect whether inspection was invoked with module mode on or off. Seems like NewModuleRunners returns empty when the modules aren't loaded so perhaps len(moduleRunners) == 0 && len(moduleBlocks) > 0 is enough to indicate that module mode could have but wasn't used. This rule, if it's going to attempt to detect required providers in child modules, should presumably include a warning when those modules are unavailable.

Agreed. For rules that depend on child modules, it might be nice to have an API that returns whether module inspection is enabled.
Whether a warning should be included is a bit debatable. In my opinion, it should include a warning if the user explicitly enables this rule, and ignore false negatives if enabled by default.

When I cut out rules to the ruleset-terraform plugin, I plan to introduce a concept called "preset" (imagine something like `eslint:recommended`) to enable many rules that are currently turned off. Maybe we need to change this behavior at that time

> TestRunner expects all files to be in the same dir, and we'd probably want to add a separate method for testing configs across multiple modules.

Yes. I think the test helpers have a lot of room for improvement. The `GetModuleContent` options are already always ignored.

wata727:
The current TFLint design concept assumes that all inspections are completed within a single module, and considering such rules (requires child modules) may require a redesign of the APIs.

wata727:
> Rules cannot directly detect whether inspection was invoked with module mode on or off.

It might be a good idea to add the `Module` to the global configuration that applies to plugins.
https://github.com/terraform-linters/tflint-plugin-sdk/blob/v0.11.0/tflint/config.go#L3-L7

PR #1480, opened 2022-08-12
plugin: Add support for GetModulePath API

wata727:
See https://github.com/terraform-linters/tflint-plugin-sdk/pull/171

This PR adds support for `GetModulePath` API added in https://github.com/terraform-linters/tflint-plugin-sdk/pull/171.


yashg-ti:
To be able to use this api in tflint-ruleset's check function, what's the required way of using the tflint command on the root file? I'm assuming it's some recursive approach of progressing through the terraform tree. If so how to make this recursion happen?

I'm using the iterative approach currently, where I run the tflint command individually on each child module, thinking it picks the path up from the module path parameter in the command, which is kept relative to the root file directory. But that doesn't seem to be the case as it's not working, and returning 'nil' for child modules.

PR #1482, opened 2022-08-12
formatter(compact): print `hcl.Diagnostics` errors in compact format

### LINKED ISSUE #1461: Add compact formatting for errors
While issues have a compact format, errors currently do not and are pretty printed normally. For parser errors with an associated range, we could probably print them in a compact format with the message only. For other errors, the existing behavior should stay and the full error string should be printed. 

See #1460

bendrucker:
When `--formatter compact` is set and an error occurs that is/wraps `hcl.Diagnostics`, print each diagnostic in a format similar to the format used for issues raised by rules. In an environment like GitHub Actions where output is matched and associated with lines in a PR, this could help show syntax errors that prevented TFLint's execution in the context that triggered them.

The JSON and Sarif formatters similarly handle `hcl.Diagnostics` and while there's not much boilerplate more code could definitely be shared among the formatters in the future.

Closes #1461

### DIFF
--- a/formatter/compact.go
+++ b/formatter/compact.go
@@ -1,8 +1,10 @@
 package formatter
 
 import (
+	"errors"
 	"fmt"
 
+	hcl "github.com/hashicorp/hcl/v2"
 	"github.com/terraform-linters/tflint/tflint"
 )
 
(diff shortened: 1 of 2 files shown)


wata727 (approved):
Looks good to me.

My only concern is that this syntax error will look like an issue. In my experience, the next user reaction is "Why this warning cannot be suppressed with the `tflint-ignore` annotation?".

Similar to issues, there is a way to support ignoring `hcl.Diagnostics` by annotations, but it still makes more sense to treat `hcl.Diagnostics` as issues.

Still, I understand that this doesn't make things much worse, as the existing friendly error display is already misleading users. If this behavior misleads many users, it's better to return a simple fatal error.

bendrucker:
Good point, unlike Sarif/JSON it's fairly hard to tell when a compact-formatted line originates from a rule (issue) versus an error.

PR #1531, opened 2022-10-03
chore: improve installation script

qinjin:
Verify checksum via `sha256sum` by default but can be skipped when`SKIP_VERIFY_CHECKSUM` is set to `yes` or `y`.

This provides flexibility to verify the checksum while not enforced and failed if the sha256sum doesn't exist.

### DIFF
--- a/install_linux.sh
+++ b/install_linux.sh
@@ -41,8 +41,9 @@ get_latest_release() {
 }
 
 download_path=$(mktemp -d -t tflint.XXXXXXXXXX)
-download_zip="${download_path}/tflint.zip"
+download_zip="${download_path}/tflint_${os}.zip"
 download_executable="${download_path}/tflint"
+checksum_file="${download_path}/checksums.txt"
 
 if [ -z "${TFLINT_VERSION}" ] || [ "${TFLINT_VERSION}" == "latest" ]; then
   echo "Looking up the latest version ..."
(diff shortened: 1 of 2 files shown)


bendrucker (commented):
Is there a reason you think this shouldn't be enabled by default, with the option to disable it? Also, if verification fails and there's an option to disable it, the resulting error message should inform the user.

qinjin:
Thank you @bendrucker! The rational behind is to provide verification while still keep flexibilities but I see your suggestion sounds good! So that the PR has been updated to:
* Verify checksum by default and can be disabled by `SKIP_VERIFY_CHECKSUM` env variable.
* Providing a useful message either when sha256sum not found or when checksum verification failed.
* The verification is performed before installation and cleanup dir when failed.

I see the container test is failed, will look into fix that later.

qinjin:
@bendrucker Could you help to take another look? Thanks!

qinjin:
@bendrucker Any chance to take a look at this? Thank you! 🙏

install_linux.sh
```diff
@@ -64,6 +65,30 @@ fi
...
+if [ "${SKIP_VERIFY_CHECKSUM}" != "y" ] && [ "${SKIP_VERIFY_CHECKSUM}" != "yes" ]; then
+  echo "Verifying checksum with sha256sum..."
+  set +e
+
+  cd "${download_path}"
+  curl -s -L -o "${checksum_file}" "https://github.com/terraform-linters/tflint/releases/download/${version}/checksums.txt"
+  sha256sum --ignore-missing -c "${checksum_file}" >/dev/null 2>&1
+
+  retVal=$?
+  if [ $retVal -eq 127 ]; then
```
bendrucker, line 78: Use `command -v` to detect the existence of a program rather than exit code 127

bendrucker: I'm also not sure about the logic of implicitly skipping checksum validation when `sha256sum` isn't present. If that's a concern, I'd be inclined to leave checksumming disabled by default and tell users that if they enable it they must provide `sha256sum` to enable checksum verification.

install_linux.sh
```diff
@@ -64,6 +65,30 @@ fi
...
+    echo "  verify: sha256sum not found, ignore verifying checksum and continue installation!"
+  else
+    if [ $retVal -ne 0 ]; then
+      echo "  verify: sha256sum verification failed with error code $retVal, exit installation!"
+      rm -rf "${download_path}"
```
bendrucker, line 83: If needed, move the existing cleanup logic to a `trap` than runs on exit rather than duplicating it

wata727:
@qinjin Any updates? Still interested in completing this?

wata727:
Closing this PR due to no recent activity.

PR #1546, opened 2022-10-12
Convert bug report issue template to form

bendrucker:
This PR converts TFLint's bug report issue template into an [issue form](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms). I also plan to convert the feature template in a later PR.

I've been noticing a lot of low-effort bug reports that skip over critical details that would be needed to diagnose and solve the issue. Most often, this missing information is Terraform configuration in a large/complex module that the submitter doesn't want to hassle with isolating. Understandable, but an incomplete bug report does no one any good.

In addition to structuring the existing template as a form, I've made the following additions/changes:

* Users are encouraged to report ruleset-specific issues in their corresponding repository.
* The summary is expected to be plain, concise text, whereas the _Introduction_ section of the template called for an HCL code block, without explanation of what that content should be.
* Users are required to specify the _Command_ they used. While required, the value `tflint` is populated by default.
* Users are required to specify the _Terraform Configuration_. In the very rare case that the issue is not related to the root module (e.g., plugin downloading), users can enter a placeholder value. I'll either add a note about this to the description or consider making another form specifically for plugin issues.
* Users are required to specify the _TFLint Configuration_.
* Users are required to specify the _Output_.
* Users are required to specify the _TFLint Version_. For now, this is a simple `input`.
* Users are allowed to specify the _Terraform Version_, but are only encouraged to do so for cases where Terraform can parse the config but TFLint cannot.
* Users are asked to specify the operating systems where they've reproduced the problem. Previously this would have come from `terraform -v` which prints `$GOOS_$GOARCH` as part of its output. We could add this to `tflint -v` and go back to a `textarea` for that and remove this field entirely. I also internally debated whether this should be a `dropdown` or `checkboxes`. It's common to use macOS locally and Linux on the server so I went with checkboxes. 

Syntax for issue forms:

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#input


wata727 (approved):
Nice 👍

Issue #1565, opened 2022-10-24
Error: Circular reference found


sagar2610:
## Introduction

After installing the new version, 0.42.0, I start getting circular dependency errors in my local variables.

**locals.tf**
```hcl

locals {
  resource_prefix            = "${var.project}-${var.usecase}"
  resource_prefix_underscore = "${var.project}_${var.usecase}"
}
```
**test.tf**
```

locals {
  pilot_ecrs = [
    "${local.resource_prefix}-main",
    "${local.resource_prefix}-tests"
  ]
}

module "prereq" {
  source               = "../../test"
  tags                 = local.default_tags
  resource_prefix      = local.pilot_resource_prefix
  kms_key_arns         = local.kmskey_arns
  ecr_repos            = local.pilot_ecrs
}
./infrastructure/project
Failed to prepare rule checking; failed to eval an expression in prerequisites.tf:27; prerequisites.tf:13,8-29: circular reference found; local.ecrs -> local.resource_prefix -> local.resource_prefix:

Error: circular reference found

  on prerequisites.tf line 13, in locals:
  13:     "${local.resource_prefix}-tests"

local.ecrs -> local.resource_prefix -> local.resource_prefix
```
-->

miguelangelmorenochacon:
Same problem here, I've reproduced it with

```hcl
locals {
  sample_value = "whatever"

  sample_map = merge({
    sample-key = {
      another-sample-key = local.sample_value
    }
    sample-key-2 = {
      another-sample-key = local.sample_value
    }
  })
}

resource "null_resource" "sample" {
  for_each = local.sample_map
}
```

wata727:
Thank you for reporting this issue. I'm working on a fix for this bug.

wata727:
v0.42.1 has been released.
https://github.com/terraform-linters/tflint/releases/tag/v0.42.1

PR #1566, opened 2022-10-24
Fix incorrect circular reference detection

wata727:
Fixes https://github.com/terraform-linters/tflint/issues/1565

Circular reference detection released in v0.42.0 is broken. The detection logic builds a call stack and determines that it is circular-referenced if the same address appears multiple times.

This call stack should literally be the stack. It should be popped after every push instead of clearing it at the end. The current detection logic detects erroneous circular references when the same local value appears multiple times in the middle.

```mermaid
graph LR
  local.foo ---> local.bar
  local.bar ---> local.baz
  local.bar ---> local.baz
```

This PR fixes the detection logic by using the stack correctly.

### DIFF
--- a/terraform/evaluator.go
+++ b/terraform/evaluator.go
@@ -20,22 +20,22 @@ type ContextMeta struct {
 	OriginalWorkingDir string
 }
 
-type CallGraph struct {
-	edges map[string]addrs.Reference
+type CallStack struct {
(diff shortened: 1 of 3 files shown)


Shr3ps:
We have this too in multiple pipelines failing since today with this version.

Issue #1351, opened 2022-04-11
Keyless signing


wata727:
I'm a fan of [keyless signing](https://blog.chainguard.dev/zero-friction-keyless-signing-with-kubernetes/) ideas. In TFLint, we distribute signatures and a public key to verify that the release was created by us. This private key only exists in my local machine, so an attacker cannot replace it with a malicious binary unless my computer is compromised.

On the other hand, It's not good that only I can sign. Ideally, committers should always be able to create releases with valid signatures. However, sharing a private key has long been a headache for me because it makes key management problems.

Keyless signing solves this problem. If you can use the [GitHub Actions OIDC provider](https://github.blog/changelog/2021-10-27-github-actions-secure-cloud-deployments-with-openid-connect/) and confirm that the token is issued from the appropriate repository, you can verify that the release is created by us. Certificates are stored in Rekor, so you don't even have to think about managing public keys

This idea could also be used to verify plugins. We are currently [embedding the public key](https://github.com/terraform-linters/tflint/blob/v0.35.0/plugin/signature.go#L62-L162) in TFLint for major ruleset plugins such as tflint-ruleset-aws to ensure that the plugin is distributed from the appropriate repository. For other third parties, you need to put a long `signing_key` in the config file. With keyless signing, users can use signed and secure binaries without any additional configuration.

However, the problem with this way is that it is vulnerable if an attacker gets the namespace after the owner renames the repository. An attacker-signed release in this way is indistinguishable from an owner-signed release because the OIDC subject claims only contain the repository name.

A mitigation for this problem is to generate a file that stores the checksum, such as `go.sum`. I believe that this will allow us to adopt keyless signing.

mitchellrj:
Could you provide an example of how to verify the signature? Thanks

wata727:
GitHub recently introduced Artifact Attestations.
https://github.blog/2024-05-02-introducing-artifact-attestations-now-in-public-beta/

It may be a good idea to introduce the [attest-build-provenance action](https://github.com/actions/attest-build-provenance) into the plugin repository and implement a verifier in TFLint like the GitHub CLI.
https://github.com/cli/cli/blob/v2.49.0/pkg/cmd/attestation/verify/verify.go

PR #1587, opened 2022-11-13
Deprecate GPG key signed signature

wata727:
See https://github.com/terraform-linters/tflint/issues/1351

In preparation for the move to keyless signing, we are deprecating the signature verification with GPG keys. This PR removes the section about the GPG key from the README and encourages users to move to [cosign](https://github.com/sigstore/cosign).

The public key and signatures included in releases have not yet been removed. They will be supported until May 1, 2023, which is the key expiration date. After the expiration date, the key will not be updated and signatures by the GPG key will no longer be included in releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,18 +40,10 @@ NOTE: The Chocolatey package is NOT directly maintained by the TFLint maintainer
 
 ### Verification
 
-GnuPG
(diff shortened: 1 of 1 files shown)


wata727:
According to https://github.com/terraform-linters/tflint/pull/1588, keysless signing is still experimental. Maybe we shouldn't deprecate verification with GPG keys right now...

wata727:
Cosign 2.0 is out. Keyless signing is now GA 🎉 
https://blog.sigstore.dev/cosign-2-0-released

Issue #1508, opened 2022-09-10
Failed to load modules from outside of the module's directory


wata727:
## Introduction

Enabling module inspection and inspecting from outside the module's directory will cause the module to fail to load.

```console
$ tflint --module ./work
Failed to load configurations; work/main.tf:1,1-18: `instance` module is not found. Did you run `terraform init`?; :

Error: `instance` module is not found. Did you run `terraform init`?

  on work/main.tf line 1, in module "instance":
   1: module "instance" {
```

```console
$ cd work; tflint --module
# No errors
```

See also https://github.com/terraform-linters/tflint/discussions/1435

## Expected Behavior

It loads successfully with no errors.

```console
$ tflint --module ./work
# No errors
```

## Actual behavior

It failed to load.

## Step to Reproduce

```console
$ mkdir work
$ cat <<EOS > work/main.tf
module "instance" {
  source = "./module"
}
EOS
$ mkdir work/module
$ cat <<EOS > work/module/main.tf
resource "aws_instance" "main" {
  instance_type = "tx.micro"
}
EOS
$ tree work
work
├── main.tf
└── module
    └── main.tf

1 directory, 2 files
```

```console
$ cd work
$ terraform get
- instance in module
$ cd ../
$ tflint --module ./work
```

## Additional Context

```console
$ tflint -v
TFLint version 0.40.0
+ ruleset.azurerm (0.18.0)
$ terraform -v
Terraform v1.2.6
on linux_amd64
```

bendrucker:
#1315 (`-chdir`) would prevent this edge case. I can make some time to work on that soon. The ability to pass multiple modules and partial modules (single files) has been responsible for a lot of bugs and unexpected behavior.

wata727:
Thank you for the comment, but I'm wondering if the `-chdir` should be implemented.
Previously, TFLint used Terraform's native loader, so the current directory was important for module loading. I think the `-chdir` should be implemented in this design.

However, the current directory is not necessarily important for module loading, as there is now room for improving the module loader to TFLint's use case. I think the use case where `-chdir` is needed is often to inspect multiple modules. In that case, I believe the most appropriate solution is a recursive inspection rather than `-chdir`.

I think extending the module loader to allow modules to be loaded without depending on the current directory would be a good solution to this issue.

bendrucker:
[...]

Issue #1315, opened 2022-02-14
Implement `-chdir` option to mimic terraform behaviour


sergei-ivanov:
`tflint --help` says:
```
Usage:
  tflint [OPTIONS] [FILE or DIR...]
```
...and I assumed that if I pass a directory to it, that directory will be considered root for scanning. But apparently it's not the case, the root is still the current directory, which means I always need to `cd` into the target directory before running `tflint`.

I guess we need a separate option like `-chdir` in Terraform, as was also mentioned [here](https://github.com/terraform-linters/tflint/issues/527#issuecomment-866134884). It would also allow scripting recursive checks without the need for `cd` or `pushd`/`popd` for every module.

bendrucker:
Of note, Terraform handles this option different than all the rest of its command options. While command options follow the command (e.g., `terraform apply -target`), `-chdir` modifies all commands and comes first (`terraform -chdir=./foo apply`). IIRC Terraform does some custom flag parsing to achieve this and we'd probably borrow that code.

bendrucker:
See #1580 for an instance of this difference in behavior tripping up users. In Terraform, `file("./foo")` works when you call `terraform -chdir=/my/module apply`. In TFLint, where the working directory is not actually changed, this call fails.

Issue #1355, opened 2022-04-14
Recursive inspection (again)


wata727:
See also https://github.com/terraform-linters/tflint/issues/527, https://github.com/terraform-linters/tflint/pull/841, https://github.com/terraform-linters/tflint/issues/1315

Recursive inspection for Terraform modules is a feature requested by many users. So far, I have thought that TFLint cannot support it for the following reasons:

- Some expressions depend on the execution directory and the root module (e.g. `path.cwd`, `terraform.workspace`)
- TFLint uses Terraform's internal API to load modules, which does not support recursive loading

However, it is bad to need to write shell scripts to perform inspections. To make matters worse, some of the examples I've seen haven't been able to inspect the module correctly. Proper implementation of recursive inspection requires a high degree of knowledge of Terraform and TFLint, which is too hard for end-users.

Fortunately, the second limitation has been mitigated by the impact of Terraform's package internalization. We are already working to reduce the dependency on Terraform's internal API, and in the future the module loader can be completely rewritten for TFLint.

The first limitation is a difficult one, but it may be worth exploring the possibility of heuristic evaluation rather than a complete reproduction of Terraform semantics.

andredesousa:
I'm using the next command to check my modules:
`tflint modules/**`
and I see a result like:

1 issue(s) found:

modules/kubernetes/main.tf:32:18: Error - "Standard_D2_v2sdsds" is an invalid value as vm_size (azurerm_kubernetes_cluster_default_node_pool_invalid_vm_size)

wata727:
I'm now looking into other tools that have the same issue to achieve this feature.

## tfsec

https://github.com/aquasecurity/tfsec

tfsec is a security-focused linter for Terraform, similar to TFLint. tfsec recursively searches for modules by default. Check whether it is a module by whether it contains a tf file.
https://github.com/aquasecurity/defsec/blob/v0.73.0/pkg/scanners/terraform/scanner.go#L279-L325

Since tfsec focuses on obvious security errors, it may be less complicated than TFLint. In the case of TFLint, the rules you want to apply to the root module and child modules may be different, so it would be difficult to simply adopt this policy.

## Terragrunt

https://github.com/gruntwork-io/terragrunt

Terragrunt is a wrapper made to work with multiple Terraform modules. It's not a linter, but it would be helpful to see how it provides an interface to configure multiple modules with different contexts.

[...]

PR #1612, opened 2022-12-03
Add --chdir option

wata727:
Fixes https://github.com/terraform-linters/tflint/issues/1315
Fixes https://github.com/terraform-linters/tflint/issues/1508
See also https://github.com/hashicorp/terraform/pull/26087

This PR adds `--chdir` option for inspection. This is intended to match Terraform's behavior regarding other directories.

You can inspect against a Terraform configuration in another directory like this:

```console
$ tflint --chdir=environment/production
```

Similar behavior could be reproduced by passing the directory as an argument, but `--chdir` differs in the following points:

- Load a module manifest file under the directory passed by `--chdir`, not the current directory.
  - This resolves https://github.com/terraform-linters/tflint/issues/1508.
  - It means `terraform -chdir=dir get; tflint --chdir=dir --module` works fine.
- `file("./config.tmpl")` is resolved against the directory passed by `--chdir`, not the current directory.
  - This is the same behavior as Terraform and is preferable in many cases.

For the above reasons, we recommend that users using directory arguments switch to the `--chdir` option. The directory argument still works fine at this point, but is deprecated and may be removed in a future version. To avoid confusion, using `--chdir` option and a directory argument together is an error.

~~The `--chdir` behavior is the same as [Terraform's behavior](https://developer.hashicorp.com/terraform/cli/commands#switching-working-directory-with-chdir), but there are some TFLint-specific considerations:~~

- ~~Config file (`.tflint.hcl`) is processed before acting on the `--chdir` option.~~
- ~~Files specified with relative paths like `--var-file` and `varfile` on config files are resolved against the original working directory.~~

**UPDATED:** The above exceptions were eventually removed. `--chdir` is basically equivalent to `cd dir; tflint`.

However, unlike Terraform, `--chdir` cannot be used with `tflint --init` or `tflint --langserver`. This is because I believe that `--chdir` is necessary only in limited cases.

In the future, this behavior is intended to be used for [recursive inspection](https://github.com/terraform-linters/tflint/issues/1355). Recursive inspection is equivalent to running `--chdir` for each module and returning the results.

## ToDo

- [x] Add tests
- [x] Add E2E tests
- [x] Add documentation
- [x] Define behavior when `--chdir` and directory argument are passed
- [x] Return an error if `--chdir` is passed outside of inspection


wata727:
@bendrucker Could you give me your opinion on the behavior?

bendrucker (commented):
I'd expect this behavior to work exactly like `terraform -chdir`, meaning the `chdir` call happens before anything else, and is equivalent to `cd dir/ && tflint`. I'm assuming the difference comes down to the intent to use the internals in the future for recursive inspection. But I think that implementation detail should be hidden and the flag should match Terraform's behavior.

As is, users can use a shared plugin dir (which is the default?) and an absolute path to config for recursion:


```sh
config=$(realpath .tflint.hcl)
for mod in mod1 mod2; do
  tflint --chdir $mod --config $config
done
```

Not clear to me why there should be exceptions for varfiles either. If you want to share, you can use a similar pattern. Future recursive inspection could essentially do this internally, passing an absolute path for the config and var files down so that the directory can be changed. These files don't contain any relative paths of their own. 

I think we should be preparing to move `init` and `languageserver` to commands, with a compatibility later to to transform the existing args format. Probably `version` and `as-plugin` as well. These aren't options that modify behavior in the traditional sense. They do something so substantially different that they deserve their own sub-command. Ideally `--chdir` should support all commands, including `init`.

In a world with subcommands, `--chdir` could exactly match Terraform, which requires global flags to be passed before subcommands:

```sh
terraform -chdir=mod1 init
terraform -chdir=mod1 validate -json
```

```sh
tflint --chdir=mod1 init
tflint --chdir=mod1 inspect --format json
```

Happy to do a pass on the code as well once we discuss the behavior.

wata727:
Thank you for the feedback!

Agreed that the flag should match Terraform's behavior. As you say, one of the main reasons for the first exception is for recursive inspection. You might find this behavior counter-intuitive for the `-chdir` option, but I believe this makes sense since [the `terraform.rc` file already ignores `-chdir`](https://developer.hashicorp.com/terraform/cli/commands#switching-working-directory-with-chdir).

For the second exception, I didn't want the following case to be an error:

```console
$ tflint --chdir=dir --var-file=dir/varfile.tfvars
```

When entering the path for `--var-file`, I think it is intuitive to specify the path relative to the original working directory.

> I think we should be preparing to move init and languageserver to commands, with a compatibility later to to transform the existing args format. Probably version and as-plugin as well.

Yes. I think so too. However, I would like `tflint` without subcommands to behave as same as `tflint inspect`. I think it is necessary to investigate whether this can be satisfied.

bendrucker:
I think we'll manage to find ways to preserve backwards compatibility for arguments. The "command not found" handler could call `inspect` with the given args, for example. 

Re `.terraformrc`, I assume you're referring to this:

> Settings in the [CLI Configuration](https://developer.hashicorp.com/terraform/cli/config/config-file) are not for a specific subcommand and Terraform processes them before acting on the -chdir option.

Typically, this file is in `$HOME` and shouldn't contain any relative paths. I'm not even sure whether relative paths would be handled relative to the config file or the cwd. 

Seems like the only time this has any visible impact on normal use-cases is this:

> If a terraform.d/plugins directory exists in the current working directory then Terraform will also include that directory, regardless of your operating system. This behavior changes when you use the -chdir option with the init command. In that case, Terraform checks for the terraform.d/plugins directory in the launch directory and not in the directory you specified with -chdir.

https://developer.hashicorp.com/terraform/cli/config/config-file#provider_installation

In keeping with Terraform behavior, I'd expect var files to be loaded after changing directory, since var files are generally part of the configuration.

Given the following file tree:

```
.
└── mod
    ├── main.tf
    └── vars.tfvars
```

This is the behavior:

```sh
$ terraform -chdir=mod apply -var-file=vars.tfvars

No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

/var/folders/m6/7nt7y_s54lsbrwsjb_xsmx7m0000gn/T/varfile
$ terraform -chdir=mod apply -var-file=mod/vars.tfvars
╷
│ Error: Failed to read variables file
│
│ Given variables file mod/vars.tfvars does not exist.
╵
```

wata727:
Finally, I decided to load the config file in the changed directory. Fixed https://github.com/terraform-linters/tflint/pull/1612/commits/a2b6654abf8896e38652ffe1eb8c669bdc36407e

I've tried thinking about why it loads the config file in the original working directory, but I haven't found a reasonable reason. The plugin process working directory issue will be resolved by allowing plugins to refer to the original working directory.

If there are no problems, I plan to merge this PR within a few days.

bendrucker:
Will review the latest comments and give this a try today

Issue #1536, opened 2022-10-05
Git config like .tflint.hcl files combination


pjroth:
## Introduction

I have a git repo with lots of directories that represent my deployed infrastructure.  I'm using the [Gruntwork](https://github.com/gruntwork-io) infrastructure-live setup as can be seen in [their example](https://github.com/gruntwork-io/terragrunt-infrastructure-live-example).

In my directories that start with `/dev` which represent my dev infrastructure.  I want to allow `main` as a target for terraform module `source` lines so that I can deploy to DEV work that is not yet tagged/released such as:
```
module "my-app" {
  source                     = "git::ssh://git@bitbucket.org/my-org/my-repo.git//terraform?ref=main"

  // etc
}
```

In my qa and prod directories, I want to allow **only** released version tags and NOT allow `main` such as:
```
module "my-app" {
  source                     = "git::ssh://git@bitbucket.org/my-org/my-repo.git//terraform?ref=1.2.3"

  // etc
}
```

This means I need to configure tflint differently depending on which directory tree I'm under.  Specifically I want to disable the rule `terraform_module_pinned_source` for my `dev` environment but enable it for my qa and prod environments.  

## Proposal

Allow `.tflint.hcl` files to "combine" if multiple files exist in defined locations.  This would happen by looking "up" directories until the root is reached combining them into one where the more specific directory takes precedence.  This would allow me to put "common" config into my home directory the root of `/dev` to disable the rules I need to disable for my dev files 

A slightly different solution is to not look in all directories, but define some locations, perhaps GLOBAL -> HOME DIR -> current directory.  This solution is inspired withs how [git config files](https://git-scm.com/docs/git-config#FILES) work.  Copying the git config solution would be very nice and consistent since many users are likely familiar with git config.

Thank you for this great tool!

## References

1. https://git-scm.com/docs/git-config#FILES

bendrucker:
I've mentioned this in the past:

https://github.com/terraform-linters/tflint/pull/841#issuecomment-658415116

Walking upward is simple/consistent.

I played with it on a branch a while back and just didn't finish. The facilities to merge configs are there.

bendrucker:
In a large repo, putting the config at the top level leads to bloat and divorces the config from the relevant context (module). In other words, if you want to customize something for one module and then later decide to delete that module, you're likely to orphan the config at the top level. 

Hierarchical configuration is still encouraged by ESLint:

[...]

PR #1622, opened 2022-12-18
Add --recursive option

wata727:
Fixes https://github.com/terraform-linters/tflint/issues/527
Fixes https://github.com/terraform-linters/tflint/issues/1355
Closes https://github.com/terraform-linters/tflint/pull/841

This PR adds `--recursive` option for inspection. This is useful in monorepo with multiple Terraform modules:

```console
$ tree
.
├── dir1
│   └── main.tf
└── dir2
    └── main.tf

2 directories, 2 files

$ tflint --recursive
2 issue(s) found:

Warning: variable "foo" is declared but not used (terraform_unused_declarations)

  on dir1/main.tf line 5:
   5: variable "foo" {

Reference: https://github.com/terraform-linters/tflint-ruleset-terraform/blob/v0.2.1/docs/rules/terraform_unused_declarations.md

Warning: variable "foo" is declared but not used (terraform_unused_declarations)

  on dir2/main.tf line 5:
   5: variable "foo" {

Reference: https://github.com/terraform-linters/tflint-ruleset-terraform/blob/v0.2.1/docs/rules/terraform_unused_declarations.md

```

In recursive mode, it walks each directory with `--chdir`. This means that it is roughly equivalent to the following shell script:

```bash
#!/bin/bash

for dir in $(find . -type d -not -path "*/.*"; do
    if ls *.tf 1> /dev/null 2>&1; then
        tflint --chdir=dir
    fi
done
```

See also https://github.com/terraform-linters/tflint/pull/1612 for behavior regarding `--chdir`. In addition, there are some operational notes:

- If you want to apply a different config for each module, you can put a `.tflint.hcl` in each module.
  - On the other hand, if you want to apply the same config, you need to pass the absolute path using something like realpath(1). e.g. `tflint --recursive --config=$(realpath .tflint.hcl)`
  - Redundant config issue may be solved by https://github.com/terraform-linters/tflint/issues/1536.
- It is possible to run module inspections recursively, but there is no way to run `terraform init` recursively.
  - In other words, to run recursive module inspection, you need to use a shell script that runs `terraform init` recursively.
- There is currently no way to run `tflint --init` recursively.
  - This will be implemented in another pull request.
- You can pass the inspection configuration with environment variables or CLI flags, but you cannot set a different configuration for each module. In this case, you should use config files.
- Hidden directories and directories that do not contain *.tf files are excluded from recursive inspection.
- The `format` and `force` in the config file are ignored in recursive mode. CLI flags are always respected.
- Directory and file arguments are invalid in recursive mode.
- Recursive inspection is performed sequentially for each module. It can be very slow for large projects.

Finally, this recursive mode is not suitable for all projects. This is an experimental feature and the behavior may change frequently in future releases. If this is unacceptable, it is safe to implement a shell script using `--chdir`.


wata727:
@bendrucker Any thoughts on this?

bendrucker:
Will review either today or tomorrow

cmd/inspect.go
```diff
@@ -15,142 +16,179 @@ import (
...
+func (cli *CLI) withinChangedDir(dir string, proc func() error) error {
+	if dir != "." {
+		err := os.Chdir(dir)
+		if err != nil {
+			return fmt.Errorf("Failed to switch to a different working directory; %w", err)
+		}
 	}
 
-	return ExitCodeOK
+	if err := proc(); err != nil {
```
bendrucker, line 176: If `proc()` fails, the working directory is never reset to its original value. It should probably be reset regardless of the inner result. Since modules are processed in series I don't think this would have any impact right now.

docs/user-guide/config.md
```diff
@@ -46,7 +48,16 @@ $ tflint --config other_config.hcl
...
+Change the output format. The following values are valid:
+
+- default
+- json
+- checkstyle
+- junit
+- compact
+- sarif
+
+Config files are ignored in recursive mode.
```
bendrucker, line 60:
```suggestion
In recursive mode (`--recursive`), this field will be ignored in configuration files and must be set via a flag.
```

docs/user-guide/config.md
```diff
@@ -68,6 +79,8 @@ Return zero exit status even if issues found. TFLint returns the following exit
 - 1: Errors occurred
 - 2: No errors occurred, but issues found
 
+Config files are ignored in recursive mode.
```
bendrucker, line 82:
```suggestion
In recursive mode (`--recursive`), this field will be ignored in configuration files and must be set via a flag.
```

bendrucker (approved):
Nice! Few small comments. 

While reviewing this, it occurred to me that TFLint could potentially run faster in large repositories by inspecting modules in parallel. Using `Chdir` will make that difficult. Not sure there's any viable alternative given that some Terraform modules depend on the working directory.

cmd/inspect.go
wata727: Makes sense. Initially, I thought that if an error occurred, it should be returned to the caller without any changes, but It's obviously counterintuitive that the directory is changed outside of `withinChangedDir`.

wata727:
Thank you for your review!

> it occurred to me that TFLint could potentially run faster in large repositories by inspecting modules in parallel. Using Chdir will make that difficult. Not sure there's any viable alternative given that some Terraform modules depend on the working directory.

Yes, this would be difficult because it doesn't allow you to assign a different working directory to each goroutine. I would like to implement it naively first, and consider countermeasures if this becomes a problem in many projects.

One possible idea is to reimplement the `terraform.Loader` and `terraform.Evaluator` to be independent of the working directory and achieve goroutine-based parallelization. There is a concern that this will diverge the implementation from the Terraform codebase and make it difficult to follow upstream.

Another idea is to spawn a new process with `exec.Command` for each module. This is the same approach as Terragrunt:
https://github.com/gruntwork-io/terragrunt/blob/v0.42.5/configstack/running_module.go#L170-L173
https://github.com/gruntwork-io/terragrunt/blob/v0.42.5/shell/run_shell_cmd.go#L69

This works well in high-CPU machines but has an issue of inconvenient aggregation of errors and results.

PR #1627, opened 2022-12-25
chores: Fix flaky CI

wata727:
A few days ago GitHub Actions started failing frequently. Although the detailed cause is unknown, it seems that the frequency of hitting the API rate limit on GitHub Actions is increasing.

In this PR, we pass `GITHUB_TOKEN` to e2e/install workflows to avoid rate-limiting. If the `GITHUB_TOKEN` is set in the install script, it will send a request with the token.

Also, there was another problem with the install script on the `hashicorp/terraform` container image. The cause of this is unknown, but it seems to be an Alpine Linux issue due to a missing curl-dev dependency. Added curl-dev and fixed the issue.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -29,3 +29,5 @@ jobs:
         cache: true
     - name: Run e2e
       run: make e2e
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


iTiamo:
What permissions does the token need to have? I am integrating tflint in my CI and noticing 401 authorization errors every now and then. I am passing a GITHUB_TOKEN with the public_repo scope. Should that be enough?

PR #1638, opened 2022-12-30
Add --filter option

wata727:
See also https://github.com/terraform-linters/tflint/pull/1636

This PR introduces the `--filter` option as an alternative to file arguments. This is the same as the file argument, except that it doesn't check for the existence of the file:

```console
$ tflint main.tf
$ tflint --filter=main.tf
```

Introduction of this option deprecates all arguments. It can be replaced as follows:

```console
$ tflint --chdir=dir                  # replacement of "tflint dir"
$ tflint --filter=main.tf             # replacement of "tflint main.tf"
$ tflint --chdir=dir --filter=main.tf # replacement of "tflint dir/main.tf"
```

Arguments are still available, but will be removed in a future version, so early migration to `--chdir`/`--filter` is recommended.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ TFLint inspects files under the current directory by default. You can change the
 ```
 $ tflint --help
 Usage:
-  tflint [OPTIONS] [FILE or DIR...]
+  tflint --chdir=DIR/--recursive [OPTIONS]
 
 Application Options:
   -v, --version                                                 Print TFLint version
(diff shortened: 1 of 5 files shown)


bendrucker:
This should probably be paired with an `--exclude` flag. And maybe should support glob patterns. Filtering out `*.tf.json` or some local convention like `*-generated.tf` seems like something users would want. It's possible to do in a shell script with `find` and repeated flags but the general idea with these flags is to avoid writing those wrapper scripts.

wata727:
Good suggestion. Agreed that glob patterns are needed.
The original goal was to deprecate arguments, so I didn't include it that far, but I'll investigate if this PR addresses it.
The `--exclude` flag would be worth considering implementing in another PR.

wata727:
Makes sense. I'm working on something else right now, so I'm happy if you could tackle implementing `--exclude`.

PR #1654, opened 2023-01-25
Add `--minimum-failure-severity` flag, sets minimum issue severity for non-zero exit

### LINKED ISSUE #1556: Allow minimum severity for non-zero exit to be set via flag
## Introduction

Users may want to generate issues at low severity levels and display them, e.g., in logs or GitHub PR annotations. They may not always want to fail when issues are reported. Currently, TFLint exits non-zero (2) when any issues are detected.

## Proposal

Allow users to set a minimum severity for exiting with an error code. For example:

```sh
tflint --min-error-severity warning
```

This would cause issues at the `info` severity level to be printed, but TFLint would still return a 0 exit code.

## References

* #1555

nicolajv:
Resolves #1556 

Add the --minimum-severity option allowing only certain severities to result in a non-zero exit code


cmd/inspect.go
```diff
@@ -92,6 +92,25 @@ func (cli *CLI) inspect(opts Options, args []string) int {
 	cli.formatter.Print(issues, nil, cli.sources)
 
 	if len(issues) > 0 && !force {
```
wata727, line 94: If you cut out the logic into a function, you can write it as follows and I think it will be easier to understand.

```suggestion
	if len(issues) > 0 && !force && exceedsMinimumSeverity(issues, opts.MinimumSeverity) {
```

integrationtest/cli/cli_test.go
```diff
@@ -196,6 +203,20 @@ func TestIntegration(t *testing.T) {
...
+			dir:     "issues_found",
+			status:  cmd.ExitCodeIssuesFound,
+			stdout:  fmt.Sprintf("%s (aws_instance_example_type)", color.New(color.Bold).Sprint("instance type is t2.micro")),
+		},
+		{
+			name:    "`--minimum-severity` option with issues and severity level warning",
+			command: "./tflint --minimum-severity=warning",
+			dir:     "issues_found",
+			status:  cmd.ExitCodeOK,
+			stdout:  fmt.Sprintf("%s (aws_instance_example_type)", color.New(color.Bold).Sprint("instance type is t2.micro")),
```
wata727, line 218: Since the `aws_instance_example_type` severity is error, it would be wrong for this to return ExitCodeOK.
You can use the following rules in the integration tests:
https://github.com/terraform-linters/tflint/tree/7b4785d2a7515d536c8faf2bb16096829414b1dd/plugin/stub-generator/sources/testing/rules

wata727 (commented):
Thank you for your contributions. Left some comments.

By the way, the naming may have room for improvement. The `--minimum-seveirty` is the same as present in tfsec, but tfsec filters the issues. This difference in behavior may confuse users.

We have several options:

1. Filter issues like tfsec
2. Rename (e.g. Ben's suggestion is `--min-failure`, any other good name would be nice)
3. Assume no confusion and adopt `--minimum-severity`

What do you think?

bendrucker:
My thought was `minimum-failure-severity` would be appropriate (as opposed to my original suggestion with `error`. I would assume `--minimum-severity` filters out issues below the specified severity from the output. Error is ambiguous (also the name of a severity) and failure is the best alternative I can up with to describe a non-zero exit code. It's fairly commonly used in high ranking pages in a Google search for `bash non-zero exit code`, maybe as much as error:

https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html
https://www.cyberciti.biz/faq/bash-get-exit-code-of-command/
https://tldp.org/LDP/abs/html/exit-status.html

cmd/inspect.go
```diff
@@ -92,6 +92,25 @@ func (cli *CLI) inspect(opts Options, args []string) int {
...
+			case "warning":
+				minSeverity = 1
+			case "error":
+				minSeverity = 2
+			default:
+				return ExitCodeIssuesFound
+			}
+
+			for _, i := range issues {
+				if i.Rule.Severity() >= minSeverity {
```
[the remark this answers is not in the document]
bendrucker: Well they are defined in order, just arguably backwards, and definitely not the way this code expects:

https://github.com/terraform-linters/tflint-plugin-sdk/blob/3721c4b92974d442bb3b0985f327b1b929a0e579/tflint/issue.go#L4-L13

Error is 0 and Notice is 2.

They're also re-declared which should be resolved:

https://github.com/terraform-linters/tflint/blob/7b4785d2a7515d536c8faf2bb16096829414b1dd/tflint/issue.go#L24-L31

cmd/inspect.go
```diff
@@ -92,6 +92,25 @@ func (cli *CLI) inspect(opts Options, args []string) int {
 	cli.formatter.Print(issues, nil, cli.sources)
 
 	if len(issues) > 0 && !force {
+		if opts.MinimumSeverity != "" {
+			var minSeverity tflint.Severity
+			switch strings.ToLower(opts.MinimumSeverity) {
+			case "warning":
+				minSeverity = 1
+			case "error":
+				minSeverity = 2
```
bendrucker, line 101: This should be handled alongside the severity type, e.g. `func NewSeverity(s string) (Severity, error)`. An error would be returned if `s` is not a recognized severity. Handling it here is too error prone, e.g. if a new severity is added someone has to find and update this code.

bendrucker: I guess with the addition of `choice` to the struct tag an update is required anyway, but it would still be good to not hardcode this. At a minimum these severity codes should be exported/referenced instead of directly hardcoding the ints.

cmd/option.go
```diff
@@ -26,6 +26,7 @@ type Options struct {
 	Recursive          bool     `long:"recursive" description:"Run command in each directory recursively"`
 	Filter             []string `long:"filter" description:"Filter issues by file names or globs" value-name:"FILE"`
 	Force              bool     `long:"force" description:"Return zero exit status even if issues found"`
+	MinimumSeverity    string   `long:"minimum-severity" description:"Sets minimum severity for exiting with a non-zero error code" value-name:"SEVERITY"`
```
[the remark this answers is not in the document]
nicolajv: Amazing, I've switched it to this

cmd/inspect.go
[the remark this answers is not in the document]
nicolajv: Ah I made a hasty assumption here, my bad! I do think it would make more sense to have them in an order from info to error, which would allow for these kinds of comparisons. For now I've implemented a function to convert them to the values I need, but it of course necessitates changing that method if you were to add new severities... But in general it looks like sweeping changes would be needed to add new severities, since there are switch statements using them all over the codebase.
I also removed the redeclared severity constant and references the SDK everywhere instead. Let me know if this is something that should be saved for another PR 🤪

cmd/inspect.go
nicolajv: Makes sense, I did this

integrationtest/cli/cli_test.go
nicolajv: Added more and improved tests 👍

cmd/inspect.go
nicolajv: Like I mentioned above I think adding a new severity would be a task requiring many changes anyway, but I do agree the logic around severity should not be in inspect.go. I the functions which handle the conversion from the string into Severity and into in32 for comparison into the issue.go file. Not sure if this is the best space for it, but that seemed to be the home of severity.
I also removed the redeclaration of severity and changed references to point to the declaration in the SDK.

nicolajv:
> Thank you for your contributions. Left some comments.
> 
> By the way, the naming may have room for improvement. The `--minimum-seveirty` is the same as present in tfsec, but tfsec filters the issues. This difference in behavior may confuse users.
> 
> We have several options:
> 
> 1. Filter issues like tfsec
> 2. Rename (e.g. Ben's suggestion is `--min-failure`, any other good name would be nice)
> 3. Assume no confusion and adopt `--minimum-severity`
> 
> What do you think?

I went with `minimum-failure-severity` for now as suggested by bendrucker. Makes sense to me to avoid the overlap with tfsec, especially as I think this feature should definitely not filter out issues; the use case this is important for is running in CI where you want to fail on a certain level of severity but continue and just print on the lower ones.

wata727 (approved):
👍

bendrucker (approved):
Nice work on all the refactoring!

sergei-ivanov:
Somehow this new flag does not seem to be working as expected in the latest release. If I only have a bunch of warnings, I get return code zero, no matter what I set `minimum-failure-severity` to. If I have any errors, I always have a non-zero return code.

bendrucker:
Can you open an issue? There are integration tests against this behavior so it's not obvious how it wouldn't be working.

PR #1658, opened 2023-01-28
corrected inspection spelling on line 16

Ankushpandey-ti:
inspection was written as inspecion on line 16 in read me


wata727 (approved):
👍

PR #1668, opened 2023-02-03
docs: document Trunk integration

sxlijin:
Document integration with Trunk Check.

---

👋 hi there! I popped in about a year ago in #1297 to fix some error handling stuff.

We've had some users pop in during that time asking about our terraform support, and we figured that adding ourselves to your documentation would be a nice way for other users to discover that they can use terraform via Trunk Check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,10 @@ docker run --rm -v $(pwd):/data -t ghcr.io/terraform-linters/tflint
 
 If you want to run on GitHub Actions, [setup-tflint](https://github.com/terraform-linters/setup-tflint) action is available.
 
+### Trunk Check
+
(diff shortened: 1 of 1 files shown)


bendrucker:
Thanks for your past contribution, but a link back to a commercial product doesn't seem appropriate

sxlijin:
Totally get it! We figured there was no harm in asking, since different projects have tended to have different stances on linking to commercial products in their docs.

bendrucker:
Thanks, more than the commercial status the issue is that we don't have an index of "caller" tools, eg pre-commit hooks, MegaLinter, etc. A section that lists out popular tools integrated with TFLint would be welcome if it's diverse/complete. For now that probably belongs in its own docs page instead of the readme.

PR #1696, opened 2023-02-21
Update annotations.md

vitaly-dt:
The comment should be above the resource

### DIFF
--- a/docs/user-guide/annotations.md
+++ b/docs/user-guide/annotations.md
@@ -3,8 +3,8 @@
 Annotation comments can disable rules on specific lines:
 
 ```hcl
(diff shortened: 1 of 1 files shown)


vitaly-dt:
Hi [wata727](https://github.com/wata727)
It wasn't working for me until I saw a thread in some forum saying to apply it to the resource, and it worked.
I was using latest terraform aws provider and the latest tflint version.

bendrucker:
Closing as this is definitely not a correct change. An issue _could_ be emitted on the resource declaration or a specific expression. In the case of the real example, `aws_instance_invalid_type`, the latter behavior is used since the `instance_type` attribute is what the rule targets, not the entire resource.

Issue #1693, opened 2023-02-19
GA: Plugin API


wata727:
## Introduction

TFLint is a Terraform static analysis framework and exposes an API for plugins to provide its functionality widely. The [SDK](https://github.com/terraform-linters/tflint-plugin-sdk) is already 3 years old since the release of v0.1.0, and looks stable enough after migrating to gRPC.

API compatibility is important for custom ruleset developers. However, the SDK's current status is experimental, and developers must be prepared for sudden breaking changes.

## Proposal

We are ending the experimental status of the TFLint plugin API and making it GA.

Plugin API compatibility is guaranteed as much as possible. Older versions of the SDK support will be dropped approximately 6 months after the next version is released. The following features assist in this gradual migration:

- https://github.com/terraform-linters/tflint-plugin-sdk/pull/203
- https://github.com/terraform-linters/tflint-plugin-sdk/pull/202

However, please note that we may break this promise due to significant feature changes. This is just one guideline.

Before we announce GA, the following issues are blockers:

- https://github.com/terraform-linters/tflint-plugin-sdk/issues/193
- https://github.com/terraform-linters/tflint-plugin-sdk/issues/196
- https://github.com/terraform-linters/tflint-plugin-sdk/issues/192
- https://github.com/terraform-linters/tflint/issues/266
- https://github.com/terraform-linters/tflint/issues/1351
- https://github.com/terraform-linters/tflint/issues/1634
- https://github.com/terraform-linters/tflint/issues/1669

## References

- https://github.com/terraform-linters/tflint-plugin-sdk

PR #1727, opened 2023-03-26
plugin: Drop support for plugin SDK v0.12

wata727:
This PR drops support for plugins built with SDK v0.12.

SDK v0.12 is a version released over 6 months ago. It is in accordance with the future deprecation plan that versions that have passed more than 6 months since the release of the next version (v0.13) will be out of the scope of support. See https://github.com/terraform-linters/tflint/issues/1693 for details.

## TODO

- [ ] Add an integration test


wata727:
The `SDKVersion` endpoint was added from v0.14, so there is no way to distinguish between v0.12 and v0.13.
This change will be made in the next version at the same time as the end of support for v0.13.

Issue #883, opened 2020-08-21
Inform users when a new version is available


rquadling:
Running 0.17 via pre-commit-terraform on a 0.13 terraform setup says that the variable validation is experimental which it isn't in terraform 0.13.

It would be useful to know have tflint inform us of a new version as it is not always easy to remember what tools work with what version.

Terraform lets you know of a new release even though you haven't update terraform. That pattern would be extremely useful.

rquadling:
Exactly what I poorly described.

Each week I run a build of a load of tools (some of which can only supply the latest build, no versioning). The pipeline then checks what's new (essentially running the `--version` check on the installed code) and reports back what has changed.

I then create tickets to do the upgrades of the codebases that require that tool. This way, I'm in control of what tools are used in the pipelines _AND_ I get notified when there's a change.

This provides a far saner approach than "just install the latest in the pipeline" or doing nothing and then being years out of date.

PR #1735, opened 2023-04-05
Add remote version check

uthark:
Fixes https://github.com/terraform-linters/tflint/issues/883


uthark:
@wata727 please take a look

cmd/version.go
```diff
@@ -81,3 +87,38 @@ func getPluginVersions(opts Options) []string {
 
 	return versions
 }
+
+// Checks GitHub releases and prints new version, if current version is outdated.
+// requires GitHub releases to follow semver.
+func (cli *CLI) printLatestReleaseVersion() {
```
bendrucker, line 93: There should be a way to disable this, e.g., for users who want to run TFLint offline. See https://developer.hashicorp.com/terraform/cli/commands#upgrade-and-security-bulletin-checks.

cmd/version.go
```diff
@@ -81,3 +87,38 @@ func getPluginVersions(opts Options) []string {
...
+	if err != nil {
+		cli.formatter.Print(tflint.Issues{}, fmt.Errorf("Failed to check updates; %w", err), map[string][]byte{})
+	}
+	latestVersion, err := version.NewSemver(*latest.Name)
+	if err != nil {
+		cli.formatter.Print(tflint.Issues{}, fmt.Errorf("Failed to parse version; %w", err), map[string][]byte{})
+	}
+	compare := tflint.Version.Compare(latestVersion)
+	if compare < 0 {
+		fmt.Fprintf(cli.outStream, "New version available: %s\n", *latest.HTMLURL)
```
bendrucker, line 104: Terraform prints, for example:

```
Your version of Terraform is out of date! The latest version
is 1.4.4. You can update by downloading from https://www.terraform.io/downloads.html
```

It would be nice to do something similar, namely:

* The new version
* The URL

You already have the latter here.

cmd/version.go
```diff
@@ -81,3 +87,38 @@ func getPluginVersions(opts Options) []string {
...
+	}
+}
+
+func getLatestVersion() (*github.RepositoryRelease, error) {
+	ghClient := github.NewClient(nil)
+	releases, _, err := ghClient.Repositories.ListReleases(context.Background(),
```
bendrucker, line 110: This should use `/releases/latest` and not list releases:

https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release

That will handle excluding prereleases and drafts. Alternatively, if we had to use list, we should be fetching smaller pages than the 30 returned by default.

bendrucker (changes requested):

wata727:
@uthark Any updates? Still interested in finishing this?

uthark:
I’ll take a look on the weekend and get back to you.
Best regards,
Oleg.

PR #1751, opened 2023-04-25
plugin: support GitHub Enterprise Server as plugin source

### LINKED ISSUE #1643: Support installing plugins from GitHub Enterprise
TFLint plugins are installed from GitHub, using its releases API. While there are no plans to support other plugin sources, we should support alternate GitHub installations (GitHub Enterprise). 

See `BaseURL`: https://pkg.go.dev/github.com/google/go-github/github#Client

Currently the code creates a new GitHub client for each installation rather than sharing across the whole command run. So it's theoretically possible to mix plugins from different hosts without major refactoring. Currently, the internal plugin config has `SourceOwner` and `SourceRepo`. It would also need `SourceHost`. That could then be used to construct the base URL for the GitHub client for that plugin. 

Debug logs that hardcode `https://api.github.com` should be amended to use `BaseURL`. 

Originally requested in https://github.com/terraform-linters/tflint/issues/1202#issuecomment-1376331049

bendrucker:
GitHub Enterprise Server is a private installation of GitHub. It provides a compatible API, but on a hostname other than `github.com`. This PR removes the requirement that plugin sources begin with `github.com` and assumes any other hostname is a GHES server hostname. If it's not, the client would likely receive a 404, which hopefully is enough "validation." But if needed, we could make a request an look for a header to explicitly fail with "this server is not GitHub."

I factored out the request logs into a logging transport to avoid the need to to reverse engineer URLs. Otherwise, `client.BaseURL.JoinPath` is the next best way to generate a URL relative to the client. 

Still thinking about how to test this. The only reasonable way would seem to be to run a mock server using `httptest` and then pass its URL, or at least the hostname part, as the SourceHost. I believe this is always `localhost` on a random port, which technically isn't expected in SourceHost, but should work in practice right now. However, given the amount of boilerplate involved in mocking out the responses from GitHub, we could potentially just merge this and let users try it, knowing that we at least have regression testing for `github.com`. 

Closes #1643

### DIFF
--- a/plugin/install.go
+++ b/plugin/install.go
@@ -149,9 +149,12 @@ func (c *InstallConfig) fetchReleaseAssets() (map[string]*github.ReleaseAsset, e
 	assets := map[string]*github.ReleaseAsset{}
 
 	ctx := context.Background()
-	client := newGitHubClient(ctx)
 
-	log.Printf("[DEBUG] Request to https://api.github.com/repos/%s/%s/releases/tags/%s", c.SourceOwner, c.SourceRepo, c.TagName())
+	client, err := newGitHubClient(ctx, c)
+	if err != nil {
+		return assets, err
+	}
+
 	release, _, err := client.Repositories.GetReleaseByTag(ctx, c.SourceOwner, c.SourceRepo, c.TagName())
 	if err != nil {
 		return assets, err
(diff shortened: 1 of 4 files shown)


wata727 (approved):
Nice 👍

wata727:
For testing, I'm a little wondering if we should use a mock server. A mock that returns complicated responses can make tests unreliable and obscure the original intent of tests.

As far as this implementation is concerned, the only difference between GitHub.com and GHES is the behavior of `newGitHubClient`. In this case, an integration test against GitHub.com might be sufficient.

bendrucker:
Yeah definitely don't want to remove the existing internet-dependent tests, maybe just have a dedicated test for GHES. I guess I can add a unit test against the GitHub client function to exercise some of the hostname logic. Otherwise right now there's no coverage against that branch.

PR #1752, opened 2023-04-26
docs: invalid example for comma separated ignore list

juljaeg:


wata727:
Thank you for opening this PR. However, I think this is a bug rather than a documentation mistake. In fact, testing has ensured that this pattern works.
https://github.com/terraform-linters/tflint/blob/d92b37c49c1bdbaed3ef597bd3a9385a62bd8f79/tflint/annotation_test.go#L119-L132

This is probably because the regexp that defines the annotation pattern does not allow spaces.
https://github.com/terraform-linters/tflint/blob/d92b37c49c1bdbaed3ef597bd3a9385a62bd8f79/tflint/annotation.go#L13

PR #1771, opened 2023-06-01
fix(install_linux): Update unzip arguments

alexjurkiewicz:
Use `-o` (overwrite) instead of `-u` (update).

Behaviourally, there shouldn't be any difference from this change.

This improves compatibility with BusyBox unzip, which only supports `-o`.


wata727 (approved):
Looks good to me. Thanks.

PR #1787, opened 2023-06-19
feat: added ignore dir

noelgass:
Added the functionality to ignore directories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,7 @@ Application Options:
       --color                                                   Enable colorized output
       --no-color                                                Disable colorized output
       --fix                                                     Fix issues automatically
+      --ignore-dir                                              Ignore dir sources
 
 Help Options:
   -h, --help                                                    Show this help message
--- a/cmd/cli.go
+++ b/cmd/cli.go
@@ -15,6 +15,7 @@ import (
 	"github.com/terraform-linters/tflint/formatter"
 	"github.com/terraform-linters/tflint/terraform"
 	"github.com/terraform-linters/tflint/tflint"
+	"golang.org/x/exp/slices"
 )
 
 // Exit codes are int values that represent an exit code for a particular error.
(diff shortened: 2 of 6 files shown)


noelgass:
Maybe you guys can push me into the right direction to fix the checks. I've no experience with go tbh.

wata727:
Thank you for working on this.

But I'm wondering if `--ignore-dir` is the right solution for your issue. I believe this probably generalizes to https://github.com/terraform-linters/tflint/issues/1536.
What kind of directories do you want to ignore with this option?

bendrucker:
Similar thought, if not recursive configuration, a more generic `--exclude` flag seems worth considering:

https://github.com/terraform-linters/tflint/pull/1638#issuecomment-1370092980

That would allow you to `--exclude one-file.tf` but also `--exclude ./many/modules/**`.

noelgass:
Hey everyone,

Let me summarize what I aim to accomplish.

While developing Terraform modules, I provide an example directory to provide usage examples for the modules. To improve convenience, I would like to exclude the examples directory from recursive scraping. This is because it does not need to adhere to the default linting settings applied to the rest of the project, such as the requirement for `outputs.tf` or `versions.tf` files.

I believe there is room for improvement in terms of naming, and we should adopt a more generic approach, as suggested by @bendrucker. For instance, Checkov has a useful flag called `--skip-path` that allows skipping specific paths (files or directories). I would like to achieve a similar functionality in this case.

Unfortunately, I am relatively new to Go and this is my first attempt at writing Go code. Therefore, I would greatly appreciate your assistance.

Cheers

Noël

wata727:
Thank you for sharing your use case. I understand what you want to achieve.
In this case, as Ben said, implementing the `--exclude` flag would be appropriate.

Perhaps #1638 will be helpful. Instead of restricting the directories to walk, just filter the issues, much like `--filter`.
If you have any questions, please feel free to ask.

noelgass:
Thanks @wata727.

@bendrucker are you yet implementing this?

If not, I might take a look into it, but due to my lack of knowledge, in terms of go, it might take some time and might not be as good as other solutions.

wata727:
Sorry. I was a little misunderstood.
The `--filter` is an option that applies to each module, so you cannot specify a specific module like `--filter=path/to/dir`. This is the same for `--exclude`, which is the inverse of `--filter`. It may not be suitable for your use case.

However, it may be inconvenient that `--filter` cannot be used across modules. It might be worth considering changing the behavior of `--filter`. Or, as originally suggested, add new flags like `--ignore-dir`, or define a way to declare inspection settings for multiple modules.

Either way, we probably need to think from the design.

PR #1860, opened 2023-09-21
feat: support linting hcl files (terragrunt)

alikhil:
Related https://github.com/terraform-linters/tflint-plugin-sdk/pull/287

### DIFF
--- a/cmd/inspect.go
+++ b/cmd/inspect.go
@@ -222,7 +222,7 @@ func (cli *CLI) setupRunners(opts Options, dir string) ([]*tflint.Runner, error)
 	}
 	annotations := map[string]tflint.Annotations{}
 	for path, file := range files {
-		if !strings.HasSuffix(path, ".tf") {
+		if !strings.HasSuffix(path, ".tf") && !strings.HasSuffix(path, ".hcl") {
 			continue
 		}
 		ants, lexDiags := tflint.NewAnnotations(path, file)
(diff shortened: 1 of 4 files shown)


bendrucker:
Was there any sort of prior discussion about the wisdom of this? TFLint is built to statically analyze Terraform configuration, not HCL in general. My gut reaction is that this isn't appropriate and it's not readily apparent how this would be used.

wata727:
Agreed with Ben. `.hcl` files have no meaning in Terraform semantics, so parsing `.hcl` files as modules results in inconsistencies.

According to https://github.com/terraform-linters/tflint-plugin-sdk/pull/287, the goal seems to be to be able to retrieve `terragrunt.hcl` with `GetFile` API. But I'm not sure if this is pointing in the right direction. Maybe we should provide an API that can receive any file as `[]byte`, not just `.tf` files.

bendrucker:
Closing based on the lack of response. TFLint is designed to load Terraform modules, not individual files. All `.tf` and `.tf.json` are loaded because they are what compose a module. 

Anything that depends on a `.hcl` extension, which does not have the same strict semantic meaning as `.tf`, is heading down the wrong path. _If_ we were to give plugins the ability to retrieve arbitrary HCL files, it should be decoupled from the module loader and fetch the specified file from disk on demand. A plugin might want to make assertions on a `.tfvars` file, for example. 

There can be no guarantees about these files. Not their extension, and not even that they will be in the same directory as the module. 

If you're interested in pursuing these subtleties we could accept these additions to the plugin API. But it's not going to be as simple as tacking on conditions like this. More likely it needs to be a new plugin API that explicitly signals intent, eg `LoadHCLFile`.

Issue #1865, opened 2023-09-26
Linux install failing


ddelange:
### Summary

`install_linux.sh` fails in Bitbucket Pipelines with an alpine image:
```console
$ apk add curl coreutils bash unzip musl-dev gcc python3-dev py3-pip
$ curl -s https://raw.githubusercontent.com/terraform-linters/tflint/master/install_linux.sh | bash
arch=amd64
os=linux_amd64
====================================================
Looking up the latest version ...
curl: (22) The requested URL returned error: 403
Downloading TFLint 
curl: (22) The requested URL returned error: 404
Downloaded successfully
====================================================
Unpacking /tmp/tflint.GgqLkXBDN1/tflint.zip ...
unzip:  cannot find or open /tmp/tflint.GgqLkXBDN1/tflint.zip, /tmp/tflint.GgqLkXBDN1/tflint.zip.zip or /tmp/tflint.GgqLkXBDN1/tflint.zip.ZIP.
Installing /tmp/tflint.GgqLkXBDN1/tflint to /usr/local/bin/ ...
install: cannot stat '/tmp/tflint.GgqLkXBDN1/tflint': No such file or directory
Failed to install tflint
$ TFLINT_VERSION=v0.48.0 curl -s https://raw.githubusercontent.com/terraform-linters/tflint/v0.48.0/install_linux.sh | bash
arch=amd64
os=linux_amd64
====================================================
Looking up the latest version ...
curl: (22) The requested URL returned error: 403
Downloading TFLint 
curl: (22) The requested URL returned error: 404
Downloaded successfully
====================================================
Unpacking /tmp/tflint.kT03zGhfah/tflint.zip ...
unzip:  cannot find or open /tmp/tflint.kT03zGhfah/tflint.zip, /tmp/tflint.kT03zGhfah/tflint.zip.zip or /tmp/tflint.kT03zGhfah/tflint.zip.ZIP.
Installing /tmp/tflint.kT03zGhfah/tflint to /usr/local/bin/ ...
install: cannot stat '/tmp/tflint.kT03zGhfah/tflint': No such file or directory
Failed to install tflint
```

In the install script, I see curl `--fail` flag:
```
 -f, --fail          Fail silently (no output at all) on HTTP errors
     --fail-early    Fail on first transfer error, do not continue
```

Which is why the script continues with `Downloaded successfully`.

Any ideas?

### Command

tflint

### Terraform Configuration

```terraform
--
```

### TFLint Configuration

```HCL
[...]

PR #1866, opened 2023-09-26
Fix linux install failing

ddelange:
Fixes #1865

PR #1869, opened 2023-09-28
Fix linux install failing

ddelange:
Fixes https://github.com/terraform-linters/tflint/issues/1865, continuation of https://github.com/terraform-linters/tflint/issues/1866

### DIFF
--- a/install_linux.sh
+++ b/install_linux.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -euxo pipefail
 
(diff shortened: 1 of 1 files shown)


bendrucker (changes requested):
This is still titled incorrectly or needs more information. The absence of `set -e` is a bug. It is not the root cause of a failure, only of a bad error. The first step should be to fix that and amend the test case I added to ensure coverage (pipe the script to `bash`). 

If you want to refactor on top of that that's fine but should be submitted as a separate PR. Don't mix in refactoring, even when warranted (`set -ou`), with bug fixes. And if it's not refactoring, a proper description should accompany the PR explaining why other code had to change. A PR should stand alone with a summary of intent and not require readers to chase through discussion in issues.

ddelange:
Fair enough 👍

PR #1870, opened 2023-09-29
install: move `-e` out of shebang

ddelange:
Ensure `install_linux.sh` still exists non-zero upon first error when the script is piped to bash (instead of run directly, which respects shebang).


bendrucker (approved):
Thanks!

PR #1892, opened 2023-10-24
feat(annotations): Allow a comment at the end of a tflint-ignore annotation

papkos:
So that it is possible to add a reason why we ignore a rule.

Inspiration from [`nolintlint` linter](https://github.com/golangci/golangci-lint/blob/master/pkg/golinters/nolintlint/README.md).

### DIFF
--- a/docs/user-guide/annotations.md
+++ b/docs/user-guide/annotations.md
@@ -31,7 +31,26 @@ It is possible to add a reason for why a rule is ignored:
 
 ```hcl
 resource "aws_instance" "foo" {
-    # tflint-ignore: aws_instance_invalid_type # A good reason
-    instance_type = "t1.2xlarge"
+    # This instance type is new and TFLint doesn't know about it yet
+    # tflint-ignore: aws_instance_invalid_type
+    instance_type = "t10.2xlarge"
+}
+```
+
+Or, on the same line:
+
+```hcl
+resource "aws_instance" "foo" {
+  # tflint-ignore: aws_instance_invalid_type # too new for TFLint
+  instance_type = "t10.2xlarge" 
+}
+```
+
+One can also use `//`-style comments
+
+```hcl
+resource "aws_instance" "foo" {
+  // tflint-ignore: aws_instance_invalid_type // too new for TFLint
+  instance_type = "t10.2xlarge" 
 }
 ```
--- a/docs/user-guide/annotations.md
+++ b/docs/user-guide/annotations.md
@@ -27,7 +27,7 @@ resource "aws_instance" "foo" {
 }
 ```
 
-It is possible to add a reason for why a rule is ignored:
+It's a good idea to add a reason for why a rule is ignored, especially temporarily:
 
 ```hcl
 resource "aws_instance" "foo" {
@@ -46,7 +46,7 @@ resource "aws_instance" "foo" {
 }
 ```
 
-One can also use `//`-style comments
+The `//` comment style is also supported, but Terraform recommends `#`.
 
 ```hcl
 resource "aws_instance" "foo" {
--- a/docs/user-guide/annotations.md
+++ b/docs/user-guide/annotations.md
@@ -26,3 +26,12 @@ resource "aws_instance" "foo" {
     instance_type = "t1.2xlarge"
 }
 ```
+
+It is possible to add a reason for why a rule is ignored:
+
+```hcl
+resource "aws_instance" "foo" {
+    # tflint-ignore: aws_instance_invalid_type # A good reason
+    instance_type = "t1.2xlarge"
+}
+```
--- a/tflint/annotation.go
+++ b/tflint/annotation.go
@@ -10,7 +10,7 @@ import (
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 )
 
-var annotationPattern = regexp.MustCompile(`tflint-ignore: ([^\n*/]+)`)
+var annotationPattern = regexp.MustCompile(`tflint-ignore: ([^\n*/#]+)`)
 
 // Annotation represents comments with special meaning in TFLint
 type Annotation struct {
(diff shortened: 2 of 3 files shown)


docs/user-guide/annotations.md
```diff
@@ -26,3 +26,12 @@ resource "aws_instance" "foo" {
...
 ```
+
+It is possible to add a reason for why a rule is ignored:
+
+```hcl
+resource "aws_instance" "foo" {
+    # tflint-ignore: aws_instance_invalid_type # A good reason
+    instance_type = "t1.2xlarge"
+}
+```
```
wata727, line 37: It would be nice to mention what types of comment formats are allowed.
In my understanding, you can add comments after the `#` and `//` at the end of the annotation comment.

Strictly speaking, comments starting with `*` or `/` are probably valid for reasons of regexp, but I would like to make it clear that this is not what we intended.

tflint/annotation.go
```diff
@@ -40,7 +40,7 @@ func NewAnnotations(path string, file *hcl.File) (Annotations, hcl.Diagnostics)
 			continue
 		}
 		ret = append(ret, Annotation{
-			Content: match[1],
+			Content: strings.TrimSpace(match[1]),
```
wata727, line 43: Note: The `TrimSpace` is called internally in `IsAffected`, so this change probably won't affect the behavior.
https://github.com/terraform-linters/tflint/blob/8eecbda45d83f75594932840ed1892a3e571ddcc/tflint/annotation.go#L59

But I think it's good practice to pre-remove extra spaces here.

wata727 (commented):
Looks good. I left a comment about the documentation.

docs/user-guide/annotations.md
bendrucker, line 37: When placing the annotation on the previous line, also breaking the comment onto its own line(s) seems preferable. But this is a nice addition for users of same-line annotations, or very short comments.

GitHub PR suggestions don't allow for back-ticked blocks inside, here's what I have in mind:


<hr>

It is possible to add a reason for why a rule is ignored:

```hcl
    # This instance type is new and TFLint doesn't know about it yet
    # tflint-ignore: aws_instance_invalid_type
    instance_type = "t10.2xlarge"
}
```

Or, on the same line:

```hcl
    instance_type = "t10.2xlarge" # tflint-ignore: aws_instance_invalid_type # too new for TFLint
}
```

papkos: > GitHub PR suggestions don't allow for back-ticked blocks inside, here's what I have in mind:

It actually does work! (Thanks @FluffyBucket for showing me!) You just need to use more than 3 backticks on the outermost block.


````suggestion
It is possible to add a reason for why a rule is ignored:

```terraform
    # This instance type is new and TFLint doesn't know about it yet
    # tflint-ignore: aws_instance_invalid_type
    instance_type = "t10.2xlarge"
}
```

Or, on the same line:

```terraform
    instance_type = "t10.2xlarge" # tflint-ignore: aws_instance_invalid_type # too new for TFLint
}
```
````

I'll add this, together with @wata727's suggestion about `//`

File: tflint/annotation.go (hunk 2 above)
Comment (papkos) [reply tflint/annotation.go:43]: If I were to remove this trimming here, the tests would need to have extra spaces at the end of the `Content:` fields, which to me is counter intuitive, and easy to miss.
I think it is better to clean the `Content` as early as we can.

You also wrote that 
> But I think it's good practice to pre-remove extra spaces here.

so it's ok to leave this change in?

---

This is how the tests would have to look without the trimming:

```patch
Index: tflint/annotation_test.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/tflint/annotation_test.go b/tflint/annotation_test.go
--- a/tflint/annotation_test.go	(revision 5f5f91283906530c0cc3a82f0d4ea9d41248ab21)
+++ b/tflint/annotation_test.go	(date 1698227877022)
@@ -33,7 +33,7 @@
 
 	expected := Annotations{
 		{
-			Content: "aws_instance_invalid_type, terraform_deprecated_syntax",
+			Content: "aws_instance_invalid_type, terraform_deprecated_syntax ",
 			Token: hclsyntax.Token{
 				Type:  hclsyntax.TokenComment,
 				Bytes: []byte("/* tflint-ignore: aws_instance_invalid_type, terraform_deprecated_syntax */"),
@@ -69,7 +69,7 @@
 			},
 		},
 		{
-			Content: "aws_instance_invalid_type",
+			Content: "aws_instance_invalid_type ",
 			Token: hclsyntax.Token{
 				Type:  hclsyntax.TokenComment,
 				Bytes: []byte("// tflint-ignore: aws_instance_invalid_type // With reason\n"),
@@ -81,7 +81,7 @@
 			},
 		},
 		{
-			Content: "aws_instance_invalid_type",
+			Content: "aws_instance_invalid_type ",
 			Token: hclsyntax.Token{
 				Type:  hclsyntax.TokenComment,
 				Bytes: []byte("# tflint-ignore: aws_instance_invalid_type # With reason\n"),
```

wata727: > so it's ok to leave this change in?

Yes!

wata727 (approved):
Nice!

Merge after @bendrucker's approval.

docs/user-guide/annotations.md
```diff
@@ -26,3 +26,31 @@ resource "aws_instance" "foo" {
     instance_type = "t1.2xlarge"
 }
 ```
+
+It is possible to add a reason for why a rule is ignored:
```
bendrucker, line 30:
```suggestion
It's a good idea to add a reason for why a rule is ignored, especially temporarily:
```

docs/user-guide/annotations.md
```diff
@@ -26,3 +26,31 @@ resource "aws_instance" "foo" {
...
+Or, on the same line:
+
+```hcl
+resource "aws_instance" "foo" {
+  # tflint-ignore: aws_instance_invalid_type # too new for TFLint
+  instance_type = "t10.2xlarge" 
+}
+```
+
+One can also use `//`-style comments
```
bendrucker, line 49:
```suggestion
The `//` comment style is also supported, but Terraform recommends `#`.
```

Issue #1502, opened 2022-09-05
Enable local module inspection by default


wata727:
## Introduction

Currently, module inspection is disabled by default. This is because it needs to run `terraform init` to build the module tree. This decision was made to lower the hurdles for using TFLint as much as possible.

But for local modules, this assumption is not true. You can access the module content without running `terraform init`. Previously, local modules were built from the module manifest file just like any other module, so it was difficult to enable only local modules by default.

## Proposal

By redesigning the Loader, the module tree can be built for local modules even without a module manifest file.

Change `buildChildModules` to move Loader's `LoadModule` logic under the terraform package.
https://github.com/terraform-linters/tflint/blob/v0.39.3/terraform/config.go#L64-L112
https://github.com/terraform-linters/tflint/blob/v0.39.3/tflint/loader.go#L222-L277

## References

https://github.com/terraform-linters/tflint/blob/v0.39.3/terraform/config.go#L64-L112
https://github.com/terraform-linters/tflint/blob/v0.39.3/tflint/loader.go#L222-L277

raymondchen625:
+1.
`--recursive` doesn't work very well in the local-module scenario. To avoid `terraform_required_providers` and `terraform_required_version` warnings, I have to put a copy of the `terraform` block in all my local modules. That's unnecessary and hard to maintain.
Because my `tflint` runs in CI and the terraform backend is remote, it's not practical to grant CI runner access to the remote backend to run `terraform init`. Local module inspection would be a nice feature.

bendrucker:
Doesn't seem like either case is dependent on this feature.

`required_providers` is required in all modules. You may choose to leave the `version` unset in some modules and inherit it from the parent. You cannot leave `source` unset, except for providers in the `hashicorp` namespace, for which `required_providers` is always technically optional.

This feature would simply act like `--recursive`, but walk the graph of `module` calls instead of a directory. So if your goal is having rules that behave differently for child modules, that's not necessarily covered here.

`terraform init -backend=false` is a thing for getting your modules. So lack of backend credentials shouldn't be stopping anyone from using module mode.

raymondchen625:
> `terraform init -backend=false` is a thing for getting your modules. So lack of backend credentials shouldn't be stopping anyone from using module mode.

Thanks for pointing out the `-backend=false` option. It seems I shouldn't use `--module` at all. Module inspection only checks the module's input variables.
[...]

PR #1918, opened 2023-11-23
Inspect local module calls by default

wata727:
Fixes https://github.com/terraform-linters/tflint/issues/1502

This PR enables local module calls inspection by default. Previously, even if it was a local module, module call inspection was not possible unless the module was obtained in advance using `terraform init` (or `terraform get`) and executed with the `--module` flag. This PR effectively improves the module walker implementation, allowing modules to be loaded without relying on module manifests.

## ToDo

- [ ] Complete the implementation
  - [ ] Detect circular references
- [ ] Add tests
- [ ] Revise documentation


wata727:
@bendrucker Could you please review this?

bendrucker:
Will do!

docs/user-guide/compatibility.md
```diff
@@ -143,7 +143,7 @@ Similar to support for meta-arguments, some rules may process a dynamic block as
 
 ## Modules
 
-Resources contained within modules are ignored by default, but when the [Module Inspection](./module-inspection.md) is enabled, the arguments of module calls are inspected.
+Local modules are inspected by default. This allows you to identify issues caused by input values as well.
```
bendrucker, line 146: This is confusing and suggests that local modules are inspected in the same way as recursive mode.
 
```suggestion
TFLint doesn't automatically inspect the content of modules themselves. However, by default, it will analyze their content in order to raise any issues that arise from attributes in module calls.
```

bendrucker (approved):

docs/user-guide/compatibility.md
wata727: Thanks!

PR #1934, opened 2023-12-09
Stop using backticks for emphasis

wata727:
Sometimes TFLint uses messages that use backticks for emphasis or quotation. e.g:

```
$ tflint --init
Installing `aws` plugin...
```

This may look natural in markdown, but the output on the console is a bit unnatural.

This PR removes backticks for such uses and makes the output message consistent.


cmd/cli.go
```diff
@@ -109,30 +109,27 @@ func (cli *CLI) Run(args []string) int {
 
 func unknownOptionHandler(option string, arg flags.SplitArgument, args []string) ([]string, error) {
 	if option == "debug" {
-		return []string{}, errors.New("`debug` option was removed in v0.8.0. Please set `TFLINT_LOG` environment variables instead")
-	}
-	if option == "fast" {
-		return []string{}, errors.New("`fast` option was removed in v0.9.0. The `aws_instance_invalid_ami` rule is already fast enough")
```
wata727, line 115: This option handler has been removed. AWS rules are already not included in TFLint core, so they don't make sense.

PR #1961, opened 2024-01-24
Added snap installation option.

tingdahl:
I've been serving tflint using a snap for about 6 months, about 400 weekly users. I can continue that (I normally update within a few days after a new upstream release), but I'm also open to merge the snap build code into the main repo if you're interested of that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,13 @@ Bash script (Linux):
 curl -s https://raw.githubusercontent.com/terraform-linters/tflint/master/install_linux.sh | bash
 ```
 
+Snap (Ubuntu and other Linux distributions with snapd installed)
+
+```console
+sudo snap install tflint
+```
+NOTE: The [tflint](https://snapcraft.io/tflint) snap is NOT directly maintained by the TFLint maintainers. The latest version is always available by manual installation.
+
 Homebrew (macOS):
 
 ```console


bendrucker (commented):
Hmm, would like @wata727's opinion here too. Chocolatey makes a bit more sense since we don't offer any 1 step installation method for Windows. Whereas for Linux there's a universal install script. 

Also, for Terraform HashiCorp supports `apt` and `yum`, but not `snap`:

https://developer.hashicorp.com/terraform/install

My inclination here is that we shouldn't open the door to various Linux package managers unless we're going to make the whole effort to support at least the set that HashiCorp does. Which I don't think is a good use of time right now.

wata727:
Thank you for your contribution. I have no particular opinion this matter.
It might make sense to add this if a lot of users ask for it, but for now I'd like to support @bendrucker's opinion.

bendrucker:
The comment just raises more questions which starts to feel like bikeshedding. All the things you said about snaps being different/better need to be explained to _users_, not just in this thread. If it's the same story as Chocolatey (manual version updates) it's not worth it. If there's zero maintenance and you want to make the case for us to do the work to have an official distribution, we can hear that out. The middle path of advertising a third party package for a niche package manager isn't worth moving forward with right now.

tingdahl:
Ben,

I understand your position. I am willing to:

1. Create a PR to integrate the snap packaging (1 directory with a yaml
file).
2. Help you to maintain the yaml file, test new releases and promote them
to the stable channel when tests are done.
3. Invite you to be have write access on snapcraft.io, where the snap is
built.

This would release every commit to the latest/edge channel for all
architectures (x64, arm64, armhf, s390 and power),  but manual work is
needed to promote tagged releases to the stable channel.

Should you not be happy with my contribution,  you can either dump the snap
or maintain it yourself.

My end goal is the snap, and how that happens is secondary.

Hence,  we can either go down that path, or you can reject the current PR.
I agree with you that the half-baked solution is not ideal.

Cheers,


Kristofer

bendrucker:
Can you give us some sense of what's entailed here? Code/config examples or links to docs, but also a PR is fine if that's easier. Adding a config file and/or automation is fine. Administrative work (promoting releases) is what I'm trying to avoid.

tingdahl:
For now,

The easiest is to send the links to:
1. The changes on my forked repo:
https://github.com/tingdahl/tflint/blob/master/snap/snapcraft.yaml . That
is the file that I would need to move into the snapcraft directory in
upstream tflint. It would naturally need to be updated with your repo and a
few other details.
2. Documentation on snaps is best found here:
https://snapcraft.io/docs/snapcraft  . Snaps are the method to install
almost all software on Ubuntu (from web-browsers to terraform to ...
everything).

I do really not want to force myself upon you. One way ahead is to revisit
the issue in 6 months. We can then see where my install-base is and
depending on the snap's popularity, we can make a decision. That will give
this some time to sink in.

Cheers,

Kristofer



Den sön 28 jan. 2024 kl 21:50 skrev Ben Drucker ***@***.***>:

wata727:
Hmm, I don't really have an opinion on adding links with an annotation of "not directly maintained by the TFLint maintainers", but I'm not positive about adding files like `snapcraft.yaml` to a repository and maintaining them.

This comes from the regret of [removing precommit hooks](https://github.com/terraform-linters/tflint/pull/1247) in the past because it could not be maintained.

wata727:
I'm sorry, but I'm closing this pull request. Once again, I sincerely thank you for your contributions in maintaining this snap package. If you have any other thoughts, please let us know.

tingdahl:
No worries.

Issue #1968, opened 2024-02-05
Panics when emitting issues on child blocks


bendrucker:
### Summary

When a ruleset emits an issue with the declaration range of a child block, TFLint will panic if that issue is emitted from a child module. The range itself covers just the block name, which TFLint interprets as an expression, and attempts to parse it for variables. That fails and causes a panic.

See https://github.com/terraform-linters/tflint-ruleset-opa/issues/85

This issue is not directly related to the OPA ruleset plugin. Given the configuration below, the emitted issue range covers `rule` (no braces). TFLint should not assume that the declaration range of a block is a valid expression, because it's not. This happens to work for labeled blocks (e.g. `resource "foo" "bar"`, but causes errors for unlabeled blocks which are just one-word identifiers and syntactically look like an expression. 

### Command

tflint

### Terraform Configuration

```terraform
# main.tf

module "m" {
  source = "./module"
}

# module/main.tf
resource "signalfx_detector" "test" {
  rule {}
}
```

### TFLint Configuration

```HCL
plugin "opa" {
  enabled = true
  version = "0.5.0"
  source  = "github.com/terraform-linters/tflint-ruleset-opa"
}
```

### Output

```Shell
panic: module/main.tf:2,3-7: Invalid reference; A reference to a resource type must be followed by at least one attribute access, specifying the resource name.

goroutine 88 [running]:
github.com/terraform-linters/tflint/tflint.listVarRefs({0x10342b058?, 0x140001349c0?})
	github.com/terraform-linters/tflint/tflint/runner.go:326 +0x150
github.com/terraform-linters/tflint/tflint.(*Runner).listModuleVars(0x1400068f7a0, {0x10342b058?, 0x140001349c0?})
	github.com/terraform-linters/tflint/tflint/runner.go:314 +0x34
github.com/terraform-linters/tflint/tflint.(*Runner).EmitIssue(0x1400068f7a0, {0x10ad35998?, 0x140001ec3c0}, {0x140001442a0, 0x2d}, {{0x1400012e600, 0x14}, {0x3, 0x3, 0x58}, ...}, ...)
	github.com/terraform-linters/tflint/tflint/runner.go:240 +0x6c
github.com/terraform-linters/tflint/plugin.(*GRPCServer).EmitIssue.func1(...)
	github.com/terraform-linters/tflint/plugin/server.go:202
github.com/terraform-linters/tflint/tflint.(*Runner).WithExpressionContext(...)
[...]

PR #1969, opened 2024-02-05
Ignore module issues that are not valid expressions

bendrucker:
When filtering emitted issues from child modules, ignore any issue covering an expression that cannot be parsed for references. Fixes a panic described in #1968.

`EmitIssue` may be called for a range that is not actually a valid Terraform expression. But TFLint just [checks](https://github.com/terraform-linters/tflint/blob/5ed807d0d4ab05fc81d9b3c9db623d3ee937ba2e/plugin/server.go#L213) whether the range's source is a valid _HCL_ expression. Something like `my_block` is a valid HCL expression, but it's not a valid expression in Terraform when it's a child block of a `resource`. Some declaration ranges (e.g., blocks with labels) are not valid HCL expressions either, but unlabeled block declarations (name, without braces) are syntactically valid expressions.

So `listVarRefs` must handle invalid expressions. Or callers must be able to specify that they are emitting an issue on a non-expression range, which would likely require a breaking change somewhere.

Closes https://github.com/terraform-linters/tflint/issues/1968
Closes https://github.com/terraform-linters/tflint-ruleset-opa/issues/85

### DIFF
--- a/plugin/server.go
+++ b/plugin/server.go
@@ -188,6 +188,9 @@ func (s *GRPCServer) EvaluateExpr(expr hcl.Expression, opts sdk.EvaluateExprOpti
 }
 
 // EmitIssue stores an issue in the server based on passed rule, message, and location.
+// It attempts to detect whether the issue range represents an expression and emits it based on that context.
+// However, some ranges may be syntactically valid but not actually represent an expression.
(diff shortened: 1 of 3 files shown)


wata727 (approved):
Thank you for working on this!
Can you please release v0.50.3 after merging this?

ericsaboia:
@bendrucker thanks for looking into this issue and fixing it so quickly! 👏

Issue #1943, opened 2023-12-17
Improve the speed of recursive inspections


wata727:
## Introduction

After merging https://github.com/terraform-linters/tflint/pull/1918, modules that called many child modules experienced a noticeable performance degradation. The company I work for has over 200 root modules with calls to over 40 child modules. Previously, recursive inspections took about 1 minute to complete, but now that local modules are being called, it now takes nearly 5 minutes.

Below are the minimal steps to reproduce this issue:

```hcl
plugin "terraform" {
    enabled = false
}
plugin "google" {
    enabled = true
    version = "0.26.0"
    source  = "github.com/terraform-linters/tflint-ruleset-google"
}
```

```hcl
module "module" {
  source = "./module"
  count  = 50
}
```

```console
$ time tflint
real    0m1.642s
user    0m1.779s
sys     0m1.753s

$ time tflint --call-module-type none
real    0m0.056s
user    0m0.037s
sys     0m0.065s
```

Although the performance degradation in this example is small, it can add up to significant slowdowns.

## Proposal

There are 3 solutions to this kind of issue.

### Improve the speed of inspection for a single module

We can improve the speed of recursive inspections by speeding up inspecting on a single module. This can mainly be achieved by parallelizing runners.

In the current design, each module call starts a new gRPC server, and the ruleset checks for each one sequentially. So if a ruleset has 100 rules and a module calls 50 child modules, the rules will be evaluated 5000 times in total. This design has the advantage that the ruleset does not need to distinguish between root and child modules, but it is too much waste in terms of performance.

This design issue has also been tracked in https://github.com/terraform-linters/tflint-plugin-sdk/issues/193, which may solve it, but in the short term processing them in parallel may improve performance. Specifically, the following implementation could potentially be performed in goroutines:

https://github.com/terraform-linters/tflint/blob/95926c29ef9e601ec9de74792181a6a0585a0eb1/cmd/inspect.go#L179-L184

However, we need to think about thread safety issues introduced by parallelism.

**EDIT:** Merged in https://github.com/terraform-linters/tflint/pull/1944

### Parallelization of recursive inspection

By performing recursive inspection in parallel, the overall inspection time is shortened. This is what we expected when we designed the recursion inspection, but the current design requires switching working directories, so parallelization with goroutines is not possible. See https://github.com/terraform-linters/tflint/pull/1622#pullrequestreview-1223633739

[...]

PR #2021, opened 2024-04-13
Recursive inspection in parallel

wata727:
Fixes https://github.com/terraform-linters/tflint/issues/1943

As explained in #1943, this PR parallelizes recursive inspection by multi-processes.

The process that runs `tflint --recursive` is called the coordinator, and it starts a worker process for each directory. The worker process executes with `--chdir` in each directory and outputs issues in the dedicated JSON format, just like a normal TFLint command. The coordinator aggregates JSON output from each process via stdout, and provides feedback to the user.

```mermaid
flowchart TD

    User -- "tflint --recursive" --> Coordinator
    Coordinator -- "tflint --act-as-worker --chdir=subdir1" --> Worker1
    Coordinator -- "tflint --act-as-worker --chdir=subdir2" --> Worker2
    Coordinator -- "tflint --act-as-worker --chdir=subdir3" --> Worker3
```

The default parallelism is the number of CPUs. This can be controlled with the `--max-workers` flag, but should not need to be changed in most cases. In my tested repository (over 200 root modules with calls to over 40 child modules), 4 cores cut the time in half.

**Before**

```console
$ time tflint --recursive
real    5m10.744s
user    5m53.837s
sys     2m36.451s
```

**After**

```console
$ time tflint --recursive
real    2m24.805s
user    6m50.139s
sys     2m33.531s
```


cmd/option.go
```diff
@@ -147,3 +150,77 @@ func (opts *Options) toConfig() *tflint.Config {
...
+
+// Return commands to be executed by worker processes in recursive inspection.
+// All possible CLI flags are delegated, but some flags are ignored because
+// the coordinator process that starts the workers is responsible.
+func (opts *Options) toWorkerCommands(workingDir string) []string {
+	commands := []string{
+		"--act-as-worker",
+		"--chdir=" + workingDir,
+		"--force", // Exit status is always ignored
+	}
```
wata727, line 162: I'm concerned about command injection here, but I don't think there's a problem with giving arbitrary arguments.
https://semgrep.dev/docs/cheat-sheets/go-command-injection/

Issue #2005, opened 2024-03-13
Allow plugins located on github.com and GHES to be installed at once as an authenticated request


wata727:
## Introduction

TFLint plugins are primarily intended to be hosted on github.com and is installed using the GitHub API.
Currently, we support authenticated requests with `GITHUB_TOKEN` for the following purposes:

- Avoid rate limits
- Install from private repositories

https://github.com/terraform-linters/tflint/blob/v0.50.3/docs/user-guide/plugins.md#avoiding-rate-limiting

Apart from this, plugins can also be hosted on GitHub Enterprise Server (GHES). See https://github.com/terraform-linters/tflint/pull/1751

The problem is that since both github.com and GHES use the same `GITHUB_TOKEN`, there is no way to install a plugin hosted on both at once as an authenticated request. See also https://github.com/terraform-linters/tflint/discussions/2004

## Proposal

Add `access_token_env` as an attribute of the `plugin` block.

```hcl
plugin "custom" {
  source  = "github.example.com/example/tflint-ruleset-example"
  version = "0.1.0"
  enabled = true

  access_token_env = "GITHUB_ENT_TOKEN"
}
```

Plugins with `access_token_env` set refer to the declared environment variable instead of `GITHUB_TOKEN` when sending authenticated requests. We'll probably need to change the code below:

https://github.com/terraform-linters/tflint/blob/9bad28062283d16a05ed1375a99c700c9058ca3c/plugin/install.go#L253-L259

Perhaps this is a minimal change, but I'm still not sure if this is the ideal interface. We may need to extend the `source` attribute to object, or new concepts like the `connection` block, or an authentication mechanism other than environment variables.

---

**EDIT:** @bendrucker suggests a way to provide per-host environment variables like TFE https://github.com/terraform-linters/tflint/issues/2005#issuecomment-2002166525. I prefer this suggestion as it is better than extending the existing syntax.

We may be able to use something like terraform-svchost to normalize hostnames.
https://github.com/hashicorp/terraform/blob/v1.5.7/internal/command/cliconfig/credentials.go#L120

## References

- https://github.com/terraform-linters/tflint/discussions/2004
- https://github.com/terraform-linters/tflint/blob/v0.50.3/docs/user-guide/plugins.md#avoiding-rate-limiting
- https://github.com/terraform-linters/tflint/blob/v0.50.3/plugin/install.go#L253-L259

bendrucker:
[...]

PR #2025, opened 2024-04-27
plugin: Add support for host-specific GitHub tokens

wata727:
Fixes https://github.com/terraform-linters/tflint/issues/2005

This PR adds support for host-specific GitHub tokens like `GITHUB_TOKEN_example_com`.

Previously, the `GITHUB_TOKEN` environment variable was supported during plugin installation for private repositories and to avoid rate limits. However, as explained in #2004 and #2005, when using GitHub.com and GHES together as plugin source hosts, there was a problem where different access tokens could not be used for each.

To solve this, we will introduce environment variables that are only used in GHES. For example, if a plugin is hosted on example.com, `GITHUB_TOKEN_example_com` will only be used when installing a plugin whose `source` is "example.com".

```hcl
# GITHUB_TOKEN will be used
plugin "foo" {
  source = "github.com/org/tflint-ruleset-foo"
}

# GITHUB_TOKEN_example_com will be used preferentially and will fall back to GITHUB_TOKEN if not set.
plugin "bar" {
  source = "example.com/org/tflint-ruleset-bar"
}
```

This allows you to install plugins from GitHub.com and GHES at the same time by setting `GITHUB_TOKEN` and `GITHUB_TOKEN_example_com`.

Please note that host names used as environment variables must be normalized in Punycode. You can also use "_" instead of "." and "__" instead of "-".

### DIFF
--- a/docs/user-guide/plugins.md
+++ b/docs/user-guide/plugins.md
@@ -74,6 +74,20 @@ To increase the rate limit, you can send an authenticated request by authenticat
 
 It's also a good idea to cache the plugin directory, as TFLint will only send requests if plugins aren't installed. The [setup-tflint action](https://github.com/terraform-linters/setup-tflint#usage) includes an example of caching in GitHub Actions.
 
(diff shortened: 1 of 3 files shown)


bendrucker (approved):
Nice!

Issue #1878, opened 2023-10-09
Add support for Terraform v1.6+


wata727:
## Introduction

Terraform v1.6 added the following features related to TFLint.

- Unknown value refinements
  - https://github.com/hashicorp/terraform/pull/33234
  - This concept has already been introduced in HCL and cty and has also been applied to TFLint, but Terraform's built-in functions have been expanded on this concept in v1.6. If a plugin checks refinements, you may not be able to get full refinements without this change.
- Testing framework
  - https://github.com/hashicorp/terraform/pull/33325
  - https://github.com/hashicorp/terraform/pull/33339
  - https://github.com/hashicorp/terraform/pull/33683
  - These updates are required for linting `.tftest.hcl` and `tftest.json` files.

## Proposal

Update [the embedded Terraform fork](https://github.com/terraform-linters/tflint/tree/v0.48.0/terraform) to v1.6.

However, this change will be affected by the licensing change to BUSL. As a result, TFLint will be bound to MPL and BUSL. See https://github.com/terraform-linters/tflint/discussions/1826
Fortunately, this update does not immediately impact (refinement is an edge case) Terraform v1.6 users, so we can postpone this decision further. At this point, I'm on the side of accepting the license change. OpenToFu could be a solution, but maintaining full compatibility without any references to Terraform's implementation would be difficult. As long as TFLint exists as a "Terraform" linter, this option may not be practical.

## References

- https://github.com/hashicorp/terraform/releases/tag/v1.6.0
- https://github.com/hashicorp/terraform/tree/v1.6.0/internal/addrs
- https://github.com/hashicorp/terraform/tree/v1.6.0/internal/configs
- https://github.com/hashicorp/terraform/tree/v1.6.0/internal/lang
- https://github.com/hashicorp/terraform/tree/v1.6.0/internal/terraform
- https://github.com/terraform-linters/tflint/issues/1778

bendrucker:
CLI compatibility isn't the issue here. It's a question of the Go packages and their public (exported) API. Those are explicitly not part of the Terraform compatibility promise. 

_Legal speculation ahead. I am not a lawyer. This is not legal advice._

US courts have held that APIs are generally excluded from copyright protections. Only implementation (source code) is protected. 

This is straightforward for closed-source products, e.g., storage offerings with an S3-compatible API. And decidedly not for "source available" licenses like BUSL. Directly copying Terraform's internals (code) for a competing product violates the license. 

[...]

PR #2030, opened 2024-04-29
terraform: Add support for Terraform v1.6/v1.7/v1.8

wata727:
Fixes https://github.com/terraform-linters/tflint/issues/1878

This PR adds support for Terraform v1.6, v1.7, and v1.8. Specifically, support for the following features will be added:

- Tighter unknown value refinements
  - In most cases, this behavior is not relevant to end users, but it is beneficial to plugin developers who use `cty.Value` directly, in that the unknowns of unknown values are smaller.
- The `nonsensitive` function now accepts a non-sensitive value
- Provider-defined functions
  - Provider-defined functions always return unknown values, except for Terraform-specific functions described below. So it won't help you discover new issues, but it will help you avoid errors.
- Terraform-specific functions (`provider::terraform::*`)
- New `issensitive` function

A lot of the changes are just pulling upstream changes, but there are some tricks when it comes to supporting provider-defined functions.

Evaluating a provider-defined function requires running the Terraform provider plugin beforehand, which is clearly beyond the "static" scope of TFLint. Therefore, we had to adopt a different evaluation strategy. The idea here is to dynamically generate an `EvalContext` based on function call references, similar to address references.

If the expression contains a call to a provider-defined function, generate a mock function to satisfy the call, ensuring that it always returns an unknown value. This allows you to avoid errors in the evaluation of any provider-defined functions without special error handling.

However, in order to adapt this idea to dynamic block expansion, we now needed a walker to collect expression references to `hcl.Body` (note that expressions inside dynamic blocks must always be evaluated beforehand). We can use `dynblock.WalkVariables` for `[]hcl.Traversal`, but there is no such thing for expressions, so we have to extend the `tfhcl` package.

Although Terraform-specific functions (`provider::terraform::*`) are provider-defined functions, they are essentially equivalent to built-in functions, so they are implemented to return valid values. However, the implementation is similar to other built-in functions, so the approach is slightly different than upstream.

---

The most important part of this change is that the `terraform` package is now licensed under BUSL 1.1. The changed files that refer to upstream licensed under BUSL have `SPDX-License-Identifier: BUSL-1.1` added to the header. As you can see, only a small number of files are currently licensed under BUSL.

However, since it includes BUSL-licensed files when generating the executable forms (release binaries), it becomes bound to both licenses. In other words, MPL-licensed files are still free to use, distribute, and fork as MPL, but release binaries may be subject to Hashicorp's BUSL restrictions.

To inform you of this caveat, this PR also includes updates regarding LISENCE files. There may be other changes that need to be made, but I'll work on those separately.

### DIFF
--- a/LICENSE-BUSL
+++ b/LICENSE-BUSL
@@ -1,0 +1,92 @@
+License text copyright (c) 2020 MariaDB Corporation Ab, All Rights Reserved.
+"Business Source License" is a trademark of MariaDB Corporation Ab.
+
+Parameters
+
+Licensor:             HashiCorp, Inc.
+Licensed Work:        Terraform Version 1.6.0 or later. The Licensed Work is (c) 2024
+                      HashiCorp, Inc.
+Additional Use Grant: You may make production use of the Licensed Work, provided
+                      Your use does not include offering the Licensed Work to third
+                      parties on a hosted or embedded basis in order to compete with 
+                      HashiCorp's paid version(s) of the Licensed Work. For purposes 
+                      of this license:
+
+                      A "competitive offering" is a Product that is offered to third
+                      parties on a paid basis, including through paid support 
+                      arrangements, that significantly overlaps with the capabilities 
+                      of HashiCorp's paid version(s) of the Licensed Work. If Your 
+                      Product is not a competitive offering when You first make it 
+                      generally available, it will not become a competitive offering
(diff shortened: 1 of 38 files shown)


wata727:
Since this change is important to the project, I would like to get approval from co-maintainer @bendrucker.
Could you take a look on this?

bendrucker (approved):

wata727:
Thanks!

PR #2038, opened 2024-05-06
release: Introduce Artifact Attestations

wata727:
See https://github.com/terraform-linters/tflint/issues/1351

GitHub recently introduced Artifact Attestations.
https://github.blog/2024-05-02-introducing-artifact-attestations-now-in-public-beta/

This PR adds a step to build Artifact Attestations in the release workflow. It's currently in public beta, so it's not recommended for all users, but it's simpler than Cosign, so you might want to migrate to it in the future.

In the future, we hope that Artifact Attestations will also be used for plugin verification.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,7 @@ on:
 permissions:
   contents: write
   id-token: write
+  attestations: write
 
 jobs:
   goreleaser:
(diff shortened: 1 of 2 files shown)


chenrui333 (commented):
this is really nice!!

PR #2101, opened 2024-08-14
chore: bump to use go 1.23.0

chenrui333:
- https://go.dev/doc/devel/release#go1.23.0

### DIFF
--- a/cmd/inspect_parallel.go
+++ b/cmd/inspect_parallel.go
@@ -58,7 +58,7 @@ func (cli *CLI) inspectParallel(opts Options) int {
 		}
 		if worker.err != nil {
 			// If the worker is canceled, suppress the error message.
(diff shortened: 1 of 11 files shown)


cmd/inspect_parallel.go
```diff
@@ -58,7 +58,7 @@ func (cli *CLI) inspectParallel(opts Options) int {
 		}
 		if worker.err != nil {
 			// If the worker is canceled, suppress the error message.
-			if errors.Is(context.Canceled, worker.err) {
+			if errors.Is(worker.err, context.Canceled) {
```
wata727, line 61: 😅

wata727 (approved):
Thanks!

cmd/inspect_parallel.go
chenrui333: ;)

PR #2111, opened 2024-09-02
chore: Update to owenrumney/go-sarif/v2

### LINKED ISSUE #2017: Sarif output should use the final schema version
### Summary

When the sarif output is used it will use the schema https://json.schemastore.org/sarif-2.1.0-rtm.5.json which is apparently now deprecated because a final version was released. The correct schema should be https://json.schemastore.org/sarif-2.1.0.json.

Steps to reproduce:

1. run tflint anywhere with the -f=sarif flag set
2. See the output use the schema "$schema": "https://json.schemastore.org/sarif-2.1.0-rtm.5.json"

I got the info from this [validator ](https://sarifweb.azurewebsites.net/Validation )which prints:
[SARIF1011](http://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html): The '$schema' property must refer to the final version of the SARIF 2.1.0 schema. This enables IDEs to provide Intellisense for SARIF log files. The SARIF standard was developed over several years, and many intermediate versions of the schema were produced. Now that the standard is final, only the OASIS standard version of the schema is valid.

### Command

tflint -f=sarif

### Terraform Configuration

```terraform

```

### TFLint Configuration

```HCL

```

### Output

```Shell
{
  "version": "2.1.0",
  "$schema": "https://json.schemastore.org/sarif-2.1.0-rtm.5.json",
  "runs": [
    {
      "tool": {
        "driver": {
          "name": "tflint",
          "version": "0.50.3",
          "informationUri": "https://github.com/terraform-linters/tflint"
        }
      },
      "results": []
    },
    {
      "tool": {
        "driver": {
          "name": "tflint-errors",
          "version": "0.50.3",
          "informationUri": "https://github.com/terraform-linters/tflint"
        }
      },
      "results": []
    }
  ]
}
```

### TFLint Version

0.50.3

### Terraform Version

_No response_

### Operating System

- [ ] Linux
- [ ] macOS
- [X] Windows

oWretch:
Update `owenrumney/go-sarif` to v2.3.3

Fixes #2017

### DIFF
--- a/formatter/sarif_test.go
+++ b/formatter/sarif_test.go
@@ -26,14 +26,15 @@ func Test_sarifPrint(t *testing.T) {
 			Issues: tflint.Issues{},
 			Stdout: fmt.Sprintf(`{
   "version": "2.1.0",
-  "$schema": "https://json.schemastore.org/sarif-2.1.0-rtm.5.json",
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/main/sarif-2.1/schema/sarif-schema-2.1.0.json",
   "runs": [
     {
       "tool": {
         "driver": {
+          "informationUri": "https://github.com/terraform-linters/tflint",
           "name": "tflint",
-          "version": "%s",
-          "informationUri": "https://github.com/terraform-linters/tflint"
+          "rules": [],
+          "version": "%s"
         }
       },
       "results": []
@@ -41,9 +42,10 @@ func Test_sarifPrint(t *testing.T) {
     {
       "tool": {
         "driver": {
+          "informationUri": "https://github.com/terraform-linters/tflint",
           "name": "tflint-errors",
-          "version": "%s",
-          "informationUri": "https://github.com/terraform-linters/tflint"
+          "rules": [],
+          "version": "%s"
         }
       },
       "results": []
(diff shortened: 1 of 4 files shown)


chenrui333 (approved):

chenrui333:
@oWretch thanks for your first contribution to tflint!

formatter/sarif_test.go
```diff
@@ -26,24 +26,26 @@ func Test_sarifPrint(t *testing.T) {
...
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/main/sarif-2.1/schema/sarif-schema-2.1.0.json",
   "runs": [
     {
       "tool": {
         "driver": {
+          "informationUri": "https://github.com/terraform-linters/tflint",
           "name": "tflint",
-          "version": "%s",
-          "informationUri": "https://github.com/terraform-linters/tflint"
+          "rules": [],
```
wata727, line 36: Are empty rules valid? 🤔

oWretch: From https://github.com/oasis-tcs/sarif-spec/blob/123e95847b13fbdd4cbe2120fa5e33355d4a042b/Schemata/sarif-schema-2.1.0.json#L3045 yes an empty rule block is valid. I had to add this as the new report included an empty rules block on generation, instead of omitting it in the rc version.

wata727: Thank you for the explanation. Indeed, this seems to be valid even with the [SARIF online validator](https://sarifweb.azurewebsites.net/Validation).

PR #2130, opened 2024-09-30
Update config.md

rquadling:
Fix the `ignore-module` example as the example used the deprecated `module` rather than `call_module_type`.

I'm unsure which value between `all` and `local` is appropriate, but feel `all` is more appropriate.


bendrucker (approved):
In this context all is definitely applicable as these modules are from the Terraform registry

Issue #2146, opened 2024-10-23
Quiet(er) option


pvickery-ParamountCommerce:
## Introduction

The `--quiet` option was removed in `0.11.0`, as it became the default, but I think it would be useful to add it back to allow even quieter output.
 https://github.com/terraform-linters/tflint/blob/8bb73bb470b3dad843fec425aa46d389e8c9f0b7/CHANGELOG.md?plain=1#L1497

For example, I run tflint from [pre-commit-terraform](https://github.com/antonbabenko/pre-commit-terraform) on a large repo with many projects and when a single project fails tflint, I get the following output for all the projects before any useful information about the findings. 

```
====================================================
working directory: dir1

Plugin "aws" is already installed
====================================================
working directory: dir2

Plugin "aws" is already installed
====================================================
working directory: dir3

Plugin "aws" is already installed
...
```
 

## Proposal
Can these logs be made option by using `--quiet`?

## References

https://github.com/terraform-linters/tflint/blob/8bb73bb470b3dad843fec425aa46d389e8c9f0b7/cmd/init.go#L27-L28
https://github.com/terraform-linters/tflint/blob/8bb73bb470b3dad843fec425aa46d389e8c9f0b7/cmd/init.go#L68
Related issue: https://github.com/terraform-linters/tflint/issues/1849

wata727:
Alternatively, it may be worth considering changing the behavior of `--recursive --init` so that it doesn't output anything if nothing happens in the subdir.

If we introduce `--quiet`, it is important to introduce subcommands first, since its name can be read as applying to other commands as well https://github.com/terraform-linters/tflint/issues/1618

PR #2150, opened 2024-10-31
Move recursive init output to debug when there are no changes

pvickery-ParamountCommerce:
With the current version of tflint, there can be a lot of output even when there are no changes. This PR moves the unneeded logs to DEBUG if there are no changes. 

Fixes https://github.com/terraform-linters/tflint/issues/2146

```bash
❯ tflint --recursive --config=$(pwd)/.tflint.hcl --init

Installing plugins on each working directory...

working directory: dir1

Plugin "aws" is already installed
working directory: dir2

Plugin "aws" is already installed
working directory: dir3

Plugin "aws" is already installed
...
```

```bash
❯ ~/projects/dist/tflint --recursive --config=$(pwd)/.tflint.hcl --init

Installing plugins on each working directory...

All plugins are already installed
```

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -2,7 +2,9 @@ package cmd
 
 import (
 	"fmt"
+	"log"
 	"os"
+	"strings"
 
 	"github.com/fatih/color"
 	"github.com/spf13/afero"
@@ -17,19 +19,25 @@ func (cli *CLI) init(opts Options) int {
 		return ExitCodeError
 	}
 
+	var builder strings.Builder
+
 	if opts.Recursive {
 		fmt.Fprint(cli.outStream, "Installing plugins on each working directory...\n\n")
 	}
 
+	any_installed := false
 	for _, wd := range workingDirs {
+		builder.Reset()
 		err := cli.withinChangedDir(wd, func() error {
+			installed := false
 			if opts.Recursive {
-				fmt.Fprint(cli.outStream, "====================================================\n")
-				fmt.Fprintf(cli.outStream, "working directory: %s\n\n", wd)
+				builder.WriteString("====================================================\n")
+				builder.WriteString(fmt.Sprintf("working directory: %s\n\n", wd))
 			}
 
 			cfg, err := tflint.LoadConfig(afero.Afero{Fs: afero.NewOsFs()}, opts.Config)
 			if err != nil {
+				fmt.Fprint(cli.outStream, builder.String())
 				return fmt.Errorf("Failed to load TFLint config; %w", err)
 			}
 
@@ -45,6 +53,8 @@ func (cli *CLI) init(opts Options) int {
 
 				_, err := plugin.FindPluginPath(installCfg)
 				if os.IsNotExist(err) {
+					fmt.Fprint(cli.outStream, builder.String())
+					builder.Reset()
 					fmt.Fprintf(cli.outStream, "Installing \"%s\" plugin...\n", pluginCfg.Name)
 
 					sigchecker := plugin.NewSignatureChecker(installCfg)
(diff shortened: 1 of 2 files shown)


pvickery-ParamountCommerce:
@wata727 I made this PR from your suggestion https://github.com/terraform-linters/tflint/issues/2146#issuecomment-2447820056

I just had to change and remove some test cases because of the modified behavior. Let me know what you think

cmd/init.go
```diff
@@ -45,6 +52,8 @@ func (cli *CLI) init(opts Options) int {
 
 				_, err := plugin.FindPluginPath(installCfg)
 				if os.IsNotExist(err) {
+					installed = true
+					fmt.Fprint(cli.outStream, builder.String())
```
wata727, line 56: Printing working directory here can result in verbose output for each plugin. e.g.

**before**
```
working directory : subdir

Installing "aws" plugin...
Installed "aws" (source: github.com/terraform-linters/tflint-ruleset-aws, version: 0.34.0)
Installing "google" plugin...
Installed "google" (source: github.com/terraform-linters/tflint-ruleset-google, version: 0.30.0)
```

**after**
```
working directory : subdir

Installing "aws" plugin...
Installed "aws" (source: github.com/terraform-linters/tflint-ruleset-aws, version: 0.34.0)
working directory : subdir

Installing "google" plugin...
Installed "google" (source: github.com/terraform-linters/tflint-ruleset-google, version: 0.30.0)
```

wata727: If this change helps to minimize installation logging, we may be able to adopt simpler output.

```
Installing "aws" plugin in subdir...
Installed "aws" (source: github.com/terraform-linters/tflint-ruleset-aws, version: 0.34.0)
Installing "google" plugin in subdir...
Installed "google" (source: github.com/terraform-linters/tflint-ruleset-google, version: 0.30.0)
```

pvickery-ParamountCommerce: True, I was just focused on minimizing output when no plugins are installed but we could extend this to have minimal output for all recursive calls. I am OK with how it is but I can make that change if you prefer.

wata727: Yeah, it's OK if you do it or not. However, please fix https://github.com/terraform-linters/tflint/pull/2150#discussion_r1827019301.
I just thought it might be easier to fix this issue by changing the output.

pvickery-ParamountCommerce: Ah thanks, I missed that. I've updated it to handle non-recursive calls correctly.

cmd/init.go
```diff
@@ -45,7 +52,8 @@ func (cli *CLI) init(opts Options) int {
 
 				_, err := plugin.FindPluginPath(installCfg)
 				if os.IsNotExist(err) {
-					fmt.Fprintf(cli.outStream, "Installing \"%s\" plugin...\n", pluginCfg.Name)
+					installed = true
+					builder.WriteString(fmt.Sprintf("Installing \"%s\" plugin...\n", pluginCfg.Name))
```
wata727, line 56: This message must be output in real time. It doesn't make sense to see "Installing..." printed after the installation is complete.

pvickery-ParamountCommerce: Sounds good. I updated it to immediately print the builder string and then send the rest of the output directly to stdout when a plugin is installed (or tried to install)

cmd/init.go
```diff
@@ -62,14 +72,29 @@ func (cli *CLI) init(opts Options) int {
...
+				continue
 			}
 
 			if opts.Recursive && !found {
-				fmt.Fprint(cli.outStream, "No plugins to install\n")
+				builder.WriteString("No plugins to install\n")
+			}
+
+			if installed || !opts.Recursive {
+				fmt.Fprint(cli.outStream, builder.String())
```
wata727, line 88: This `installed` flag is shared across all working directories, so the init output after the working directory where the plugin is installed will unintentionally be printed even if none are installed.

cmd/init.go
```diff
@@ -62,14 +72,29 @@ func (cli *CLI) init(opts Options) int {
 				}
 
 				if err != nil {
+					fmt.Fprint(cli.outStream, builder.String())
 					return fmt.Errorf("Failed to find a plugin; %w", err)
 				}
 
-				fmt.Fprintf(cli.outStream, "Plugin \"%s\" is already installed\n", pluginCfg.Name)
+				builder.WriteString(fmt.Sprintf("Plugin \"%s\" is already installed\n", pluginCfg.Name))
+				continue
```
wata727, line 80: Seems like a redundant continue.

cmd/init.go
```diff
@@ -62,14 +72,29 @@ func (cli *CLI) init(opts Options) int {
 				}
 
 				if err != nil {
+					fmt.Fprint(cli.outStream, builder.String())
 					return fmt.Errorf("Failed to find a plugin; %w", err)
 				}
 
-				fmt.Fprintf(cli.outStream, "Plugin \"%s\" is already installed\n", pluginCfg.Name)
+				builder.WriteString(fmt.Sprintf("Plugin \"%s\" is already installed\n", pluginCfg.Name))
```
wata727, line 79: This output is also buffered, so if the plugin is installed, all previous "already installed" messages are printed, but none are printed after that, so it's probably correct that it should never be printed.

cmd/init.go
pvickery-ParamountCommerce: Ah do you want to only show the directories that installed plugins? I set it up to show all when any plugins have been installed, but either is fine with me

wata727: Yes, I think the current behavior makes it difficult to understand the conditions under which the messages are shown. For example, consider the following structure:

```
├── aaa  # no plugins
├── bbb  # installed plugins
└── ccc  # no plugins
```

The output in `aaa` is cleared by `builder.Reset()` so nothing is shown. The output in `bbb` is shown because the plugins are installed. The output in `ccc` is shown because the `installed` flag is set.

In this case, it is difficult to understand why the output in `aaa` is not shown and the output in `ccc` is shown.

pvickery-ParamountCommerce:
Sorry for the delay, @wata727 I wasn't able to work on this for a few weeks.
Thanks again!

wata727 (approved):
👍

PR #2159, opened 2024-11-10
Introduce plugin keyless verification

wata727:
Fixes https://github.com/terraform-linters/tflint/issues/1351
Closes https://github.com/terraform-linters/tflint/pull/1902

TFLint has a mechanism to verify that plugins installed with `tflint --init` have not been tampered with by a malicious actor. Previously, the only way to do this was for plugin developers to sign the checksum files of their plugin binaries with their PGP key.

However, this approach has some problems: First, the plugin developer needs to manage their own PGP key, second, they need to consider how to distribute the public key.

To solve the problems, this PR introduces keyless verification as a new plugin verification mechanism. For the basic idea of ​​keyless verification see below:

- https://www.sigstore.dev/how-it-works
- https://docs.sigstore.dev/about/overview/
- https://docs.sigstore.dev/about/security/
- https://docs.sigstore.dev/about/threat-model/

The core technologies adopted here are [Artifact Attestations](https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds) and  [sigstore-go](https://github.com/sigstore/sigstore-go). Artifact attestation is a new feature published by GitHub that stores and distributes [Sigstore bundle](https://github.com/sigstore/protobuf-specs/blob/v0.3.2/protos/sigstore_bundle.proto) containing Fulcio-issued certificates and transparency logs corresponding to artifacts (checksum files for plugin binaries). The sigstore-go is a client library that provides policy-based verification of Sigstore bundles and is already adopted by [`gh attestation verify`](https://github.com/cli/cli/pull/8698).

Keyless verification ensures that the artifact was built by the GitHub Actions workflow for that repository. More specifically, it verifies that the SAN of the Fulcio-issued certificate matches the prefix of the plugin's `source` attribute. It is important to note that this approach differs from traditional ones in that it does not rely on trusting the signer. In this respect, it is weaker than the PGP approach, since an attacker could release a valid plugin simply by taking control of the repository.

This feature is introduced as an experimental feature, in other words, TFLint will only try to verify a plugin by artifact attestations if it is not signed with a PGP key. This means that official plugins (e.g. tflint-ruleset-aws), or that are already signed with a PGP key are not affected by this change. Even if there is no artifact attestation, there will be no error but just a warning "Plugin not verified" as before.

Here's a flowchart to help you understand if you're affected by this change:

```mermaid
flowchart TD
  q1{"Is it official plugin? (e.g. tflint-ruleset-aws)"} --->|Yes| na(Not Affected)
  q1 --->|No| q2
  q2{"Is it already signed with a PGP key?"} --->|Yes| na
  q2 --->|No| q3
  q3{"Does the repository have artifact attestation?"} --->|No| na
  q3 --->|Yes| a(Affected)
```

Third-party plugin developers can immediately benefit from this feature by simply adding `actions/attest-build-provenance` to their GitHub Actions release workflow. Please refer to https://github.com/terraform-linters/tflint-ruleset-template/pull/103 as an example PR. If you have already created a plugin from the latest template repository, make sure the repository has artifact attestations.

In the future, I plan to remove the built-in PGP key and migrate to this mechanism for official plugins. To test this behavior, this PR introduces a new `TFLINT_EXPERIMENTAL` environment variable. Running TFLint with `TFLINT_EXPERIMENTAL=true` will ignore the built-in keys for official plugins and use artifact attestations for verification, and will error out installation of third-party plugins that are not verified by PGP keys or artifact attestations. This way is inspired by [`COSIGN_EXPERIMENTAL`](https://github.com/sigstore/cosign/blob/v2.4.1/VERSIONING.md?plain=1#L84-L85). I also considered adding the new option to the `plugin` block instead of an environment variable, but I wasn't sure if we would still need it, so I went with this approach.


wata727:
@bendrucker Could you take a look at this? Any feedback on the phase-in of plugin keyless verification?

docs/user-guide/plugins.md
```diff
@@ -132,3 +134,11 @@ plugin "terraform" {
...
+
+## Keyless verification (experimental)
+
+If the plugin developer has generated [Artifact Attestations](https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds), TFLint will automatically verify them and prove that the plugin binary was built in that repository.
+
+This verification is experimental and optional: it is only attempted if there is no PGP public signing key, and if there is no artifact attestation, a warning will be output, not an error. If you want to require all plugin installs to be signed with a PGP signing key or an artifact attestation, you can force this behavior to be enabled by setting the `TFLINT_EXPERIMENTAL=1`. This behavior will be the default in future versions, but is subject to change without notice.
+
+Note that this validation, like the PGP signing key, does not guarantee that the plugin is secure. Moreover it only guarantees the repository it was built from, not the signer, so it is not secure if an attacker has control over the repository.
```
bendrucker, line 144:
```suggestion
Note that this validation, like the PGP signing key, does not guarantee that the plugin is secure. It only attests the source repository/revision from which it was built. It prevents direct upload of malicious release artifacts to GitHub or manipulation of download requests. If an attacker has control over the repository and can perform execution during a build, any resulting malicious release will still be considered "verified."
```

bendrucker (approved):
Nice! Just a small comment around the documentation.

PR #2176, opened 2024-12-06
install: handle GitHub API JSON without newlines

### LINKED ISSUE #2175: Trying to download mentions_count as a version
### Summary

We are running tflint on many of our CI's

From time to time it fails installing with the following error:

```
Downloading TFLint mentions_count
curl: (22) The requested URL returned error: 404 

Exited with code exit status 22
```

We are installing tflint this way:
```
curl  --retry 3 -s https://raw.githubusercontent.com/terraform-linters/tflint/master/install_linux.sh | TFLINT_VERSION=latest bash
```

My guess is that this grep https://github.com/terraform-linters/tflint/blob/master/install_linux.sh#L46 is not catching the tag_name but another row. Not exactly sure why at the moment. 

**Downloading TFLint mentions_count** seems totally wrong.

### Command

TFLINT_VERSION=latest bash

### Terraform Configuration

```terraform
n/a
```

### TFLint Configuration

```HCL
n/a
```

### Output

```Shell
Downloading TFLint mentions_count
curl: (22) The requested URL returned error: 404 

Exited with code exit status 22
```
```


### TFLint Version

0.54.0

### Terraform Version

1.9

### Operating System

- [X] Linux
- [ ] macOS
- [ ] Windows

bendrucker:
Handles GitHub API responses that do not include newlines. Previously, if the `releases/latest` response lacked newlines within its JSON, `get_latest_release` would return `mentions_count` since that is the last double-quoted string on the line grep matched (the whole object).

Now, `sed` will only capture the string value after `"tag_name"`. It looks past a colon and optional surrounding whitespace and captures the next double quoted group. This is compatible with both compact and multiline JSON.

We still need `grep` to handle multiline JSON, since we need to exclude all lines other than the one we want. Otherwise the output is the original object with a single malformed line containing the latest version. GNU `sed` can handle multiline input with `-z` so keeping `grep` for that preserves portability.

Closes #2175

### DIFF
--- a/install_linux.sh
+++ b/install_linux.sh
@@ -43,8 +43,8 @@ get_latest_release() {
       headers=(-H "Authorization: Bearer ${GITHUB_TOKEN}")
   fi
   curl --fail -sS "${headers[@]}" "https://api.github.com/repos/terraform-linters/tflint/releases/latest" | # Get latest release from GitHub api
-    grep '"tag_name":' |                                                                                    # Get tag line
-    sed -E 's/.*"([^"]+)".*/\1/'                                                                            # Pluck JSON value
+    grep '"tag_name":' |                                                                                    # Filter other lines from multiline JSON
+    sed -E 's/.*"tag_name"[[:space:]]*:[[:space:]]*"([^"]+)".*/\1/'                                         # Get tag_name value
 }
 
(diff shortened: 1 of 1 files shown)


wata727 (approved):
👍

PR #2208, opened 2025-01-14
chore(dependabot): group dependency updates

chenrui333:
group all dependencies update in one pr per package-ecosystem, https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates#prioritizing-meaningful-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,10 @@ updates:
     directory: "/"
     schedule:
       interval: weekly
+    groups:
+      all:
+        patterns:
+          - "*"
     open-pull-requests-limit: 10
 
   # Maintain dependencies for GitHub Actions
@@ -11,9 +15,17 @@ updates:
     directory: "/"
     schedule:
       interval: weekly
+    groups:
+      all:
+        patterns:
+          - "*"
 
   # Maintain dependencies for docker
   - package-ecosystem: "docker"
     directory: "/"
     schedule:
       interval: weekly
+    groups:
+      all:
+        patterns:
+          - "*"


bendrucker:
Yeah to me this feature is built for large packages that are published as smaller units (e.g., AWS SDK) or related packages (e.g., for a linter). In these cases merging many updates one at a time can be at best tedious (SDK) and at worst impossible (in the case of tightly coupled dependencies, e.g., a tool and its plugins). Grouping on ecosystem spoils the signal of a test failure for a given dependency update. A big reason for automated dependency updates is to see exactly where something breaks rather than having to handle big version jumps across many packages.

chenrui333:
yeah, I can see the risk of group updates would incur breaking changes, but in reality, it just never happens. 

I also just saw this feature recently (apparently it has been around for quite some time).

I'd like to propose to merge it and see if it indeed cause some issues (I also tried in my other projects too) 

let me know what you guys think.

bendrucker:
> in reality, it just never happens

This is normal though, most dependency updates are noops, but occasionally they matter.

What would be the goal of grouping all dependencies regardless in a single PR? If it's to reduce toil, auto-merging PRs would seem to be a better way to accomplish that without giving up incremental visibility when something does fail.

wata727:
While I suppose it's fine in most cases, I feel uncomfortable updating packages of different concerns together.
I would like to hold off on merging until this change is truly necessary.

chenrui333:
At this point, I guess we are comfortable with what we have, we can totally revisit this in the future. Thanks!

Issue #2228, opened 2025-02-06
Linux installer reverts to arm 32 bit install on aarch64 when dpkg is not installed


jimmydavies:
### Summary

[This code here](https://github.com/terraform-linters/tflint/blob/master/install_linux.sh#L6-L14) checks the architecture of the system before installing. It looks like this
```bash
get_machine_arch () {
    machine_arch=""
    case $(uname -m) in
        i386)     machine_arch="386" ;;
        i686)     machine_arch="386" ;;
        x86_64)   machine_arch="amd64" ;;
        arm64)    machine_arch="arm64" ;;
        aarch64)  dpkg --print-architecture | grep -q "arm64" && machine_arch="arm64" || machine_arch="arm" ;;
    esac
    echo $machine_arch
}
```

For the `aarch64` case when dpkg is not installed it falls back to `arm` architecture, which has caused issues for us where an `aarch64` system tried to run an `arm` binary and fails.

I'm not well versed in the nuances of machine architecture but my assumption is that `aarch64` will always be 64 bit and so `aarch64` in this case should always resolve to `arm64`, these two competing spec have since merged anyway ([see this article](https://www.phoronix.com/news/MTY5ODk))

I think the `aarch64` case should just be
```bash
aarch64) machine_arch="arm64";;
```

### Command

curl -s https://raw.githubusercontent.com/terraform-linters/tflint/master/install_linux.sh | bash

### Terraform Configuration

```terraform
NA this issue is with the installer
```

### TFLint Configuration

```HCL
NA this issue is with the installer
```

### Output

```Shell
curl -s https://raw.githubusercontent.com/terraform-linters/tflint/master/install_linux.sh | bash

main: line 13: dpkg: command not found
arch=arm
os=linux_arm
====================================================
Looking up the latest version ...
Downloading TFLint v0.55.1
Downloaded successfully
====================================================
Unpacking /tmp/tflint.XXXXlGOelO/tflint.zip ...
Archive:  /tmp/tflint.XXXXlGOelO/tflint.zip
  inflating: tflint
Installing /tmp/tflint.XXXXlGOelO/tflint to /usr/local/bin/ ...
'/tmp/tflint.XXXXlGOelO/tflint' -> '/usr/local/bin/tflint'
Cleaning temporary downloaded files directory /tmp/tflint.XXXXlGOelO ...
====================================================
Current tflint version
bash: line 105: /usr/local/bin//tflint: cannot execute binary file: Exec format error
```

[...]

PR #2229, opened 2025-02-07
Remove dependency on dpkg in install_linux.sh

jimmydavies:
For the purposes of tflint arm64 and aarch64 are equivalent and so we don't need to keep the logic that checks for the user space being 32 bit. This allows us to not rely on dpkg being installed, but more importantly stops 64 bit systems from falling back to 32bit tflint install when dpkg is not present


jimmydavies:
Fixes #2228

PR #2230, opened 2025-02-09
Add support for tflint-ignore-file annotations in JSON

isobit:
I have a use case where I'm generating `.tf.json` files for a bunch of different root modules; these `.tf.json` files define several locals for convenience, which may or may not be used by the root modules they're placed in. I wanted to be able to ignore the `terraform_unused_declarations` rule without disabling it globally, but I realized that there wasn't a way to do this since the annotation system didn't support JSON files at all. This change adds limited support for `tflint-ignore-file` annotations in a top-level comment property.

### DIFF
--- a/cmd/inspect.go
+++ b/cmd/inspect.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/hashicorp/go-version"
 	"github.com/hashicorp/hcl/v2"
@@ -223,9 +222,6 @@ func (cli *CLI) setupRunners(opts Options, dir string) (*tflint.Runner, []*tflin
 	}
 	annotations := map[string]tflint.Annotations{}
 	for path, file := range files {
-		if !strings.HasSuffix(path, ".tf") {
-			continue
-		}
 		ants, lexDiags := tflint.NewAnnotations(path, file)
 		diags = diags.Extend(lexDiags)
 		annotations[path] = ants
--- a/docs/user-guide/annotations.md
+++ b/docs/user-guide/annotations.md
@@ -79,3 +79,52 @@ resource "aws_instance" "foo" { # tflint-ignore-file: aws_instance_invalid_type
   instance_type = "t1.2xlarge"
 }
 ```
+
+The `tflint-ignore-file` annotation is also supported in Terraform JSON by
+using a top-level [comment property](https://developer.hashicorp.com/terraform/language/syntax/json#comment-properties):
+
+```json
+{
+  "//": "tflint-ignore-file: aws_instance_invalid_type",
+  "resource": {
+    "aws_instance": {
+      "foo": {
+        "instance_type": "t2.micro"
+      }
+    }
+  }
+}
+```
+
+As with annotations in HCL files, multiple rules can be specified as a
+comma-separated list:
+
+```json
+{
+  "//": "tflint-ignore-file: aws_instance_invalid_type, other_rule",
+  "resource": {
+    "aws_instance": {
+      "foo": {
+        "instance_type": "t2.micro"
+      }
+    }
+  }
+}
+```
+
+Similarly, annotations in JSON can be followed with arbitrary comments, but the annotation must be the first thing in the comment property string:
+
+```json
+{
+  "//": "tflint-ignore-file: aws_instance_invalid_type # This instance type is new and TFLint doesn't know about it yet",
+  "resource": {
+    "aws_instance": {
+      "foo": {
+        "instance_type": "t2.micro"
+      }
+    }
+  }
+}
+```
+
+The `tflint-ignore` annotation is not supported in JSON configuration.
--- a/docs/user-guide/annotations.md
+++ b/docs/user-guide/annotations.md
@@ -55,6 +55,8 @@ resource "aws_instance" "foo" {
 }
 ```
 
+## Files
+
 To disable an entire file, you can also use the `tflint-ignore-file` annotation:
 
 ```hcl
@@ -80,8 +82,9 @@ resource "aws_instance" "foo" { # tflint-ignore-file: aws_instance_invalid_type
 }
 ```
 
-The `tflint-ignore-file` annotation is also supported in Terraform JSON by
-using a top-level [comment property](https://developer.hashicorp.com/terraform/language/syntax/json#comment-properties):
+## JSON
+
+The `tflint-ignore-file` annotation is also supported in Terraform JSON by using a top-level [comment property](https://developer.hashicorp.com/terraform/language/syntax/json#comment-properties):
 
 ```json
 {
--- a/tflint/annotation.go
+++ b/tflint/annotation.go
@@ -1,6 +1,7 @@
 package tflint
 
 import (
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"slices"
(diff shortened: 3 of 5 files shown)


cmd/inspect.go
```diff
@@ -223,9 +222,6 @@ func (cli *CLI) setupRunners(opts Options, dir string) (*tflint.Runner, []*tflin
 	}
 	annotations := map[string]tflint.Annotations{}
 	for path, file := range files {
-		if !strings.HasSuffix(path, ".tf") {
-			continue
-		}
```
wata727, line 228: This also needs to be fixed.
https://github.com/terraform-linters/tflint/blob/fcbcce59c42d22c8ea69697c3c81ae85ee941413/langserver/handler.go#L181-L183

tflint/annotation.go
```diff
@@ -66,6 +76,63 @@ func NewAnnotations(path string, file *hcl.File) (Annotations, hcl.Diagnostics)
...
+		return ret, diags
+	}
+
+	rawComment, ok := data["//"]
+	if !ok {
+		return ret, diags
+	}
+
+	var comment jsonAnnotationComment
+	if err := json.Unmarshal(rawComment, &comment); err != nil {
```
wata727, line 106: If there are any comments other than annotations, an error will occur here.

```json
{
  "//": "This file is auto-generated. DO NOT EDIT!"
}
```

isobit: Oh yep, oversight on my part, will fix.

docs/user-guide/annotations.md
```diff
@@ -79,3 +79,24 @@ resource "aws_instance" "foo" { # tflint-ignore-file: aws_instance_invalid_type
...
 }
 ```
+
+The `tflint-ignore-file` annotation is also supported in Terraform JSON by
+using a top-level [comment property](https://developer.hashicorp.com/terraform/language/syntax/json#comment-properties):
+
+```json
+{
+  "//": {
+    "tflint-ignore-file": "aws_instance_invalid_type"
```
bendrucker, line 89: Are there any other examples you can offer of this map comment structure being used? The documentation illustrates comment properties with string values. In that sense, this proposal clashes with the official guidance, as a JSON document that already contains a comment property (string) would be unable to conform to this object syntax.

Seems like `//` could be parsed equivalently to an HCL document with a first-line comment.

docs/user-guide/annotations.md
```diff
@@ -79,3 +79,24 @@ resource "aws_instance" "foo" { # tflint-ignore-file: aws_instance_invalid_type
...
+      "foo": {
+        "instance_type": "t2.micro"
+      }
+    }
+  }
+}
+```
+
+Note that the `tflint-ignore` annotation is not currently supported in
+Terraform JSON config.
```
bendrucker, line 120:
```suggestion
The `tflint-ignore` annotation is not supported JSON configuration.
```

Avoid "currently" in documentation and similar caveats. Just say how it works right now.

tflint/annotation.go
```diff
@@ -66,6 +76,64 @@ func NewAnnotations(path string, file *hcl.File) (Annotations, hcl.Diagnostics)
 	return ret, diags
 }
 
+// jsonAnnotations finds annotations in .tf.json files. Currently only
+// file-level ignores are supported, by specifying a root-level comment
+// property (with key "//") which is an object containing a string property
+// with the key "tflint-ignore-file".
```
bendrucker, line 82:
```suggestion
// jsonAnnotations finds annotations in .tf.json files. Only
// file-level ignores are supported, by specifying a root-level comment
// property (with key "//") which is an object containing a string property
// with the key "tflint-ignore-file".
```

tflint/annotation.go
```diff
@@ -66,6 +76,64 @@ func NewAnnotations(path string, file *hcl.File) (Annotations, hcl.Diagnostics)
...
+	if err := json.Unmarshal(file.Bytes, &data); err != nil {
+		diags = append(diags, &hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "failed to unmarshal file as JSON",
+			Detail:   err.Error(),
+			Subject: &hcl.Range{
+				Filename: path,
+				// Unfortunately Start and End are difficult to determine here
+				// since encoding/json does not expose source metadata.
+			},
```
bendrucker, line 97:
```suggestion
			Subject: &hcl.Range{
				// Cannot set Start/End because encoding/json does not expose it
				Filename: path,
			},
```

Avoid feelings language (unfortunately, difficult), emphasize constraints and outcome

docs/user-guide/annotations.md
isobit: I have no examples of this being used "in the wild"; my reasoning for the map structure was only that, hypothetically, someone might want to be able to use the comment property to leave other information. It wasn't clear to me how that would work if the comment property was a string which was already occupied by the `tflint-ignore-file` annotation.

For example:

```json
{
  "//": {
    "tflint-ignore-file": "foo",
    "_1": "we ignore foo because reasons",
    "_2": "this file does stuff"
  }
}
```

I'm realizing now though that the JSON spec may allow for multiple properties with the key `//`; that may be a cleaner way to have multiple comments, at the expense of being harder to parse here (`encoding/json` discards duplicate entries). I could look for a way of handling that though. E.g.

```json
{
  "//": "this file does stuff",
  "//": "we ignore foo because reasons",
  "//": "tflint-ignore: foo"
}
```

I'm open to alternatives though, it might make sense to expect a escaped linefeed-delimited string or a list of strings. E.g.

```json
{
  "//": "this file does stuff\nwe ignore foo because reasons\ntflint-ignore: foo"
}
```

or


```json
{
  "//": [
    "this file does stuff",
    "we ignore foo because reasons",
    "tflint-ignore: foo"
  ]
}
```

bendrucker: This was my assumption, thanks. I'd treat it as a string then, with the same rules as HCL:

https://github.com/terraform-linters/tflint/blob/fcbcce59c42d22c8ea69697c3c81ae85ee941413/tflint/annotation.go#L49

So if you have some existing comment property in the config, `tflint-ignore-file` would have to be prepended to that. Followed by whitespace, then any other comment content.

The map or list patterns are not inherently unreasonable, but TFLint tries to avoid creating Terraform patterns of its own.

isobit: Makes sense to me; will update in a bit here.

wata727: > The documentation illustrates comment properties with string values

I thought the same thing, but as mentioned in https://github.com/terraform-linters/tflint/pull/2230#discussion_r1949900509, I thought a map would not be a bad choice to add other comments.

In fact, having looked at the HCL JSON parser, it doesn't seem to matter what the type of the comment value is:
https://github.com/hashicorp/hcl/blob/v2.23.0/json/structure.go#L55-L59

Personally, I'm concerned that comments like the following look like comments to the `tflint-ignore-file` annotation:

```json
{
  "//": "tflint-ignore-file: aws_instance_invalid_type # This file is auto-generated by tools/terraform_gen.go. DO NOT EDIT",
}
```

I like the following style the most:

```json
{
  "//": "tflint-ignore-file: aws_instance_invalid_type",
  "//": "This file is auto-generated by tools/terraform_gen.go. DO NOT EDIT"
}
```

However, given the difficulties involved when you have multiple keys with the same name, I believe we need to find some compromise.

bendrucker: All these options have sharp edges. Ideally I think you'd have this config if you wanted an ignore and an unrelated comment:

```json
{
  "//": "tflint-ignore-file: aws_instance_invalid_type\nThis file is auto-generated by tools/terraform_gen.go. DO NOT EDIT",
}
```

If you put them on the same line it would imply that the comment described the ignore.

If you have some existing code that generates a comment property, you can just update it to `ignore + "\n" + existing` and you're set. Expecting a map or array requires more refactoring, even if it's ultimately very minor.

Ultimately JSON is meant for generated config and not humans anyway and so I err on the side of "good enough" versus something more expressive/structured.

bendrucker (approved):
Realizing this has been dangling for a while. I think the string approach should be workable given the limited applications.

PR #2239, opened 2025-02-13
feat: bump to build with go1.24

chenrui333:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/terraform-linters/tflint
 
-go 1.23
+go 1.24.0
```
wata727, line 3: Is the patch version needed here?

chenrui333: it is needed, otherwise it would append toolchain

chenrui333:
Thanks @wata727!

PR #2274, opened 2025-03-25
go: use `govulncheck` to detect security issues

bendrucker:
Uses [`govulncheck`](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) to check each build for known Go vulnerabilities in the source.

Inspired by https://github.com/opentofu/opentofu/pull/2600, this attempts to cut through the steady drip of noisy automated reports. Where most automated reports detect inclusion of a vulnerable dependency version via package manifests (`go.mod`, `go.sum`), `govulncheck` analyzes the source code for use of vulnerable _symbols_. This has a dramatically higher signal to noise ratio. Reports from `govulncheck` should often be actionable and require an advisory. Anything not reported is most likely noise and does not merit an advisory. 

Uses the new `go tool` features in Go 1.24 to invoke the command.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,3 +40,16 @@ jobs:
       with:
         name: tflint-${{ matrix.os }}
         path: dist/tflint
+  vulnerabilities:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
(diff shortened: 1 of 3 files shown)


wata727 (approved):
Nice!

Issue #2209, opened 2025-01-15
Failed to verify signature


nikhithabyreddy:
### Summary

We have a in house plugin created to check for tags in resources so that is being installed as part of tflint init  while performing the tflint --init we are getting the below error 
14:28:05 signature.go:135: [DEBUG] Failed to verify signature: failed to verify log inclusion: not enough verified log entries from transparency log: 0 < 1
Failed to install a plugin; Failed to check checksums.txt signature: failed to verify log inclusion: not enough verified log entries from transparency log: 0 < 1

### Command

tflint --init

### Terraform Configuration

```terraform
terraform {
  required_version = "1.9.3"
```

### TFLint Configuration

```HCL
plugin "xxxx" {
  enabled = true
  version = "0.1.0"
  source  = "github.com/xxxx/tflint-ruleset-xxxx"
}
```

### Output

```Shell
14:28:05 signature.go:135: [DEBUG] Failed to verify signature: failed to verify log inclusion: not enough verified log entries from transparency log: 0 < 1
Failed to install a plugin; Failed to check checksums.txt signature: failed to verify log inclusion: not enough verified log entries from transparency log: 0 < 1
```

### TFLint Version

0.55.0

### Terraform Version

1.9.3

### Operating System

- [X] Linux
- [ ] macOS
- [ ] Windows

wata727:
Thank you for reporting this. This is due to [plugin keyless verification](https://github.com/terraform-linters/tflint/pull/2159) introduced in v0.55.
From the error, it looks like the transparency logs aren't included in the artifact attestation, but I'm not sure what cases that would happen.

Can you share the artifact attestation of the plugin you have installed?
You can see them in https://github.com/xxxx/tflint-ruleset-xxxx/attestations

nikhithabyreddy:
@wata727  Here is the artifact attestation 
[attestation-sigstore (2).json](https://github.com/user-attachments/files/18427806/attestation-sigstore.2.json)

nikhithabyreddy:
@wata727  Any update on the above issue ?

wata727:
After some research, it appears that attestations built in private repositories do not include transparency logs.
https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds#about-using-sigstore-for-artifact-attestations

Just like with GitHub CLI, we'll need to switch verification policies.
[...]

PR #2292, opened 2025-04-28
plugin: Skip keyless verification for private third-party plugins

ricardo-kh:
Comparable to #2223, which became ineffective in #2224. 

https://github.com/terraform-linters/tflint/issues/2209#issuecomment-2835204766

### DIFF
--- a/plugin/install.go
+++ b/plugin/install.go
@@ -206,8 +206,7 @@ func (c *InstallConfig) tryKeylessVerifyChecksumsSignature(sigchecker *Signature
 	// If the repository is private, artifact attestations is not always available
 	// because it requires GitHub Enterprise Cloud plan, so we skip verification here.
 	if repo.Private != nil && *repo.Private {
-		log.Printf("[DEBUG] Bypassing verification of artifact attestations for private repository")
-		return true, nil
+		return false, nil
 	}
 
 	log.Printf("[DEBUG] Download artifact attestations")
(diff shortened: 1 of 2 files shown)


plugin/install.go
```diff
@@ -206,7 +206,8 @@ func (c *InstallConfig) tryKeylessVerifyChecksumsSignature(sigchecker *Signature
 	// If the repository is private, artifact attestations is not always available
 	// because it requires GitHub Enterprise Cloud plan, so we skip verification here.
 	if repo.Private != nil && *repo.Private {
-		return false, nil
+		log.Printf("[DEBUG] Bypassing verification of artifact attestations for private repository")
+		return true, nil
```
wata727, line 210: It is correct that `tryKeylessVerifyChecksumsSignature` returns `false` because the verification is being skipped.
From what I understand, returning `false` here will still work fine.

- `Install` returns a installed path and `ErrPluginNotVerified`
  - https://github.com/terraform-linters/tflint/blob/e599327729c2b79fdcc4a6a438466cca9034390d/plugin/install.go#L170-L172
- The caller `init` will ignore this error.
  - https://github.com/terraform-linters/tflint/blob/e599327729c2b79fdcc4a6a438466cca9034390d/cmd/init.go#L53-L60

Maybe the reason is that we didn't set the `err` to nil here...

ricardo-kh: Setting the `err` to nil would also do the trick indeed.

ricardo-kh: I reverted my previous suggestion and now trying your suggestion: https://github.com/terraform-linters/tflint/pull/2292/commits/08f3cde002a951b1164c4da7019d6c7103aa4fec

wata727 (approved):
Thanks!

PR #2306, opened 2025-05-12
fix: load `format` from config file

### LINKED ISSUE #2305: `format` in config files is ignored
When `format` is set in a config file, it doesn't do anything. Only the `--format` flag works.

```hcl
config {
  format = "compact"
}
```

```sh
tflint --config config.hcl
```

```
1 issue(s) found:

Warning: terraform "required_version" attribute is required (terraform_required_version)

  on main.tf line 1:

Reference: https://github.com/terraform-linters/tflint-ruleset-terraform/blob/v0.10.0/docs/rules/terraform_required_version.md
```

```sh
tflint --format compact
```

```
1 issue(s) found:

main.tf:1:1: Warning - terraform "required_version" attribute is required (terraform_required_version)
```

bendrucker:
Fixes handling of `format` inside a `config {}` block in a TFLint configuration file. Previously, `cli.formatter.Format` was set to `options.Format` and never considered the configuration file.

Now, `NewCLI` parses the config too and uses the merged config and options to set the format. The implementation is a bit hacky because a formatter needs to be constructed early to print errors.

Closes #2305

### DIFF
--- a/cmd/cli.go
+++ b/cmd/cli.go
@@ -14,6 +14,7 @@ import (
 	"github.com/fatih/color"
 	"github.com/hashicorp/logutils"
 	flags "github.com/jessevdk/go-flags"
+	"github.com/spf13/afero"
 	"github.com/terraform-linters/tflint/formatter"
 	"github.com/terraform-linters/tflint/terraform"
 	"github.com/terraform-linters/tflint/tflint"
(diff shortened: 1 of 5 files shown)


wata727 (commented):
Thank you for working on this.

This issue seems a bit trickier than we thought. As I left in the comments, we can't simply load the config first. It looks like we need to consider a different approach.

bendrucker:
Closing in favor of #2320, makes sense that since only inspection uses `--format` it can just be handled there rather than in `cli.go`

PR #2391, opened 2025-09-24
Add JSON configuration file support

### LINKED ISSUE #2389: TFLInt JSON Configuration Support
## Introduction

TFLint currently appears to only accepts configuration files in HCL format (.tflint.hcl). This limitation makes it challenging to programmatically generate TFLint configurations from build systems and tools that natively work with JSON. Supporting JSON format for TFLint configuration files would enable better integration with automated tooling and build systems.

## Proposal

Add support for JSON-formatted configuration files in TFLint, similar to how Terraform itself supports both .tf (HCL) and .tf.json (JSON) formats. This would allow TFLint to accept configuration files named .tflint.json in addition to the existing .tflint.hcl.

For what it's worth, this should likely be simply a case of exposing the json parser of the hcl/v2 library using `hclparse.ParseJSON()`. If you're happy to proceed, I will make a PR, with tests, to introduce it to the parser.

### Use cases:

- Build systems like Bazel that have strong JSON support can programmatically generate TFLint configurations
- CI/CD pipelines can dynamically create configurations using JSON templating tools
- Teams using infrastructure-as-code generators can emit TFLint config alongside Terraform JSON
- Integration with tools that already output JSON-based configurations (e.g., policy engines, compliance tools)

### Example .tflint.json:
```json
{
  "config": {
    "format": "compact",
    "module": true
  },
  "plugin": {
    "aws": {
      "enabled": true,
      "version": "0.21.0",
      "source": "github.com/terraform-linters/tflint-ruleset-aws"
    }
  },
  "rule": {
    "terraform_naming_convention": {
      "enabled": true
    }
  }
}
```

### Implementation Approach
- Accept both `.tflint.hcl` and `.tflint.json` as valid configuration filenames
- When `.tflint.json` is detected, parse it as JSON and convert to the internal configuration structure
- Maintain backward compatibility by continuing to support HCL format
- Follow the same JSON structure conventions that Terraform uses for .tf.json files

## References

- Terraform JSON Configuration Syntax: https://www.terraform.io/docs/language/syntax/json.html
- Bazel's native JSON support via json.encode(): https://bazel.build/rules/lib/json
- Similar feature in Terraform allowing both HCL and JSON formats

rorychatterton:
closes #2389 

This change adds support for JSON-formatted TFLint configuration files (.tflint.json) in addition to the existing HCL format (.tflint.hcl).

Changes:
- Update config loading to detect file extension and parse JSON/HCL accordingly
- Add support for .tflint.json in default file discovery
- Maintain HCL preference over JSON files
- Add unit tests for JSON config loading scenarios
- Add integration tests for JSON config functionality and precedence
- Update user guide documentation with JSON format examples

The implementation uses the existing hclparse.ParseJSON() from the HCL library, following the same JSON structure conventions as Terraform's .tf.json files. Existing functionality remains unchanged.

Other:
- [x] Unit Tests Pass
- [x] Integration Tests Pass in CI
- [x] Confirm Source Map - As in the other thread, need to confirm that the sources data structure supports JSON, or whether we need to create an intermediary representation as HCL
- [x] Update Bundled Plugin to use SDK 0.23.0 

For full transparency, I used Claude Code to help generate this commit, but am responsible for reviews etc.


cmd/inspect.go
```diff
@@ -271,44 +292,44 @@ func launchPlugins(config *tflint.Config, fix bool) (*plugin.Plugin, error) {
 		if err != nil {
 			if st, ok := status.FromError(err); ok && st.Code() == codes.Unimplemented {
 				// VersionConstraints endpoint is available in tflint-plugin-sdk v0.14+.
-				return rulesetPlugin, fmt.Errorf(`Plugin "%s" SDK version is incompatible. Compatible versions: %s`, name, plugin.SDKVersionConstraints)
+				return rulesetPlugin, sdkVersions, fmt.Errorf(`Plugin "%s" SDK version is incompatible. Compatible versions: %s`, name, plugin.SDKVersionConstraints)
```
rorychatterton, line 295: This shouldn't be propegated up anymore now that validate is internal to launchPlugins

rorychatterton:
Sorry about the commit noise. I use Graphite to manage my CI in the backend and it gets a little *pushy*.

This should be ready to review now. In summary:

  ## Changes by Component

  **Plugin Version Checking** (`plugin/plugin_version.go`)
  - Extracted all version validation logic from `cmd/inspect.go` into
  centralised functions
  - Created `CheckTFLintVersionConstraints()` and
  `CheckSDKVersionSatisfiesConstraints()`
  - Added version constants: DefaultSDKVersionConstraints (>=0.16.0),
  JSONConfigSDKVersionConstraints (>=0.23.0)

  **Config Loading** (`tflint/config.go`, `tflint/config_test.go`)
  - Modified `LoadConfig()` to detect JSON files by extension and parse
  accordingly
  - Added `isJSONConfig` field and `IsJSONConfig()` method to track config
  type
  - Updated tests to cover JSON loading scenarios

  **Inspection Flow** (`cmd/inspect.go`)
  - Removed inline version checking code, now calls centralized functions
  - Split checks: TFLint constraints in `launchPlugins()` before
  `ApplyGlobalConfig`, SDK checks in `ValidatePluginVersions()` after

  **Language Server** (`langserver/handler.go`)
  - Updated to use centralized version checking functions
  - Now properly detects JSON configs via `cfg.IsJSONConfig()` and enforces
  appropriate SDK requirements

  **Testing**
  - Added `plugin/plugin_version_test.go` with SDK version validation tests
  - Added integration tests in `integrationtest/cli/` for JSON config loading
   and HCL/JSON precedence

plugin/server.go
```diff
@@ -146,7 +146,7 @@ func (s *GRPCServer) EvaluateExpr(expr hcl.Expression, opts sdk.EvaluateExprOpti
 
 	// If an ephemeral mark is contained, cty.Value will not be returned
 	// unless the plugin is built with SDK 0.22+ which supports ephemeral marks.
-	if !marks.Contains(val, marks.Ephemeral) || s.clientSDKVersion.GreaterThanOrEqual(version.Must(version.NewVersion("0.22.0"))) {
+	if !marks.Contains(val, marks.Ephemeral) || SupportsEphemeralMarks(s.clientSDKVersion) {
```
rorychatterton, line 149: Moved this hardcoded version to plugin_version.go

tflint/config.go
```diff
@@ -212,7 +245,17 @@ func loadConfig(file afero.File) (*Config, error) {
...
+	var f *hcl.File
+	var diags hcl.Diagnostics
+
+	// Parse based on file extension
+	switch {
+	case strings.HasSuffix(strings.ToLower(file.Name()), ".json"):
+		f, diags = parser.ParseJSON(src, file.Name())
+	default:
+		f, diags = parser.ParseHCL(src, file.Name())
+	}
```
bendrucker: A switch expression with case expressions is easier to read than a bare switch:

```suggestion
	switch filepath.Ext(file.Name()) {
	case ".json":
		f, diags = parser.ParseJSON(src, file.Name())
	default:
		f, diags = parser.ParseHCL(src, file.Name())
	}
```

You'll need to import `filepath` too

tflint/config.go
```diff
@@ -228,6 +271,8 @@ func loadConfig(file afero.File) (*Config, error) {
 
 	config := EmptyConfig()
 	config.sources = parser.Sources()
+	// Set the flag if this is a JSON config file
+	config.isJSONConfig = strings.HasSuffix(strings.ToLower(file.Name()), ".json")
```
bendrucker: Declare the empty config earlier so you can just do `config.isJSONConfig = true` in the `switch` statement rather than repeating the extension check.

tflint/config.go
```diff
@@ -477,6 +522,11 @@ func (c *Config) enableBundledPlugin() *Config {
 	return c
 }
 
+// IsJSONConfig returns true if the configuration was loaded from a JSON file
+func (c *Config) IsJSONConfig() bool {
+	return c.isJSONConfig
```
bendrucker: Alternatively, you could just store the config path on the config object, unexported. Or expect it as the first/only key in `sources`. And then call `return filepath.Ext(filename) == ".json"` here

bendrucker (approved):
Nice work here! Couple minor code readability comments, if you have time, it would be helpful. Otherwise I can make those edits.

rorychatterton:
I'm without a computer til the 5th of November. Happy to make the changes on my return, else also cool if you want to make them directly.

wata727 (approved):
👍 

Once you've addressed Ben's comments, we can merge this PR.

bendrucker:
Thank you!

PR #2421, opened 2025-11-26
`version`: add update notifications, json output

### LINKED ISSUE #883: Inform users when a new version is available

bendrucker:
Implements version update notifications similar to Terraform. When running `tflint --version`, users now see a notification if a newer version is available on GitHub Releases.

## Changes

- Adds `versioncheck` package that checks GitHub Releases API for latest version
- Modifies `--version` command to display update notifications when newer version available
- Adds `--format json` support to version command with structured output including update status
- Supports `TFLINT_DISABLE_VERSION_CHECK` environment variable to opt out
- Caches version check results for 48 hours in platform-specific cache directory

## Implementation Details

Version checking happens asynchronously for text output (prints version immediately, shows update after plugins). JSON format performs synchronous check to ensure complete data in single output.

Results are cached in the OS cache directory (`~/Library/Caches/tflint` on macOS, `~/.cache/tflint` on Linux) for 48 hours to minimize API calls.

The check uses GitHub's public API (60 req/hour unauthenticated) or `GITHUB_TOKEN` environment variable for higher rate limits (5000 req/hour).

Errors are logged but non-fatal - network timeouts or API failures don't break the version command.

## JSON Output

JSON output includes `update_check_enabled` field to distinguish between disabled checking vs. no update available:
- `update_check_enabled: false` → checking disabled via env var
- `update_check_enabled: true, update_available: false` → no update or check failed
- `update_check_enabled: true, update_available: true` → update available

## Testing

Verified version checking against live GitHub API with both outdated and current versions. Added unit testing against the internals.

## References

Closes #883

### DIFF
--- a/tflint/config.go
+++ b/tflint/config.go
@@ -160,6 +160,7 @@ func LoadConfig(fs afero.Afero, file string) (*Config, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed to load file: %w", err)
 		}
+		defer f.Close()
 		cfg, err := loadConfig(f)
 		if err != nil {
 			return nil, err
@@ -175,6 +176,7 @@ func LoadConfig(fs afero.Afero, file string) (*Config, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed to load file: %w", err)
 		}
+		defer f.Close()
 		cfg, err := loadConfig(f)
 		if err != nil {
 			return nil, err
@@ -185,6 +187,7 @@ func LoadConfig(fs afero.Afero, file string) (*Config, error) {
 	// Load the default config file (prefer .hcl over .json)
 	log.Printf("[INFO] Load config: %s", defaultConfigFile)
 	if f, err := fs.Open(defaultConfigFile); err == nil {
+		defer f.Close()
 		cfg, err := loadConfig(f)
 		if err != nil {
 			return nil, err
@@ -196,6 +199,7 @@ func LoadConfig(fs afero.Afero, file string) (*Config, error) {
 	// Try JSON config file if HCL not found
 	log.Printf("[INFO] Load config: %s", defaultConfigFileJSON)
 	if f, err := fs.Open(defaultConfigFileJSON); err == nil {
+		defer f.Close()
 		cfg, err := loadConfig(f)
 		if err != nil {
 			return nil, err
@@ -211,6 +215,7 @@ func LoadConfig(fs afero.Afero, file string) (*Config, error) {
 	}
 	log.Printf("[INFO] Load config: %s", fallback)
 	if f, err := fs.Open(fallback); err == nil {
+		defer f.Close()
 		cfg, err := loadConfig(f)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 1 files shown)


versioncheck/github.go
```diff
@@ -0,0 +1,48 @@
...
+		}))
+	}
+	client := github.NewClient(hc)
+
+	log.Printf("[DEBUG] Fetching latest release from GitHub API")
+	release, resp, err := client.Repositories.GetLatestRelease(ctx, repoOwner, repoName)
+	if err != nil {
+		// Check if it's a rate limit error
+		if resp != nil && resp.StatusCode == http.StatusForbidden {
+			if resp.Rate.Remaining == 0 {
```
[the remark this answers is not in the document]
bendrucker: It's a value type, not a pointer

cmd/version.go
```diff
@@ -46,29 +104,62 @@ func (cli *CLI) printVersion(opts Options) int {
 		}
 	}
 
+	// Wait for update check to complete and print notification if available
+	if updateChan != nil {
+		updateInfo := <-updateChan
+		if updateInfo != nil && updateInfo.Available {
+			fmt.Fprintf(cli.outStream, "\nYour version of TFLint is out of date! The latest version\nis %s. You can update by downloading from https://github.com/terraform-linters/tflint/releases\n", updateInfo.Latest)
```
wata727: I feel like the line breaks are a little strange. Doesn't this look more natural?

```suggestion
			fmt.Fprintf(cli.outStream, "\nYour version of TFLint is out of date! The latest version is %s.\nYou can update by downloading from https://github.com/terraform-linters/tflint/releases\n", updateInfo.Latest)
```

cmd/version.go
```diff
@@ -46,29 +104,62 @@ func (cli *CLI) printVersion(opts Options) int {
...
+		}
+	}
+
 	return ExitCodeOK
 }
 
-func getPluginVersions(opts Options) []string {
-	// Load configuration files to print plugin versions
+func (cli *CLI) printVersionJSON(opts Options, updateInfo *versioncheck.UpdateInfo) int {
```
wata727: The `printVersionJSON` does not seem to take `--chdir` or `--recursive` into account.

versioncheck/check_test.go
```diff
@@ -0,0 +1,164 @@
...
+					os.Unsetenv("TFLINT_DISABLE_VERSION_CHECK")
+				} else {
+					os.Setenv("TFLINT_DISABLE_VERSION_CHECK", original)
+				}
+			}()
+
+			if tt.envValue == "" {
+				os.Unsetenv("TFLINT_DISABLE_VERSION_CHECK")
+			} else {
+				os.Setenv("TFLINT_DISABLE_VERSION_CHECK", tt.envValue)
```
wata727: `T.Setenv` is preferable.

wata727 (approved):
👍 

E2E test on Windows is failing, so please check that.

Issue #2460, opened 2026-02-09
Error trying to fetch attestations


bodgit:
### Summary

The latest release seems to have introduced an error when trying to access the attestations API.

I’m running tflint in a private GH enterprise account and we also have an internal tflint plugin hosted in an internal private repo, which has always been verified via PGP, there are no attestations.

Now my tflint GH actions fail with the following:
```
Installing "aws" plugin...
Installed "aws" (source: github.com/terraform-linters/tflint-ruleset-aws, version: 0.45.0)
Installing "<myplugin>" plugin...
Failed to install a plugin; Failed to download artifact attestations: GET https://api.github.com/repos/<org>/<myplugin>/attestations/sha256:…: 403 Resource not accessible by integration []
Error: Process completed with exit code 1.
```
My workflow doesn’t have access to the attestation API and actually I can’t grant it access because of actions/create-github-app-token#310.

Ideally tflint should handle/ignore a 403 error and assume no attestations available.

Downgrading back to 0.60.0 gets things working again.

### Command

tflint —init

### Terraform Configuration

```terraform
n/a
```

### TFLint Configuration

```HCL
plugin "terraform" {
  enabled = true
  preset  = "all"
}

plugin "aws" {
  enabled = true
  version = "0.45.0"
  source  = "github.com/terraform-linters/tflint-ruleset-aws"
}

plugin "myplugin" {
  enabled = true

  version = "1.0.3"
  source  = "github.com/<org>/<myplugin>"

  signing_key = <<-KEY
  -----BEGIN PGP PUBLIC KEY BLOCK-----
  …
  -----END PGP PUBLIC KEY BLOCK-----
  KEY
}
```

### Output

```Shell
Installing "aws" plugin...
Installed "aws" (source: github.com/terraform-linters/tflint-ruleset-aws, version: 0.45.0)
Installing "<myplugin>" plugin...
Failed to install a plugin; Failed to download artifact attestations: GET https://api.github.com/repos/<org>/<myplugin>/attestations/sha256:…: 403 Resource not accessible by integration []
Error: Process completed with exit code 1.
```

### TFLint Version

0.61.0

### Operating System

- [x] Linux
- [ ] macOS
- [ ] Windows

PR #2483, opened 2026-03-16
plugin: Add signature mode to control plugin verifications

wata727:
Follow up of https://github.com/terraform-linters/tflint/issues/2460#issuecomment-3873767636

This PR introduces a new `plugin` attribute, `signature`. The `signature` is an option that controls the verification mode, and valid values ​​are:

- `auto`
  - Prefer `attestation` when available, then fall back to `pgp`. This is the default behavior.
- `attestation`
  - Require artifact attestations. Attestation verification in private repositories is not supported.
- `pgp`
  - Require PGP signature verification with `signing_key`.
- `none`
  - Skip plugin signature verification.

The behavior remains the same as before, and this PR does not change that behavior. Users can now more explicitly control the verification mode.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -49,7 +49,11 @@ func (cli *CLI) init(opts Options) int {
 					_, err = installCfg.Install()
 					if err != nil {
 						if errors.Is(err, plugin.ErrPluginNotVerified) {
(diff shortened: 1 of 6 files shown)


bendrucker (approved):

ricardo-kh:
Hi, is there a plan for releasing this?

**Update:** https://github.com/terraform-linters/tflint/releases/tag/v0.62.0 👍🏻

PR #2509, opened 2026-04-20
feat: remove BUSL-derived Terraform source from embedded compatibility layer

chenrui333:
## Summary

This replaces the BUSL-derived source under `terraform/` with MPL-compatible implementations and ports, then removes the repo-level BUSL metadata.

## What changed

- replace the BUSL-tagged `terraform/` source slices with MPL-compatible implementations and ports
- remove `LICENSE-BUSL` and `terraform/LICENSE-BUSL`
- update `README.md` to describe the project as MPL 2.0 only

## Validation

Passed locally in the TFLint repo:

- `go run ./plugin/stub-generator`
- `git submodule update --init --recursive`
- `env GITHUB_TOKEN="$(gh auth token)" go test ./... -count=1`

Homebrew regression pass also succeeded against a local `homebrew-core` formula patch carrying these three commits:

- `brew reinstall --formula --build-from-source ./Formula/t/tflint.rb`
- `brew test tflint`
- `brew audit --strict tflint`
- `brew linkage --test tflint`
- `brew style Formula/t/tflint.rb`

## Notes

The repo-wide green run depends on a fully prepared checkout:

- generated local plugin fixtures via `plugin/stub-generator`
- initialized fixture submodules
- authenticated GitHub API access for rate-limited plugin release-download tests


chenrui333:
I think we are in the better place to not use the terraform code than couple of years ago

and next might be rename terraform package to opentofu instead, thoughts?

wata727:
Thank you for working on this. However, unfortunately, I cannot accept this PR.

First, the `terraform` package clearly references Terraform's internal package, which is licensed under BUSL, and it cannot be proven that the new implementation does not reference BUSL.

Second, there are concerns about continuity. To keep up with future versions of Terraform, we will probably need to incorporate new BUSL-licensed code again.

The idea of ​​migrating to OpenTofu was discussed in https://github.com/terraform-linters/tflint/discussions/2465, but personally, I think it would be difficult given that TFLint is a "Terraform" linter.

I understand that this is a concern regarding continuity with Homebrew, but unfortunately, as long as we support Terraform, its removal from Homebrew may be unavoidable.

chenrui333:
okay, as long as we understand the results, that would be fine, maybe we can create a homebrew-tap in tflint org or use my tap, what do you think?

wata727:
I think it would be a good idea to create a tap in terraform-linters org.

chenrui333:
Works for me 👍

PR #2513, opened 2026-04-28
Add support for specifying --chdir multiple times

amartani:
Add support for specifying --chdir multiple times. This reuses the current support for running multiple subprocesses in parallel, used by --recursive. With this option, we can now run multiple --chdir without running _all_ --chdir, which is useful, for example, in pre-commit or CI systems on PRs where you may want to run on all modified roots rather than all roots.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ Application Options:
       --var-file=FILE                                           Terraform variable file name
       --var='foo=bar'                                           Set a Terraform variable
       --call-module-type=[all|local|none]                       Types of module to call (default: local)
-      --chdir=DIR                                               Switch to a different working directory before executing the command
+      --chdir=DIR                                               Switch to different working directories before executing the command
       --recursive                                               Run command in each directory recursively
       --filter=FILE                                             Filter issues by file names or globs
       --force                                                   Return zero exit status even if issues found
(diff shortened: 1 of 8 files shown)


wata727:
Since the `--chdir` is not originally intended to be used in multiple times, integrity with other features is important. For example, what happens if you use the `--filter` or `--recursive` with multiple `--chdir` commands?

Furthermore, the validity of allowing multiple `--chdir` needs to be considered. While `--chdir` follows the design of Terraform, Terraform only allows one `--chdir`. Unlike Terraform, is there a design justification for allowing multiple `--chdir`?

amartani:
The justification for multiple `--chdir` is the same as for `--recursive`: you generally want to run `tflint` on many roots (in CI for example), while when running raw `terraform`, you want to be more specific. This implementation is just adding more flexibility to that - instead of _every_ root with `--recursive`, you may want to run only on a subset of roots.

Multiple `--chdir` with `--filter` works in the same way as `--recursive` with `--filter` - each worker independently resolves the filter on their own directory.

Multiple `--chdir` with `--recursive` will run on the union of all the directories under all `--chdir`s.

wata727:
To be honest, I'm torn about whether to accept this proposal. While it seems fine in the short term, I'm not convinced it is the right approach from a long-term design perspective. I plan to consider it a bit further.

amartani:
We ended up implementing this feature as a wrapper script that runs `tflint` processes over a list of folders with its own max-workers control. I think it is fair if you decide this type of functionality is better implemented outside of `tflint`; there may be a bit of extra overhead, but at least in our use case, it is not an issue. Let me know what you decide. Thanks!

PR #2514, opened 2026-04-29
docs: deprecate curl pipe bash installation

### LINKED ISSUE #2510: Deprecate the `curl ... | bash` style installation option
Currently, we propose the `curl ... | bash` style command as the recommended installation option in Linux.

https://github.com/terraform-linters/tflint/blob/2376055e05a48529705050c30e9c994b702e48ec/README.md#L21-L25

However, given the current supply chain risks, this is a bad practice that could potentially allow for the injection of arbitrary bash scripts remotely.

Nevertheless, this style is widely used due to its convenience. It's worth considering a better alternative that balances security and convenience.

At the very least, it's important to avoid adopting this installation option in CI workflows or Dockerfiles. This style should only be used for temporary installations.

dbuaon:
## What

Updates the Linux installation instructions to avoid recommending the `curl ... | bash` pattern as the default installation method.

The README now points users to release archives plus verification, and keeps the install script documented only as a temporary local convenience option.

## Why

Piping remote scripts directly into a shell is convenient, but it is not a good default recommendation from a supply chain security perspective, especially for CI workflows and Dockerfiles.

## Changes

- Replace the default Linux `curl ... | bash` installation example with release archive download + verification + install steps.
- Add a note that piping remote scripts directly into a shell is not recommended for CI workflows or Dockerfiles.
- Keep the install script documented as a temporary local convenience option.

## Validation

- Reviewed the README Markdown locally.
- Ran `git diff --check`.

Fixes #2510

### DIFF
--- a/.github/workflows/install_script.yml
+++ b/.github/workflows/install_script.yml
@@ -50,8 +50,9 @@ jobs:
     - name: Non-existent version
       id: fail
       continue-on-error: true
-      # Install instructions pipe the script from curl to bash
-      # Passing scripts via stdin can have differing behavior from passing as a file arg
+      # Keep coverage for executing the installation script via stdin.
+      # This matches the deprecated convenience pattern and can behave differently
+      # from passing the script as a file argument.
       run: |
         bash < install_linux.sh
         tflint -v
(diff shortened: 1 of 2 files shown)


wata727 (approved):
Thanks!

dbuaon:
You're welcome! Happy to contribute to TFLint.

PR #2521, opened 2026-05-10
Add support for Terraform v1.15

### LINKED ISSUE #2515: Add support for Terraform 1.15
## Introduction

Terraform v1.15 added the following features related to TFLint.

- https://github.com/hashicorp/terraform/issues/38160
- https://github.com/hashicorp/terraform/pull/38217

## Proposal

Update [the embedded Terraform fork](https://github.com/terraform-linters/tflint/tree/v0.62.0/terraform) to v1.15.

Please note that this change involves licensing issue, and therefore, we do not anticipate any external contributions.

## References

- https://github.com/hashicorp/terraform/releases/tag/v1.15.0

wata727:
Fixes https://github.com/terraform-linters/tflint/issues/2515

This PR adds support for Terraform v1.15. This PR primarily includes two new features.

## Dynamic module sources
In Terraform v1.15, variables can now be used for module source and version:

```terraform
module "zoo" {
  source = "./${var.folder}"
}
```

TFLint loads modules and applies rules to module calls, but this requires evaluation of the `source` used to load those modules.

The challenge in supporting this is that the evaluation context and the module are in a circular reference. Creating the evaluation context requires a module variable definition, but loading the module requires the evaluation context.

To solve this issue, we broke down the `LoadConfig` function into `LoadRootModule` and `BuildConfig` functions. Having a root module allows us to generate the evaluation context necessary to load child modules. This is a similar approach to the upstream, but since the upstream implements graph-based module loading instead of module loading via the `config` package, we cannot simply adopt the same approach.

The second challenge is that nested module arguments can be passed as the child module source.

```terraform
module "zoo" {
  source = "./zoo"

  child_source = "./zoo_child"
}

# ./zoo module
module "zoo_child" {
  source = var.child_source
}
```

In such cases, the evaluation context must inherit the one of the parent module.

This can employ the same approach we've already used in [NewModuleRunners](https://github.com/terraform-linters/tflint/blob/v0.62.1/tflint/runner.go#L123-L136). Since the inheritance of the parent context will mostly result in the same implementation, some commonality might be possible, but we haven't included it in this PR scope here.

Furthermore, a TFLint-specific issue is the case where the `source` contains unknown values. Here, we decided to simply skip the module, as with other behaviors (e.g. `count = var.unknown`).

As a side note, the currently provided `ignore_modules` feature matches against the module's `source`, but this requires temporarily caching the evaluated `source` value in the module call. This is an anti-pattern, so in the future, we plan to switch to ignoring based on the module call name.

## `convert` function
The following function calls are now supported:

```terraform
convert([[1, 2, 3]], set(list(number)))
```

An interesting point is that we can pass type definitions to arguments without quotes. This requires a modification to the address parser.


carlcauchi:
any eta for this fix please, to support TF 1.15? @wata727

greising:
Same here, any eta for this fix please, to support TF 1.15? @wata727

Module source as variable is a MUST!!

wata727:
No ETA for now. This project is maintained in my spare time, so I can't promise a timeline.

PR #2523, opened 2026-05-11
terraform: support escaping blocks

### LINKED ISSUE #1644: Add support for Terraform escaping blocks
## Introduction

Since v0.15.4, Terraform introduced the concept of "escaping blocks". 
https://github.com/hashicorp/terraform/pull/28709

Attributes defined inside escaping blocks are merged and treated like normal attributes, except meta-arguments. For example:

```hcl
resource "aws_instance" "main" {
  _ {
    instance_type = "t2.micro" // It works!
  }
}
```

This is a mechanism for future language extensions and is discouraged in most cases, so I believe users do not use it yet. However, if used, TFLint does not support this syntax and cannot detect errors correctly.

## Proposal

Make sure `Module.PartialContent` respects the escaping blocks and returns the fetched content merged.
https://github.com/terraform-linters/tflint/blob/v0.44.1/terraform/module.go#L89

Escaping blocks are only valid at the top level of a particular block, so nested blocks aren't a concern, but we'll probably need logic to identify the block type. This has some performance implications, so we may need to consider how we implement it.

## References

- https://github.com/hashicorp/terraform/pull/28709
- https://github.com/terraform-linters/tflint/blob/v0.44.1/terraform/module.go#L89

Zakharden:
## Summary

Add support for Terraform escaping blocks in `Module.PartialContent` so attributes declared inside `_ { ... }` are exposed like normal block attributes.

This keeps meta-arguments inside escaping blocks out of block expansion by applying the escaping-block schema only when collecting partial content, after the normal expansion step.

Fixes #1644.

## Validation

- `gofmt -w terraform/module.go terraform/module_test.go`
- `go test ./terraform -run TestPartialContent`
- `go test ./terraform -run 'Test(Rebuild|PartialContent|_overrideBlocks)$'`
- `git diff --check`

Note: `go test ./terraform` currently also requires the existing loader test submodule fixtures; without those fixtures, unrelated `TestLoadConfig` cases fail while the focused module tests above pass.

### DIFF
--- a/terraform/module.go
+++ b/terraform/module.go
@@ -133,9 +133,12 @@ func (m *Module) PartialContent(schema *hclext.BodySchema, ctx *Evaluator) (*hcl
 	for _, f := range m.primaries {
 		expanded, d := ctx.ExpandBlock(f.Body, schema)
 		diags = diags.Extend(d)
-		c, d := hclext.PartialContent(expanded, schema)
+		c, d := partialContentWithEscaping(expanded, schema)
 		diags = diags.Extend(d)
 		for name, attr := range c.Attributes {
+			if content.Attributes == nil {
+				content.Attributes = hclext.Attributes{}
+			}
 			content.Attributes[name] = attr
 		}
 		content.Blocks = append(content.Blocks, c.Blocks...)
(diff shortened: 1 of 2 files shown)


wata727:
Thank you for working on this. However, since this is an undocumented feature, it may not be something that should be supported in TFLint at this time.

With that in mind, I will close this PR and issue. Thank you for your understanding.

PR #2522, opened 2026-05-11
formatter: make JUnit testcase names unique

Zakharden:
Fixes #1608.

## Summary

The JUnit formatter previously used only the rule name as each lint issue's testcase name, so multiple failures from the same rule produced duplicate testcase names. This keeps the first testcase name unchanged and appends an incrementing suffix to later occurrences in the sorted issue output, for example:

- `terraform_deprecated_interpolation`
- `terraform_deprecated_interpolation_2`
- `terraform_deprecated_interpolation_3`

This intentionally avoids adding source ranges or resource names to testcase names; detailed range context remains in the failure message/body.

## Validation

- `gofmt -w formatter/junit.go formatter/junit_test.go`
- `go test ./formatter`
- `git diff --check HEAD~1 HEAD`

PR #2538, opened 2026-05-19
formatter(junit): include range in testcase name

Zakharden:
## Why
- JUnit `testcase` names were not unique when multiple issues shared the same rule name, which breaks parsers that deduplicate by name (for example GitLab).
- This caused `tflint --format junit` output to hide duplicate failures in CI result processing.

Closes #2522

## What changed
- Include `Range` in the junit testcase `name` for each issue, producing names like `rule file:line,column-end`.
- Add regression coverage for duplicate rule names with different ranges.

## Verification
- `go test ./formatter`


bendrucker (approved):

bendrucker:
Nice, this seems like the most correct/simple option for a linter to emit uniquely identifiable case names. Let me know if I am mis-reading in having this supersede #2522.

PR #2555, opened 2026-06-05
Extract the shared runner-setup pipeline

bendrucker:
Extracts the load→annotate→vars→build→runners pipeline into a single `tflint.BuildRunners` function and routes both callers through it. The same sequence (`LoadRootModule` → `LoadConfigDirFiles` → `NewAnnotations` → `LoadValuesFiles` → `ParseVariableValues` → `BuildConfig` → `NewRunner` → `NewModuleRunners`) had emerged twice, near-verbatim, in `cmd/inspect.go` (`setupRunners`) and `langserver/handler.go` (`inspect`).

Scope is limited to runner setup. The plugin config-apply loop, `ruleset.Check`, the autofix loop, parallelism, and diagnostics stay in their callers because they differ in real ways (version-constraint checks, the `Fix` flag, parallel module runners, plugin lifecycle, LSP diagnostic conversion). Unifying those is out of scope and would force a different package home to avoid a `tflint → plugin → tflint` import cycle.

## Changes

- Adds `tflint.BuildRunners(loader, config, workingDir, dir)` in the new `tflint/runner_set.go`, lifted from `cmd`'s `setupRunners` body.
- Routes `cmd/inspect.go` and `langserver/handler.go` through it; deletes `setupRunners` and the langserver inline block.
- Keeps `terraform.NewLoader` construction in the callers. `cmd` needs the loader before setup for the worker-mode `IsConfigDir` check and after for `Files()`/`Sources()`; `langserver` builds it from its in-memory `afero.Fs` overlay and reuses it for `Files()` in the `Check` loop. Passing the loader in keeps the OS-fs vs. `CopyOnWriteFs` difference at the caller.

I standardized the diagnostic wording on `cmd`'s semicolon form (`"Failed to load the root module; %w"`), where langserver previously used a colon.

## Testing

Adds `tflint/runner_set_test.go`: a table test that builds a `terraform.Loader` over an in-memory `afero` fixture and asserts module-runner count, variable resolution (auto-loaded and explicit varfiles), and annotation attachment. This pipeline could previously only be exercised through the CLI binary or the LSP integration harness.

### DIFF
--- a/cmd/inspect.go
+++ b/cmd/inspect.go
@@ -104,7 +104,7 @@ func (cli *CLI) inspectModule(opts Options, dir string, filterFiles []string) (t
 	}
 
 	// Setup runners
-	rootRunner, moduleRunners, err := cli.setupRunners(opts, dir)
+	rootRunner, moduleRunners, err := tflint.BuildRunners(cli.loader, cli.config, cli.originalWorkingDir, dir)
 	if err != nil {
 		return issues, changes, err
 	}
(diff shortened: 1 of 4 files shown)


wata727 (approved):
👍

PR #2564, opened 2026-06-11
fix(install): close download and zip readers

RoseSecurity:
## what and why

Group error variables, switch file mode literals to 0o755, and close resources via defer in install/download code.

- Consolidate `ErrPluginNotVerified` and `ErrLegacySigningKeyUsed` into a var block.
- Use `0o755` for `os.MkdirAll` and `os.OpenFile` mode arguments.
- Defer closing of downloader in `downloadToTempFile` and of reader in `extractFileFromZipFile`; remove the explicit Close call.

### DIFF
--- a/plugin/install.go
+++ b/plugin/install.go
@@ -88,8 +88,10 @@ func (c *InstallConfig) CertificateIdentityIssuer() string {
 	return "https://token.actions.githubusercontent.com"
 }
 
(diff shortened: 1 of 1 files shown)


bendrucker:
Please avoid unrelated changes in the future. `0755` to `0o755` is fine as a modernization, as its own PR. Getting unrelated changes in the file along with an AI summary makes it hard to review.

Issue #2544, opened 2026-05-27
Failed to check ruleset; timeout waiting for connection info


vchepkov:
### Summary

We running tflint with aws plugin in gitlab jobs. We observe sporadic failures of tflint, where many jobs are created, so cluster is under heavy load. I added   TFLINT_LOG: "debug" and experimented with multiple options, but none helped to solve the issue:

  PLUGIN_UNIX_SOCKET_DIR: "/dev/shm"
  GOMAXPROCS: "2"
  TF_LINT_ARGS: "--no-parallel-runners"

There seems to be exact 5 seconds before timeout and I wasn't able to find how to change it.

### Command

tflint --chdir=tf/ --no-parallel-runners

### Terraform Configuration

```terraform
# env|grep TF
TF_CLI_CONFIG_FILE=/etc/terraform.tfrc
TFLINT_DISABLE_VERSION_CHECK=true

# cat /etc/terraform.tfrc
provider_installation {
  direct {
    exclude = ["registry.terraform.io/*/*"]
  }
  network_mirror {
    url = "https://artifactory.example.com/artifactory/api/terraform/terraform-virtual/providers/"
  }
}

disable_checkpoint = true
```

### TFLint Configuration

```HCL
bash-5.2# cat ~/.tflint.hcl 
plugin "aws" {
  enabled   = true
  source    = "github.com/terraform-linters/tflint-ruleset-aws"
  version   = "0.47.0"
  signature = "none"
}
plugin "terraform" {
  enabled   = true
  source    = "github.com/terraform-linters/tflint-ruleset-terraform"
  version   = "0.14.1"
  signature = "none"
}
```

### Output

```Shell
$ tflint --version
TFLint version 0.62.1
20:47:23 config.go:204: [INFO] Load config: .tflint.hcl
20:47:23 config.go:213: [INFO] file not found
20:47:23 config.go:216: [INFO] Load config: .tflint.json
20:47:23 config.go:225: [INFO] file not found
20:47:23 config.go:232: [INFO] Load config: /root/.tflint.hcl
20:47:23 config.go:403: [DEBUG] Config loaded
20:47:23 config.go:404: [DEBUG]   CallModuleType: local
20:47:23 config.go:405: [DEBUG]   CallModuleTypeSet: false
20:47:23 config.go:406: [DEBUG]   Force: false
20:47:23 config.go:407: [DEBUG]   ForceSet: false
20:47:23 config.go:408: [DEBUG]   DisabledByDefault: false
[...]

PR #2569, opened 2026-06-19
cmd: bound per-runner check concurrency

bendrucker:
## Summary

Per-runner check parallelism had no concurrency limit. tflint runs each ruleset's `Check` once per runner (the root module plus one runner per module call), and dispatched the module-call checks as one goroutine each with no bound. The only control was the boolean `--no-parallel-runners`, which turns the fan-out fully off. `--max-workers` did not apply. A configuration with many module calls therefore ran that many plugin `Check` RPCs at once, and the only way to reduce the load was to serialize every runner.

This bounds the per-runner fan-out by the `--max-workers` value (default: number of CPUs), so parallelism has a ceiling instead of an on/off switch.

## Motivation

`Check` is plugin code, and a plugin may do network or other rate-limited work per runner. When it does, unbounded fan-out becomes a load problem rather than a speedup.

This surfaced in terraform-linters/tflint-ruleset-aws#1113. With `deep_check` and ~60 `provider "aws"` instances, each runner opened many concurrent STS connections, and the runner count multiplied that into a connection storm that exhausted DNS and sockets. `--max-workers` had no effect. Only `--no-parallel-runners` avoided the errors, at a large speed cost. The plugin side now caches clients across runners tflint-ruleset-aws#1114, but the core fan-out stayed unbounded for any plugin that does per-runner work.

Two existing reports look like the same unbounded concurrency from other angles:

- #2094: sporadic false `circular reference found` errors under parallel runners, gone with `--no-parallel-runners`.
- #2544: `timeout waiting for connection info` under heavy load, with `--no-parallel-runners` tried as a workaround.

## Change

The per-runner loop in `inspectModule` now acquires from a semaphore sized by a worker count before launching each check. The worker count comes from `runnerWorkers()`: `--max-workers` when set, otherwise the CPU count, or 1 under `--no-parallel-runners`. The old `NoParallelRunners` branch collapses into the N=1 case of the same path.

The `--max-workers` resolution that recursive inspection already used moves to a shared `Options.maxWorkers()`, so one definition now governs concurrency on both paths. The flag description now says it bounds both directory and per-runner workers.

`--no-parallel-runners` is unchanged in behavior: it still forces fully serial checks.

### Flag

This reuses `--max-workers` rather than adding a flag, so a single, familiar control governs concurrency everywhere. If per-directory and per-runner concurrency turn out to need independent tuning, a dedicated flag (for example `--max-runner-workers`) could be layered on later without changing this default.

One interaction to note: under `--recursive`, `--max-workers` already bounds the directory subprocess pool, and it is not currently propagated to those subprocesses. So each subprocess resolves its own per-runner bound, and the effective ceiling is roughly `--max-workers` directories times `--max-workers` runners. That is bounded and far below today's unbounded fan-out, but we could also propagate the value to workers.

## Testing

`Test_checkRunners_boundsConcurrency` blocks the checks mid-flight and samples the peak number running at once, asserting it never exceeds the worker count across serial, bounded, and worker-heavy cases. An unbounded implementation fails it. `Test_checkRunners_returnsFirstError` confirms the first error surfaces and every runner is still drained. `Test_Options_runnerWorkers` covers the worker-count resolution, including `--no-parallel-runners` overriding `--max-workers`.

## Origin

Per-runner parallelism landed in #1944 and the recursive `--max-workers` pool in #2021. This reconciles the two so one concurrency model covers both paths.

### DIFF
--- a/cmd/inspect.go
+++ b/cmd/inspect.go
@@ -146,26 +146,16 @@ By setting TFLINT_LOG=trace, you can confirm the changes made by the autofix and
 			if err := ruleset.Check(plugin.NewGRPCServer(rootRunner, rootRunner, cli.loader.Files(), sdkVersions[name])); err != nil {
 				return issues, changes, fmt.Errorf("Failed to check ruleset; %w", err)
 			}
-			// Run checks for module calls are performed in parallel.
+			// Checks for module calls run in parallel, bounded by runnerWorkers
+			// (the --max-workers value, or 1 with --no-parallel-runners).
 			// The rootRunner is shared between goroutines but read-only, so this is goroutine-safe.
 			// Note that checks against the rootRunner are not parallelized, as autofix may cause the module to be rebuilt.
-			ch := make(chan error, len(moduleRunners))
-			for _, runner := range moduleRunners {
-				if opts.NoParallelRunners {
-					ch <- ruleset.Check(plugin.NewGRPCServer(runner, rootRunner, cli.loader.Files(), sdkVersions[name]))
-				} else {
-					go func(runner *tflint.Runner) {
-						ch <- ruleset.Check(plugin.NewGRPCServer(runner, rootRunner, cli.loader.Files(), sdkVersions[name]))
-					}(runner)
-				}
+			err = checkRunners(moduleRunners, opts.runnerWorkers(), func(runner *tflint.Runner) error {
+				return ruleset.Check(plugin.NewGRPCServer(runner, rootRunner, cli.loader.Files(), sdkVersions[name]))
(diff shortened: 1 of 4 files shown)


wata727:
In my experience, if a `Check` call is I/O-bound, setting the number of goroutines to be the same as the number of processes degrades performance.

The reason I didn't set an upper limit on the parallel-runner concurrency is because it depends on the number of module calls. My intuition told me there wouldn't be that many module calls, but it seems I was wrong.

How about starting with a sufficiently large fixed upper limit, like 10 or 20?

PR #2574, opened 2026-06-22
release: generate sigstore bundles rather than certs and signatures

scop:
Made this before noticing the cosign deprecation notice, but thought I'd submit it anyway.

The latter two are deprecated as of cosign 3.1.1,
ref https://github.com/sigstore/cosign/releases/tag/v3.1.1


bendrucker:
Presumably this is a breaking change though? Less appetite for that on a deprecated feature, but it would be fine to do this additively if possible.

PR #2592, opened 2026-07-16
formatter: make JUnit error testcase names unique

Kunalbehbud:
Fixes #1608

### Problem

#2538 made **issue** testcase names unique by including the range, but the **error** testcases appended after them still use the bare diagnostic summary. Multiple diagnostics sharing a summary — e.g. the same `Unsupported argument` error on two different lines — still produce duplicate `<testcase>` names inside the same `<testsuite>`, which JUnit consumers key on.

This is exactly the gap called out in the review of #2522:

> `junitErrors` sets `Name: diag.Summary` directly, so multiple diagnostics sharing a summary (e.g. two "Unsupported argument" errors on different lines) still produce duplicate `<testcase>` names in the same `<testsuite>` (#1608).

and:

> this should cover errors too, then LGTM

### Change

Include the diagnostic range in the error testcase name, mirroring the issue-testcase naming from #2538 (`<summary> <range>`, via the existing nil-safe `diagRange`). The `application_error` singleton case is untouched.

One edge worth noting: two **subject-less** diagnostics with an identical summary would still collide (both synthesize the empty range `:0,0-0`). That mirrors the pre-#2538 state for issues and seems acceptable for a case that can't point at a location — happy to add a positional suffix there if you'd prefer.

### Tests

- Updated the existing "diagnostics without subject" golden output for the new name.
- Added a "diagnostics with the same summary" case: two diagnostics sharing a summary on different lines now produce unique `<testcase>` names.

```
go test ./formatter/ -run Test_junitPrint
```

PR #2593, opened 2026-07-16
plugin: fetch attestation bundles from bundle_url

### LINKED ISSUE #2591: tflint --init panics (nil pointer in bundle.TlogEntries) verifying attestations for tflint-ruleset-aws 0.48.0 when GITHUB_TOKEN is set
### Summary

`tflint --init` crashes with a nil pointer dereference in `sigstore-go`'s `bundle.(*Bundle).TlogEntries` while verifying artifact attestations for `tflint-ruleset-aws` 0.48.0. It only happens when `GITHUB_TOKEN` is set (authenticated client → attestation verification is preferred); without a token, or with `signature = "pgp"`, installation succeeds.

### Command

```console
$ export GITHUB_TOKEN=<any valid token>
$ tflint --init
Installing "aws" plugin...
Panic: runtime error: invalid memory address or nil pointer dereference
 -> 0: main.main.func1: /main.go(25)
 -> 1: runtime.gopanic: /panic.go(860)
 -> 2: runtime.panicmem: /panic.go(336)
 -> 3: runtime.sigpanic: /signal_unix.go(931)
 -> 4: github.com/sigstore/sigstore-go/pkg/bundle.(*Bundle).TlogEntries: /bundle.go(300)
 -> 5: github.com/sigstore/sigstore-go/pkg/verify.VerifyTlogEntry: /tlog.go(40)
 -> 6: github.com/sigstore/sigstore-go/pkg/verify.(*Verifier).VerifyTransparencyLogInclusion: /signed_entity.go(809)
 -> 7: github.com/sigstore/sigstore-go/pkg/verify.(*Verifier).Verify: /signed_entity.go(606)
 -> 8: github.com/terraform-linters/tflint/plugin.(*SignatureChecker).VerifyAttestations: /signature.go(136)
 -> 9: github.com/terraform-linters/tflint/plugin.(*InstallConfig).Install: /install.go(134)
 -> 10: github.com/terraform-linters/tflint/cmd.(*CLI).init.func1: /init.go(49)
 -> 11: github.com/terraform-linters/tflint/cmd.(*CLI).withinChangedDir: /cli.go(200)
 -> 12: github.com/terraform-linters/tflint/cmd.(*CLI).init: /init.go(23)
 -> 13: github.com/terraform-linters/tflint/cmd.(*CLI).Run: /cli.go(108)
 -> 14: main.main: /main.go(38)
 -> 15: runtime.main: /proc.go(290)
 -> 16: runtime.goexit: /asm_amd64.s(1771)

TFLint crashed... :(
```

### Terraform Configuration

Not relevant — the crash happens during plugin installation, before any configuration is inspected.

### TFLint Configuration

```hcl
plugin "terraform" {
  enabled = true
  preset  = "recommended"
}

plugin "aws" {
  enabled = true
  version = "0.48.0"
  source  = "github.com/terraform-linters/tflint-ruleset-aws"
}
```

### Version

Reproduced on both v0.62.1 and v0.63.1:

```console
$ tflint --version
TFLint version 0.63.1
+ ruleset.terraform (0.15.0-bundled)
```

Environments: GitHub Actions `ubuntu-latest` (linux_amd64, installed via setup-tflint v6.3.0) and macOS arm64 (release binary). Same panic on both.

### Additional context

- Without `GITHUB_TOKEN`, `tflint --init` installs the plugin fine (PGP fallback).
- With `GITHUB_TOKEN` and `signature = "pgp"` in the plugin block, installation also works (with the legacy-signing-key deprecation warning), so this looks specific to the attestation bundles on the `tflint-ruleset-aws` v0.48.0 release rather than the download path.
- `verifier.Verify` is reached with a bundle whose tlog entries appear to be nil — possibly a malformed/partial attestation bundle on that release; a nil check with a graceful error (like the `failed to unmarshal sigstore bundle` path) would at least avoid the crash.

Kunalbehbud:
Fixes #2591

## Root cause

GitHub stopped embedding sigstore bundles in attestation list responses. As @sarahhodne noted in the issue, this matches the documented [breaking change](https://docs.github.com/en/rest/about-the-rest-api/breaking-changes) ("Remove the bundle property from attestation list responses. ... Use `bundle_url` to retrieve the attestation bundle"), except it now applies to the `2022-11-28` API version as well. The endpoint currently returns:

```json
{
  "attestations": [
    {
      "repository_id": 245765716,
      "bundle_url": "https://tmaproduction.blob.core.windows.net/attestations/...json.sn?...",
      "initiator": "user",
      "bundle": null
    }
  ]
}
```

`json.Unmarshal` unmarshals `null` into a `*bundle.Bundle` without error — it just sets the pointer to `nil` — and `VerifyAttestations` passes that `nil` straight to `verifier.Verify`, which dereferences it in `bundle.(*Bundle).TlogEntries`. That's the panic in the issue. It also reproduces in this repo's CI: the `build` workflow on #2592 hit the same panic in `Test_Install_withAttestations` ([run](https://github.com/terraform-linters/tflint/actions/runs/29538159046/job/87754264210)), and the nightly on master will hit it too.

## Changes

1. **`fetchArtifactAttestations` resolves `bundle_url`.** go-github v81 doesn't expose the field, so the endpoint is queried via `client.NewRequest`/`client.Do` with a small local struct — this keeps `github.ErrorResponse` semantics, so `isIgnorableAttestationError` keeps working for 403/404. When `bundle` is null or absent, the bundle is downloaded from `bundle_url`. The blob is served as snappy-compressed JSON (`Content-Type: application/x-snappy`), so it's snappy-decoded, mirroring what the gh CLI does in [`getBundle`](https://github.com/cli/cli/blob/trunk/pkg/cmd/attestation/api/client.go). The download intentionally uses `http.DefaultClient`: the URL is pre-signed, and API credentials must not be sent to blob storage. If a bundle is ever served uncompressed, the raw body is used as long as it's valid JSON.

2. **`VerifyAttestations` rejects empty bundles instead of crashing.** Bundles are now parsed upfront, and a null bundle returns `attestation contains an empty sigstore bundle`. Even if the API serves something unexpected again, `tflint --init` fails with an error rather than a SIGSEGV.

Uses `github.com/klauspost/compress/snappy`, which is already in the module graph as an indirect dependency, rather than adding a new module.

## Tests

- `Test_fetchArtifactAttestations` (new, offline via httptest): embedded-bundle passthrough, `bundle_url` + snappy, `bundle_url` + uncompressed fallback, null bundle with no URL, download failure.
- `Test_SignatureChecker_VerifyAttestations_errors`: new `null bundle` case — the exact crash input now returns an error.
- `Test_Install_withAttestations` (existing, live): panicked before this change exactly as reported; now passes end-to-end against the real API (list → download from `bundle_url` → decompress → verify).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 	github.com/hashicorp/logutils v1.0.0
 	github.com/jessevdk/go-flags v1.6.1
 	github.com/jstemmer/go-junit-report v1.0.0
+	github.com/klauspost/compress v1.18.6
 	github.com/mattn/go-colorable v0.1.15
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/owenrumney/go-sarif/v2 v2.3.3
@@ -125,7 +126,6 @@ require (
 	github.com/hashicorp/yamux v0.1.2 // indirect
 	github.com/in-toto/attestation v1.2.0 // indirect
 	github.com/in-toto/in-toto-golang v0.11.0 // indirect
-	github.com/klauspost/compress v1.18.6 // indirect
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mitchellh/go-wordwrap v1.0.1 // indirect
 	github.com/oklog/run v1.1.0 // indirect
--- a/plugin/install.go
+++ b/plugin/install.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/hex"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -18,6 +19,7 @@ import (
 	"strings"
 
 	"github.com/google/go-github/v81/github"
+	"github.com/klauspost/compress/snappy"
 	"github.com/terraform-linters/tflint/tflint"
 	"golang.org/x/net/idna"
 	"golang.org/x/oauth2"
@@ -296,7 +298,22 @@ func (c *InstallConfig) fetchReleaseAssets(ctx context.Context, client *github.C
 	return assets, nil
 }
 
+// attestationsResponse is like github.AttestationsResponse, but also captures
+// the bundle_url field, which go-github does not expose. GitHub may serve the
+// sigstore bundle from blob storage instead of embedding it in the response.
+type attestationsResponse struct {
+	Attestations []*attestation `json:"attestations"`
+}
+
+type attestation struct {
+	Bundle       json.RawMessage `json:"bundle"`
+	BundleURL    string          `json:"bundle_url"`
+	RepositoryID int64           `json:"repository_id"`
+}
+
 // fetchArtifactAttestations fetches GitHub Artifact Attestations based on the given artifact.
+// Attestations whose sigstore bundle is not embedded in the API response are
+// resolved by downloading the bundle from the URL in the bundle_url field.
 func (c *InstallConfig) fetchArtifactAttestations(ctx context.Context, client *github.Client, artifact []byte) ([]*github.Attestation, error) {
 	hash := sha256.New()
 	if _, err := hash.Write(artifact); err != nil {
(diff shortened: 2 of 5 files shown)


wata727:
Thank you for working on this. It looks good to me.

As this issue involves a significant bug fix, it was merged as #2600 to ensure a prompt release. Therefore, I am closing this PR. Thank you for your cooperation.

PR #2636, opened 2026-08-28
Let an override file expand a resource into more instances

### LINKED ISSUE #2151: for_each/count meta-arguments are not overridden correctly
### Summary

See also https://github.com/terraform-linters/tflint/issues/2114

Many of the overriding issues in TFLint have been fixed in https://github.com/terraform-linters/tflint/pull/2124, but there are still issues with overriding meta arguments. In Terraform, meta arguments are overridden and then resources are expanded, but in TFLint, they are overridden after expansion, so if the original and override meta arguments are different, the override expansion cannot be followed.

### Command

tflint

### Terraform Configuration

```terraform
# main.tf
resource "aws_instance" "main" {
  instance_type = "t2.micro"
}

# main_override.tf
resource "aws_instance" "main" {
  count = 2

  instance_type = "t${count.index + 1}.invalid"
}
```

### TFLint Configuration

```HCL
plugin "terraform" {
  enabled = false
}

plugin "aws" {
  enabled = true
  version = "0.34.0"
  source  = "github.com/terraform-linters/tflint-ruleset-aws"
}
```

### Output

```Shell
1 issue(s) found:

Error: "t2.invalid" is an invalid value as instance_type (aws_instance_invalid_type)

  on main_override.tf line 4:
   4:   instance_type = "t${count.index + 1}.invalid"
```

### TFLint Version

0.53.0

### Terraform Version

1.9.8

### Operating System

- [X] Linux
- [ ] macOS
- [ ] Windows

Eljees:
Closes #2151.

### The problem

Terraform merges override files into the primary configuration first and
expands `count` / `for_each` afterwards. TFLint does it the other way round:
it expands first and merges afterwards. So when an override file adds
`count` to a resource that the primary declares once, TFLint arrives at
`overrideBlocks` with several expanded override blocks and a single primary,
and merges all of them into that one primary. Every instance but the last is
lost.

Using the configuration from the issue:

```hcl
# main.tf
resource "aws_instance" "main" {
  instance_type = "t2.micro"
}

# main_override.tf
resource "aws_instance" "main" {
  count = 2

  instance_type = "t${count.index + 1}.invalid"
}
```

two instances should be reported and only one is.

### The change

`overrideBlocks` now tracks, per resource address, the run of override blocks
that share a `DefRange`, i.e. the ones that came from expanding one
declaration. The first instance of a run merges into the primary exactly as
before. Each extra instance clones a pristine, pre-merge copy of that primary
so it gets a block of its own instead of overwriting an already-merged one.

An override block with a different `DefRange` is an independent declaration
overriding the same address again. Duplicate resource blocks are not valid
Terraform, but TFLint tolerates them, so that case keeps the previous
behaviour of merging into the first primary.

### Tests

On `./terraform/`:

| | passed | failed |
|---|---|---|
| master | 204 | 2 |
| new test, without the fix | 204 | 3 |
| new test, with the fix | 205 | 2 |

The two constant failures are `TestLoadConfig` and `TestLoadConfig_withBaseDir`,
which fail the same way on an untouched checkout here because they want a
downloaded remote module.

Without the fix the new test fails as

```
module_test.go:680: PartialContent() returned 1 block(s) for a resource
overridden with count = 2, want 2 (see terraform-linters/tflint#2151)
```

The test does not stop at the count. It also reads `instance_type` back out of
each returned block and compares the two values, because `count.index` makes
them differ. I checked that this matters by mutating the fix in both
directions: returning the same block twice instead of a copy, and dropping the
extra instance altogether. Both mutants fail the new test; reverting either
one makes it pass again.

`gofmt -l` is clean and `go vet ./terraform/` passes.

One note on the wider suite: `go test ./...` also fails `integrationtest/*`
and a few plugin-dependent packages here, identically on an untouched
checkout, because this machine hits `403 API rate limit exceeded` when the
tests try to download `tflint-ruleset-aws`. I ran those packages with
`-count=1` on both trees to confirm they fail the same way with and without
the change.

### DIFF
--- a/terraform/module.go
+++ b/terraform/module.go
@@ -189,14 +189,56 @@ func overrideBlocks(primaries, overrides hclext.Blocks) hclext.Blocks {
 	// please note that the block to be overwritten cannot be uniquely determined.
 	newPrimaries := hclext.Blocks{}
 
+	// A resource override file can expand (via count/for_each) a single declaration into
+	// more instances than the primary configuration has for that address. Track, per
+	// address, the run of override blocks that share a DefRange (i.e. come from expanding
+	// the same declaration): the first instance in a run merges into the first primary as
+	// before, and any extra instance clones a pristine (pre-merge) copy of that primary
+	// instead of clobbering an already-merged block. An override block with a different
+	// DefRange is an independent declaration overriding the same address again (duplicate
+	// resource blocks aren't valid Terraform, but tflint tolerates them), and keeps the
+	// pre-existing behavior of clobbering the first primary.
+	// See https://github.com/terraform-linters/tflint/issues/2151
+	type resourceOverrideRun struct {
+		defRange hcl.Range
+		pristine *hclext.Block
+		next     int
+	}
+	resourceOverrideRuns := map[string]*resourceOverrideRun{}
+
 	for _, override := range overrides {
 		addr := blockAddr(override)
 
 		switch override.Type {
 		case "resource":
 			if primaries, exists := overridesByAddr[addr]; exists {
-				// Duplicate resource blocks are not allowed.
-				overrideResourceBlock(primaries[0], override)
+				run := resourceOverrideRuns[addr]
+				if run == nil || run.defRange != override.DefRange {
+					// Duplicate resource blocks are not allowed, so normally there is
+					// exactly one primary per address and every override merges into
+					// it. Start (or restart, for an independent declaration) a run at
+					// instance 0, applied directly exactly like before.
+					run = &resourceOverrideRun{defRange: override.DefRange, pristine: primaries[0].Copy(), next: 1}
+					resourceOverrideRuns[addr] = run
+					overrideResourceBlock(primaries[0], override)
+				} else {
+					idx := run.next
+					run.next++
+					if idx < len(primaries) {
+						overrideResourceBlock(primaries[idx], override)
+					} else {
+						// This run's override file expanded the same declaration
+						// into more instances than the primary configuration has
+						// for this address; clone the pristine (pre-merge) primary
+						// so this instance merges into its own block instead of
+						// clobbering an already-merged one.
+						clone := run.pristine.Copy()
+						overrideResourceBlock(clone, override)
+						overridesByAddr[addr] = append(overridesByAddr[addr], clone)
+						primaries = overridesByAddr[addr]
+						newPrimaries = append(newPrimaries, clone)
+					}
+				}
 			}
 
 		// The "data" block is the same as generic block except for "depends_on".
--- a/terraform/module.go
+++ b/terraform/module.go
@@ -167,6 +167,20 @@ func blockAddr(b *hclext.Block) string {
 	return b.Type
 }
 
+// deepCopyBlock copies a block including its nested blocks. hclext's Copy()
+// deep-copies attributes but shares the nested block pointers, which is not
+// enough on the clone path below: overrideResourceBlock merges the arguments of
+// a lifecycle block into the existing block in place, so two blocks sharing one
+// lifecycle block would both end up with the arguments of whichever override
+// instance was merged last.
+func deepCopyBlock(block *hclext.Block) *hclext.Block {
+	out := block.Copy()
+	for i, nested := range out.Body.Blocks {
+		out.Body.Blocks[i] = deepCopyBlock(nested)
+	}
+	return out
+}
+
 // overrideBlocks overrides the primary blocks passed with override blocks,
 // following Terraform's merge behavior.
 // https://developer.hashicorp.com/terraform/language/files/override#merging-behavior
@@ -218,7 +232,7 @@ func overrideBlocks(primaries, overrides hclext.Blocks) hclext.Blocks {
 					// exactly one primary per address and every override merges into
 					// it. Start (or restart, for an independent declaration) a run at
 					// instance 0, applied directly exactly like before.
-					run = &resourceOverrideRun{defRange: override.DefRange, pristine: primaries[0].Copy(), next: 1}
+					run = &resourceOverrideRun{defRange: override.DefRange, pristine: deepCopyBlock(primaries[0]), next: 1}
 					resourceOverrideRuns[addr] = run
 					overrideResourceBlock(primaries[0], override)
 				} else {
@@ -232,7 +246,7 @@ func overrideBlocks(primaries, overrides hclext.Blocks) hclext.Blocks {
 						// for this address; clone the pristine (pre-merge) primary
 						// so this instance merges into its own block instead of
 						// clobbering an already-merged one.
-						clone := run.pristine.Copy()
+						clone := deepCopyBlock(run.pristine)
 						overrideResourceBlock(clone, override)
 						overridesByAddr[addr] = append(overridesByAddr[addr], clone)
 						primaries = overridesByAddr[addr]
--- a/terraform/module_test.go
+++ b/terraform/module_test.go
@@ -708,6 +708,112 @@ resource "aws_instance" "foo" {
 	}
 }
 
+func TestPartialContent_OverrideChangesInstanceCountWithLifecycle(t *testing.T) {
+	files := map[string]string{
+		"main.tf": `
+resource "aws_instance" "foo" {
+  instance_type = "t2.micro"
+
+  lifecycle {
+    create_before_destroy = false
+  }
+}`,
+
+		"main_override.tf": `
+resource "aws_instance" "foo" {
+  count = 3
+
+  lifecycle {
+    create_before_destroy = count.index == 1
+  }
+}`,
+	}
+
+	fs := afero.Afero{Fs: afero.NewMemMapFs()}
+	for name, content := range files {
+		if err := fs.WriteFile(name, []byte(content), os.ModePerm); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	parser := NewParser(fs)
+	mod, diags := parser.LoadConfigDir(".", ".")
+	if diags.HasErrors() {
+		t.Fatal(diags)
+	}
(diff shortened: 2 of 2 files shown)


Eljees:
Self-review note before this gets a maintainer's time: I think the clone path here still shares nested blocks with the primary, so the `lifecycle` case can reproduce the very bug this PR fixes.

`hclext.BodyContent.Copy()` deep-copies attributes but shallow-copies nested blocks:

```go
out := &BodyContent{
    Attributes: Attributes{},
    Blocks:     make(Blocks, len(b.Blocks)),
}
for k, v := range b.Attributes {
    out.Attributes[k] = v.Copy()
}
copy(out.Blocks, b.Blocks)   // copies the *Block pointers
```

So `clone.Body.Blocks[i] == run.pristine.Body.Blocks[i] == primaries[0].Body.Blocks[i]` — all three are the same object.

That is fine for attributes, because `overrideResourceBlock` writes into `primary.Body.Attributes`, which the clone owns. But its `lifecycle` branch mutates the nested block in place:

```go
if p.Type == "lifecycle" {
    for _, override := range overrides {
        for name, attr := range override.Body.Attributes {
            p.Body.Attributes[name] = attr
        }
    }
    return true
}
```

`p` comes from `primary.Body.Blocks`, so for a clone it is the primary's own `lifecycle` block. With a primary that has a `lifecycle` block and an override that expands the resource and also carries `lifecycle`, instance 0 merges into that block directly and every extra instance then writes into the same object — so all instances end up showing the last instance's lifecycle arguments. That is the same "one instance's override applied to all" failure the new test guards against, just one level down.

The fix is presumably to deep-copy nested blocks on the clone path rather than relying on `Copy()`, plus a test with `lifecycle` on both sides.

I have not been able to run this: the repo needs Go 1.26.5 and I could not obtain that toolchain in my environment, so the above is from reading `tflint-plugin-sdk` v0.25.0 `hclext/structure.go` and `terraform/module.go`, not from a failing test. Worth confirming with an actual case before acting on it.

Eljees:
Confirmed, and fixed in the commit above.

I could run it here, so for the record: `Block.Copy()` really does share nested blocks. A probe over `hclext` v0.25.0 shows the clone's `Body` is a fresh object but `Body.Blocks[0]` is the same pointer as the original's, which is what `copy(out.Blocks, b.Blocks)` gives.

The reproducing case is a primary with a `lifecycle` block and an override that adds `count` and its own `lifecycle`:

```
[]bool{
  	false,
- 	true,
+ 	false,
}
(instances showing the same value share one lifecycle block)
```

Both instances came out with the last-merged value instead of one each, exactly as described.

The fix copies nested blocks recursively on the clone path. The lifecycle case in the test now uses three instances rather than two, because two only exercises a clone sharing with the primary; three also covers two clones sharing with each other.

One honest note on how far the test reaches. Reverting the clone to the shallow `Copy()` makes it fail:

```
--- FAIL: TestPartialContent_OverrideChangesInstanceCountWithLifecycle
```

Reverting only the pristine snapshot to a shallow copy does not. That path is not reachable through `count` expansion, since every expanded instance carries the same argument names and overwrites whatever the first merge left behind. I kept the deep copy there anyway so that the snapshot is genuinely pre-merge rather than relying on that argument, but it is not pinned by a test and I would drop it if you would rather not carry unexercised code.

`./terraform/` is at 206 passed, with `TestLoadConfig` and `TestLoadConfig_withBaseDir` failing the same way on an untouched checkout here because they want a downloaded remote module.

bendrucker:
Citations to Terraform code would be appreciated here as a way to verify ordering is correct and try to keep the implementations in sync. Approach seems reasonable.

Eljees:
Citations below are pinned to `hashicorp/terraform@9edee031a76e05a623a99873e2fd160b4385d08f` and to tflint `master@5aaf04b`.

**Ordering.** `configs.NewModule` appends every primary file first, then merges the override files, in list order: [module.go#L177-L185](https://github.com/hashicorp/terraform/blob/9edee031a76e05a623a99873e2fd160b4385d08f/internal/configs/module.go#L177-L185). The two lists come from `Parser.LoadConfigDir`, which loads `fileSet.Primary` and then `fileSet.Override`: [parser_config_dir.go#L55-L62](https://github.com/hashicorp/terraform/blob/9edee031a76e05a623a99873e2fd160b4385d08f/internal/configs/parser_config_dir.go#L55-L62). `overrideBlocks` iterates `overrides` in the same order, so the ordering this PR relies on is the upstream one.

**One config object per address ΓÇö this is where the two implementations legitimately differ.** Terraform keeps resources in `map[string]*Resource` keyed by `moduleUniqueKey()` ([module.go#L132](https://github.com/hashicorp/terraform/blob/9edee031a76e05a623a99873e2fd160b4385d08f/internal/configs/module.go#L132)), so `mergeFile` finds exactly one primary per address and merges into it ([module.go#L869-L883](https://github.com/hashicorp/terraform/blob/9edee031a76e05a623a99873e2fd160b4385d08f/internal/configs/module.go#L869-L883)). `count` and `for_each` are carried across as unevaluated expressions ([module_merge.go#L231-L236](https://github.com/hashicorp/terraform/blob/9edee031a76e05a623a99873e2fd160b4385d08f/internal/configs/module_merge.go#L231-L236)) and expansion happens later, during evaluation. Terraform therefore **merges once, then expands**.

tflint expands during loading, so `overridesByAddr[addr]` holds already-expanded instances. To land where Terraform lands, every instance the override adds has to merge into a *pristine* copy of the primary rather than into an instance that has already been merged ΓÇö merge-then-expand and expand-then-merge agree only when each expansion starts from the same base. That is what the `pristine` copy and the run tracking in this PR are for.

**Attributes.** Override attributes clobber base attributes of the same name: `mergeBody.prepareContent` [module_merge_body.go#L81-L88](https://github.com/hashicorp/terraform/blob/9edee031a76e05a623a99873e2fd160b4385d08f/internal/configs/module_merge_body.go#L81-L88). `overrideResourceBlock` L244-L246 matches this.

**Nested blocks.** "If any block of a particular type appears in `override` then it will replace _all_ of the blocks of the same type in `base`", and "We explicitly do not try to correlate and deeply merge nested blocks": [module_merge_body.go#L15-L17](https://github.com/hashicorp/terraform/blob/9edee031a76e05a623a99873e2fd160b4385d08f/internal/configs/module_merge_body.go#L15-L17) and [L90-L118](https://github.com/hashicorp/terraform/blob/9edee031a76e05a623a99873e2fd160b4385d08f/internal/configs/module_merge_body.go#L90-L118). `overrideResourceBlock` L252-L278 matches this.

**`lifecycle`.** Terraform never merges the `lifecycle` block as a block. It decodes it and merges the decoded fields one at a time ΓÇö `Connection`, `CreateBeforeDestroy`, `IgnoreChanges`, `IgnoreAllChanges`, `PreventDestroy`, `Provisioners`, `ActionTriggers`: [module_merge.go#L249-L275](https://github.com/hashicorp/terraform/blob/9edee031a76e05a623a99873e2fd160b4385d08f/internal/configs/module_merge.go#L249-L275). That is the "merged on an argument-by-argument basis" wording from the docs.

tflint reaches the same observable result by copying attributes into the primary's `lifecycle` block in place (L262-L270), and that mechanical difference is exactly why the aliasing this PR fixes is possible here and impossible upstream: Terraform mutates the decoded struct hanging off one `*Resource`, while tflint mutates an `*hclext.Block` that `Copy()` shares between the original and the clone. Hence `deepCopyBlock`.

**One case with no upstream counterpart.** Two *independent* declarations overriding the same address cannot arise in Terraform, since duplicate `resource` blocks are invalid there. The patch keeps tflint's existing behaviour for that case (merge into `primaries[0]`) and changes only the expanded-run case, which is why it keys the run on `DefRange`.

---

One more link, since it is the other half of the same point. `count`/`for_each` are evaluated during the graph walk, against the already-merged config: [node_resource_abstract.go#L507-L522](https://github.com/hashicorp/terraform/blob/9edee031a76e05a623a99873e2fd160b4385d08f/internal/terraform/node_resource_abstract.go#L507-L522), where `n.Config` is the `*configs.Resource` that already went through `NewModule`. So merge is static, in the `configs` package, and expansion is dynamic, in the `terraform` package, operating on the merged config.

(Apologies for the near-duplicate comment a moment ago â€” I have folded it in here and removed it.)

Eljees:
Citations, from Terraform's own source (`hashicorp/terraform`, current `main`):

1. Override merge happens during static config loading, before any resource-instance
expansion is even possible. `configs.NewModule()` runs `appendFile()` for every primary
file (this is where each `*Resource`'s `Count`/`ForEach` fields get set to the raw,
unevaluated HCL expression from the primary file) to completion before it starts
`mergeFile()` for the override files:
https://github.com/hashicorp/terraform/blob/main/internal/configs/module.go

2. The override itself, in `Resource.merge()`, is a plain expression swap, not an
expansion:
https://github.com/hashicorp/terraform/blob/main/internal/configs/module_merge.go
```go
if or.Count != nil {
    r.Count = or.Count
}
if or.ForEach != nil {
    r.ForEach = or.ForEach
}
```
`r.Count`/`r.ForEach` are `hcl.Expression`, not values — this is purely a config-tree
operation.

3. `count`/`for_each` get evaluated later, during the graph walk, against the
already-merged config:
https://github.com/hashicorp/terraform/blob/main/internal/terraform/node_resource_abstract.go
```go
case n.Config != nil && n.Config.Count != nil:
    count, countDiags := evaluateCountExpression(n.Config.Count, ctx, allowUnknown)
case n.Config != nil && n.Config.ForEach != nil:
    forEach, known, forEachDiags := evaluateForEachExpression(n.Config.ForEach, ctx, allowUnknown)
```
`n.Config` is the `*configs.Resource` that already went through `NewModule()`'s merge
step, so by the time count/for_each are evaluated, the override's expression has already
replaced the primary's.

So: merge is static (config-tree level, `configs` package), expansion is dynamic
(graph-walk level, `terraform` package, operating on the merged config). That matches what
this PR assumes.

PR #2642, opened 2026-09-06
langserver: log request params only at trace level

### LINKED ISSUE #2639: Language server logs full Terraform source for every didOpen and didChange notification
### Summary

`tflint --langserver` logs the complete parameters of every LSP request and notification to stderr, including the full document text from `textDocument/didOpen` and `textDocument/didChange`.

Because the server uses full document synchronization, every edit logs the entire Terraform file. This creates significant log volume and may expose credentials or other sensitive configuration in editor logs.

Reproduced with Neovim 0.12.5 without enabling `TFLINT_LOG`.

Steps:

1. Start TFLint as a language server.
2. Open the configuration below.
3. Edit the string value.
4. Inspect the language server's stderr output.

Expected: document contents are omitted or only logged when explicit debug/trace logging is enabled.

Actual: the complete document is logged after every edit.

Related to the general LSP umbrella issue #1356, but that issue does not mention logging.

### Command

tflint --langserver

### Terraform Configuration

```terraform
locals {
  api_token = "example-sensitive-value"
}
```

### TFLint Configuration

```HCL
TFLint defaults
```

### Output

```Shell
handler.go:117: Received textDocument/didChange with {"contentChanges":[{"text":"locals {\n  api_token = \"example-sensitive-value\"\n}\n"}],"textDocument":{"uri":"file:///tmp/tflint-lsp-repro/main.tf","version":2}}
```

### TFLint Version

0.61.0

### Terraform Version

1.16.0

### Operating System

- [ ] Linux
- [x] macOS
- [ ] Windows

chernetskyi:
## Summary

Fixes #2639.

The language server previously logged every request and notification together with its complete JSON parameters. Since document synchronization includes the full Terraform source, `didOpen` and `didChange` could expose configuration values and generate excessive editor logs.

This change:

- continues logging the LSP method name normally;
- moves request parameters to a `[TRACE]`-prefixed message;
- adds tests for unset, debug, and trace logging levels.

## Why the prefix matters

TFLint uses `logutils.LevelFilter`, which determines a message's level from a prefix. Unprefixed messages are not assigned a level and therefore pass through the filter even when `TFLINT_LOG` is unset. This is why the previous parameter log was emitted by default. 

Prefixing the parameter message with `[TRACE]` makes the existing filter suppress it unless `TFLINT_LOG=trace` is explicitly enabled. The method name remains unprefixed so ordinary logs still show which LSP messages were received without including their payloads.


bendrucker:
Thanks!

PR #2647, opened 2026-09-15
fix: replace deprecated postflight hook with postflight_steps in Homebrew cask

na4ma4:
This pull request refactors the Homebrew cask configuration in `.goreleaser.yml` to fix the deprecation notice. The change is replacing the previous `hooks` approach with a `custom_block` using `postflight_steps`, which is required for Homebrew v7.

**Homebrew cask configuration improvements:**

* Replaced the deprecated `hooks` and `system_command` block with a `custom_block` that defines `postflight_steps` to remove the quarantine attribute from the installed `tflint` binary on macOS. This makes the configuration align with modern Homebrew requirements.


na4ma4:
When I did a test release locally it generates the attached homebrew package.

[terraform-linters-tflint.rb.txt](https://github.com/user-attachments/files/32218364/terraform-linters-tflint.rb.txt)

sodejm (commented):
Fixes my homebrew error locally

.goreleaser.yml
```diff
@@ -49,9 +49,9 @@ homebrew_casks:
...
-    hooks:
-      post:
-        install: |
-          if OS.mac?
-            system_command "/usr/bin/xattr", args: ["-dr", "com.apple.quarantine", "#{staged_path}/tflint"]
-          end
+    custom_block: |
+      postflight_steps do
+        on_macos do
+          run "/usr/bin/xattr", args: ["-dr", "com.apple.quarantine", "#{staged_path}/tflint"]
```
kgaughan, line 55: This isn't quite what you want to do. Referring to the [Homebrew 7 migration guide](https://brew.sh/7.0.0-migration-guide/#install-step-names-and-options) and [cask cookbook](https://docs.brew.sh/Cask-Cookbook#interpolation-in-steps-blocks), this is somewhere `{{...}}` should be used for interpolation rather than regular Ruby string interpolation with `#{...}` so it gets deferred until installation.

There's an extra complication due to Go's test/template using the same format. This is what should be used here, though it looks a bit awkward:

```suggestion
          run "/usr/bin/xattr", args: ["-dr", "com.apple.quarantine", "{{"{{staged_path}}"}}/tflint"]
```

bendrucker: Thanks for this. I was hoping to not do anything other than upgrade GoReleaser but since that fix is still pending I'm revisiting this PR. @na4ma4 if you can fix this and also think a bit about how to validate. Sounds like the existing syntax worked though?

kgaughan: It works, but kind of by accident. As the docs say:

> A concrete result is safe only when it is identical for every installation represented by the JSON.

So in the interest of avoiding surprises, I think it's best to get this included.

PR #2648, opened 2026-09-15
langserver: publish transient HCL diagnostics instead of failing didChange

### LINKED ISSUE #2640: Language server treats transient invalid didChange content as a JSON-RPC handler error
### Summary

When a document is temporarily invalid during normal typing, `tflint --langserver` returns an error from the `textDocument/didChange` notification handler. The JSON-RPC layer consequently logs it as a notification handling error.

Transient syntax errors and incomplete references are unavoidable while editing. The language server should treat them as normal document states—either publishing parser diagnostics or skipping linting until the document is valid—rather than reporting an internal handler failure for every keystroke.

Steps:

1. Start TFLint as a language server.
2. Open the valid configuration below.
3. Temporarily remove the value after `example =`, as commonly happens while editing.
4. Inspect stderr.

Expected: no JSON-RPC handler error is logged for the transient invalid state.

Actual: every `didChange` received while the document is invalid produces a handler error.

Related to the general LSP umbrella issue #1356, but that issue does not cover transient edit handling.

### Command

tflint --langserver

### Terraform Configuration

```terraform
locals {
  example = "value"
}
```

### TFLint Configuration

```HCL
TFLint defaults
```

### Output

```Shell
jsonrpc2 handler: notification "textDocument/didChange" handling error: Failed to load configurations: main.tf:2,12-3,1: Missing expression; Expected the start of an expression, but found an invalid expression token.


Incomplete references similarly produce errors such as:


jsonrpc2 handler: notification "textDocument/didChange" handling error: Failed to prepare rule checking: main.tf:2,13-19: Reference to undeclared local value
```

### TFLint Version

0.61.0

### Terraform Version

1.16.0

### Operating System

- [ ] Linux
- [x] macOS
- [ ] Windows

vulragrag-star:
## Summary

Fixes #2640.

When `tflint --langserver` receives `textDocument/didChange` for a temporarily invalid document (for example an incomplete `locals { example = }` while typing), `BuildRunners` returns wrapped `hcl.Diagnostics`. The language server previously propagated that as a JSON-RPC notification handling error on every keystroke.

This change:

- unwraps `hcl.Diagnostics` from the `BuildRunners` error with `errors.AsType`;
- converts them to LSP diagnostics (including clearing previously published paths);
- returns successfully from `inspect` so `didChange` / `didOpen` publish diagnostics instead of failing the handler.

Non-HCL failures still return as handler errors.

## Test plan

- [x] `go test ./langserver/ -count=1`
- [ ] Start `tflint --langserver`, open a valid `.tf` file, delete the value after an attribute while typing, confirm stderr has no `jsonrpc2 handler: notification "textDocument/didChange" handling error` and the editor shows the HCL diagnostic instead


vulragrag-star:
Addressed in a2050363: gci import order in `handler_test.go`, and switched the inspect test off `t.TempDir` so Windows cleanup after loader handles does not fail the job.

